PR #128, opened 2017-03-24
Fix for reflecting filter description ui-tooltip.

ArtsiomYeliseyenka:


src/main/resources/public/js/src/core/util.js
```diff
@@ -201,14 +201,15 @@ define(function (require, exports, module) {
             };
             Object.defineProperty(Array.prototype, "equals", {enumerable: false});
         },
-        appendTooltip: function (content, $el, $parrent, openCallback) {
+        appendTooltip: function (content, $el, $parrent, openCallback, className) {
```
lexecon: create new tooltip view

PR #129, opened 2017-03-27
EPMRPP-24574. Fix for infoline executions tooltip (updating after dat…

ArtsiomYeliseyenka:
…a was changed)


src/main/resources/public/js/src/launches/common/InfoPanelView.js
```diff
@@ -134,6 +134,7 @@ define(function (require, exports, module) {
         },
         update: function () {
             this.loadDefectTooltip();
+            this.loadExecutionsTooltip();
```
lexecon: Create a plugin without removing the old one.
Create a view for the tooltip in this file, and put it in the tooltip. Then just change the view after changing the model.

PR #141, opened 2017-03-29
Unassign user tooltips

ArtsiomYeliseyenka:


lexecon:
added for external version

PR #146, opened 2017-04-05
Change message no filter

yana95:
Change message in case there is no any filter on a project ( when added new widget)

### DIFF
--- a/src/main/resources/public/js/src/modals/addWidget/FilterSearchView.js
+++ b/src/main/resources/public/js/src/modals/addWidget/FilterSearchView.js
@@ -19,7 +19,7 @@
  * along with Report Portal.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-define(function (require, exports, module) {
+define(function (require) {
     'use strict';
 
     var Epoxy = require('backbone-epoxy');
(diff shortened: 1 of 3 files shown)


src/main/resources/public/js/src/modals/addWidget/FilterSearchView.js
```diff
@@ -266,7 +266,8 @@ define(function (require, exports, module) {
                         self.setFilterModel(self.collection.get(self.model.get('filter_id')));
                     }
                     $('[data-js-select-filter-block]', self.$el)[(!data.content.length ? 'add' : 'remove') + 'Class']('hide');
-                    $('[data-js-filter-empty]', self.$el)[(data.content.length ? 'add' : 'remove') + 'Class']('hide');
```
lexecon: use Epoxy bindings for remove or add class

src/main/resources/public/js/src/modals/addWidget/FilterSearchView.js
```diff
@@ -80,7 +95,7 @@ define(function (require, exports, module) {
             this.renderViews = [];
             this.render();
             this.viewModel = new Epoxy.Model({
-                search: '',
+                search: "",
```
lexecon: Use esLint to fix the syntax

PR #160, opened 2017-04-11
Back to top btn fix. Sidebar refactoring. Min app height = 415px

ArtsiomYeliseyenka:
…415px.


src/main/resources/public/templates/index.html
```diff
@@ -48,5 +47,6 @@
     </div>
     <div class="flash-container"></div>
 </div>
+<section class="sidebar js-sidebar" id="pageSidebar" style="display: none;"></section>
```
lexecon: To make these changes in src/main/resources/public/index.html

PR #165, opened 2017-04-12
Code coverage configuration added

Scorpibear:


.travis.yml
```diff
@@ -1,9 +1,30 @@
...
+  - sudo apt-get install -q google-chrome-stable
+
+before_script:
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
+
+script: 
+  - ./gradlew build
+  - cd src/main/resources/public
+  - npm run test
```
avarabyeu: do we really need it while gradle build already executes test?

Scorpibear: Just looked through the recent job logs on base branch - haven't found any traces of js tests being executed - https://travis-ci.org/reportportal/service-ui/jobs/226474472
so we need it either in .travis.yml or inside gradle.build or inside grunt tasks

PR #187, opened 2017-04-28
 EPMRPP-25283: filter is not shown

yana95:
Current filter is not shown as selected in case it has more than 50 filters before


src/main/resources/public/js/src/modals/addWidget/FilterSearchView.js
```diff
@@ -286,7 +288,12 @@ define(function (require) {
                     }
                     self.collection.add(data.content, { parse: true });
                     if (self.model.get('filter_id')) {
-                        self.setFilterModel(self.collection.get(self.model.get('filter_id')));
+                        if (self.collection.get(self.model.get('filter_id'))) {
+                            self.setFilterModel(self.collection.get(self.model.get('filter_id')));
+                        } else {
+                            launchFilterCollection = new SingletonLaunchFilterCollection();
```
lexecon: launchFilterCollection may not be ready, here it is necessary to wait for its readiness.

PR #189, opened 2017-04-29
EPMRPP-25373: Open link in a new tab by Ctrl+click

yana95:


src/main/resources/public/js/src/launches/common/LaunchSuiteItemView.js
```diff
@@ -310,15 +310,11 @@ define(function (require) {
                 config.trackingDispatcher.trackEventNumber(61.2);
             }
         },
-        onClickName: function (e) {
-            var href;
+        onClickName: function () {
```
lexecon: Add similar edits to the StepItemView.js file and change changelog.

PR #226, opened 2017-05-31
update unit tests

dkurachkin:


src/main/resources/public/js/src/dashboard/GadgetView.js
```diff
@@ -181,10 +181,11 @@ define(function (require) {
         },
         onClickGadgetEdit: function () {
             var self = this;
-            config.trackingDispatcher.trackEventNumber(286);
-            (new ModalEditWidget({
-                model: this.model
-            })).show()
+            if (this.model.get('isMy')) {
```
dkurachkin: admin can remove widgets, but not edit

PR #335, opened 2017-08-23
add console log view

yana95:


src/main/resources/public/js/src/launches/logLevel/LogItemLogsTable.js
```diff
@@ -43,16 +44,33 @@ define(function (require) {
 
         events: {
             'change [data-js-attachments-filter]': 'onChangeAttachments',
-            'click .rp-grid-th[data-sorter]': 'onClickSorter'
+            'click .rp-grid-th[data-sorter]': 'onClickSorter',
+            'change [data-js-switch-to-console-view]': 'onChangeView'
+        },
+        onChangeView: function () {
+            this.renderTableHeader();
+            this.collection.load();
```
lexecon: do not use collection.load(). Destroy itemViews and create new items with new mode.

PR #338, opened 2017-08-28
add console log view

yana95:


src/main/resources/public/js/src/launches/logLevel/LogItemLogsTable.js
```diff
@@ -43,16 +45,39 @@ define(function (require) {
...
+            } else {
+                $container.addClass('block-view');
+            }
         },
-
         initialize: function (options) {
+            var startOptions = options.options;
+            var isAscSort = 'true';
+            this.appStorage = new SingletonAppStorage();
+            this.consoleView = this.appStorage.get('consoleView') || false;
```
lexecon: why do you use this.consoleView?  This variable is used only in this function.

PR #355, opened 2017-09-13
replace post/load bug with issue

yana95:


src/main/resources/public/scss/component/beta-badge.scss
```diff
@@ -23,7 +23,7 @@
   position: absolute;
   background: $COLOR--dark-pastel-green;
   color: $COLOR--white-default;
-  display: block;
+  display: inline;
```
lexecon: use display: inline-block if you want to use "width" property.

PR #357, opened 2017-09-13
Refines conditions fix.

ArtsiomYeliseyenka:


src/main/resources/public/js/src/filterEntities/FilterEntities.js
```diff
@@ -176,11 +176,29 @@ define(function (require) {
...
+            conditionText: function (condition) {
+                return Localization.filterShortCut[condition];
+            }
+        },
         events: {
             'click .option-menu > li > a': 'changeOptions'
         },
         initialize: function () {
             this.render();
+            this.listenTo(this.model, 'change:condition', function (model, value) {
```
lexecon: conditionFlipper method is no longer needed (just set the value for the model)

PR #394, opened 2017-09-26
Localization new

yana95:


src/main/resources/public/scss/loginPage/login-login.scss
```diff
@@ -121,8 +122,13 @@
       }
 
       .or {
+        position: absolute;
```
lexecon: use position relative, and text-align: center for parent block. Line must be absolute position. Do not use transform for set position element (only for animation)

src/main/resources/public/js/src/core/util.js
```diff
@@ -31,20 +31,18 @@ define(function (require, exports, module) {
     var App = require('app');
     var Validators = require('validators');
     var Service = require('coreService');
-    var SingletonDefectTypeCollection = require('defectType/SingletonDefectTypeCollection');
 
     require('baron');
     require('cookie');
     // require('nicescroll');
     // require('isLoading');
-    require('select2');
```
lexecon: require('select2') should be everywhere where this library is used

PR #458, opened 2017-11-01
EPMRPP-29789. Launch statistics area & bar widgets combined into one.

ArtsiomYeliseyenka:


src/main/resources/public/js/src/newWidgets/_C3ChartWidgetView.js
```diff
@@ -78,6 +83,31 @@ define(function (require) {
...
+                        filtering_field: 'start_time',
+                        is_negative: false,
+                        value: time.format('x') + ',' + (parseInt(time.format('x'), 10) + range)
+                    };
+                    filterModel.parseServerData(response[0]);
+                    entities = filterModel.getEntitiesObj() || [];
+                    entities.push(dateFilter);
+                    filterModel.set('newEntities', JSON.stringify(entities));
+                    link = filterModel.get('url') + 'all?' + filterModel.getOptions().join('&');
+                    if (link) {
```
lexecon: destroy filter model after using

PR #538, opened 2017-11-29
EPMRPP-31183. Added search functionality for "add shared widget" modal.

ArtsiomYeliseyenka:


src/main/resources/public/js/src/core/data-url-resolver.js
```diff
@@ -79,8 +79,11 @@ define(['app'], function (App) {
     var widgetNamesShared = function () {
         return getProjectBase() + '/widget/names/shared';
     };
-    var widgetsShared = function () {
-        return getProjectBase() + '/widget/shared';
+    var widgetsShared = function (data) {
+        return getProjectBase() + '/widget/shared?page.size=10&page.page=' + data.page;
```
lexecon: set default state for data if data not defined

src/main/resources/public/js/src/core/data-url-resolver.js
```diff
@@ -79,8 +79,11 @@ define(['app'], function (App) {
...
+    };
+    var sharedWidgetSearch = function (data) {
+        return getProjectBase() + '/widget/shared/search?term=' + data.term + '&page.size=10&page.page=' + data.page;
```
lexecon: set default state for data if data not defined

PR #623, opened 2018-01-29
R profile page

ArtsiomYeliseyenka:


app/src/pages/profilePage/personalInfoBlock/personalInfoBlock.jsx
```diff
@@ -0,0 +1,169 @@
...
+
+          </div>
+        </BlockContainer.Body>
+      </div>
+    );
+  }
+
+}
+
+PersonalInfoBlock.propTypes = {
```
DiscoElevator: For components written as a class we can define theit prop types and default props as a class static fields:
```
class MyComponent extends Component {
  static propTypes = {
    ...
  };

  static defaultProps = {
    ...
  };

  ...
}
```

So we can see component's interface near it's definition.

PR #624, opened 2018-01-30
React. added project selector component.

ArtsiomYeliseyenka:


app/src/components/main/projectSelector/projectSelector.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+import classNames from 'classnames/bind';
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import ScrollWrapper from 'components/common/scrollWrapper/scrollWrapper';
+import styles from './projectSelector.scss';
+
+const cx = classNames.bind(styles);
+
+class ProjectSelector extends Component {
```
lexecon: It should receive in props onChange function and not store the current active item in state

PR #631, opened 2018-02-02
Localization reducer

lexecon:
Add BE localization and localization switcher


app/src/controllers/lang/reducer.js
```diff
@@ -0,0 +1,10 @@
+import { CHANGE_LANG_ACTION } from './constants';
+
+export const langReducer = (state = { lang: 'be' }, { type, payload }) => {
```
DiscoElevator: Maybe it's better to have a string as a state on this level? Like `state = 'be'`. So it will be `state.lang`, not `state.lang.lang` as a result.

app/src/controllers/lang/actionCreators.js
```diff
@@ -0,0 +1,8 @@
+import { CHANGE_LANG_ACTION } from './constants';
+
+export const setLangAction = lang => ({
+  type: CHANGE_LANG_ACTION,
+  payload: {
```
DiscoElevator: Why no `payload: lang`?

app/src/routes/rootRoute.jsx
```diff
@@ -17,13 +17,17 @@ import SettingsPage from 'pages/inside/settingsPage/settingsPage';
...
+
 const LoginRoute = () => (
   <EmptyLayout>
+    <LocalizationSwitcher />
     <LoginPage />
   </EmptyLayout>
 );
 const AppRoute = () => (
   <AppLayout>
+    <LocalizationSwitcher />
```
DiscoElevator: What about admin route?

lexecon: this is only for the test, in the future this component will be deleted. But in the admin area, too, I'll add

PR #633, opened 2018-02-05
Add auth check and login action

DiscoElevator:


app/src/controllers/auth/actions.js
```diff
@@ -18,23 +10,20 @@ const fetchUserSuccess = user => ({
...
-  headers: {
-    Authorization: token || DEFAULT_TOKEN,
-  },
-}).then((res) => {
-  if (res.status === 401) {
-    throw new Error('Unauthorized');
-  }
-  return res.json();
-}).then(user => dispatch(fetchUserSuccess(user)));
+export const fetchUserAction = () => dispatch =>
```
lexecon: it's not just downloading user data, it's better to split into two actions.

app/src/common/utils/fetch.test.js
```diff
@@ -0,0 +1,58 @@
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+import { DEFAULT_TOKEN } from 'controllers/auth';
```
lexecon: mock this module

PR #643, opened 2018-02-14
Auth controller divided into Auth and User controllers.

ArtsiomYeliseyenka:


app/src/controllers/user/reducer.js
```diff
@@ -0,0 +1,10 @@
+import { FETCH_USER_SUCCESS } from './constants';
+
+export const userReducer = (state = {}, { type, payload }) => {
+  switch (type) {
+    case FETCH_USER_SUCCESS:
+      return Object.assign({}, state, { info: payload, activeProject: 'default_project' });
```
DiscoElevator: This action corresponds to user only, don't set active project here.
Since there's no actions for changing active project, please make a constant `INITIAL_STATE`, set activeProject there and use it as a default value for state in the reducer function.

app/src/controllers/user/selectors.js
```diff
@@ -0,0 +1,6 @@
+const userSelector = state => state.user || {};
+const useInfoSelector = state => state.user.info || {};
```
DiscoElevator: duplicate

app/src/controllers/user/reducer.js
ArtsiomYeliseyenka: fixed

app/src/controllers/user/selectors.js
ArtsiomYeliseyenka: fixed

PR #646, opened 2018-02-20
Fixed user controller errors. Added logout action,

ArtsiomYeliseyenka:


app/src/controllers/auth/index.js
```diff
@@ -1,4 +1,4 @@
-export { TOKEN_KEY, DEFAULT_TOKEN } from './constants';
-export { loginAction, authSuccessAction } from './actionCreators';
+export { TOKEN_KEY, DEFAULT_TOKEN, LOGOUT } from './constants';
```
DiscoElevator: `LOGOUT` should not be exported from the module. `logoutAction` is enough.

ArtsiomYeliseyenka: fixed

PR #651, opened 2018-02-22
EPMRPP-33558 UI: Add a list to URL field on for AD and LDAP

yana95:


ArtsiomYeliseyenka:
Changes should be performed from master branch.

PR #654, opened 2018-02-28
EPMRPP-33590. Added tooltip engine. Added sandbox - /:projectId/sandbox.

ArtsiomYeliseyenka:


app/src/pages/inside/dashboardPage/index.js
```diff
@@ -1,2 +1,2 @@
 export { DashboardPage } from './dashboardPage';
-export { DashboardModal } from './dashboardModal';
+export { DashboardModal } from '../sandboxPage/exampleModal';
```
DiscoElevator: remove

PR #665, opened 2018-03-15
add lock file, env config file

SergeyKovalchuk:


Guria:
why yarn
shouldn't it be a team decision?

SergeyKovalchuk:
Let's discuss it, PULL REQUEST --- It's just request)
Would you like to discuss yarn vs npm topic?

app/.env
```diff
@@ -0,0 +1,5 @@
+PROXY_PATH=http://evbyminsd6293.minsk.epam.com:8080/
```
Guria: Can't see where this config is used. Don't we need introduce smth like `dotenv` to have effect?

app/.nvmrc
```diff
@@ -0,0 +1 @@
+6.10
```
Guria: `npm i node@6.10.0` ?
https://www.npmjs.com/package/node
https://github.com/aredridel/node-bin-gen

app/.env
avarabyeu: guys, please, no any EPAM server names in config files

Guria:
I guess it could be closed now, since dotenv was already implemented

DiscoElevator:
Closed in favor of #672

PR #676, opened 2018-03-23
EPMRPP-33167: launches page

DiscoElevator:


app/src/pages/inside/launchesPage/launchSuiteGrid/launchSuiteGrid.jsx
```diff
@@ -140,49 +140,62 @@ const COLUMNS = [
     component: HamburgerColumn,
   },
   {
+    name: 'name',
```
ArtsiomYeliseyenka: I suggest to rename that key to 'id'

PR #687, opened 2018-03-28
EPMRPP-33811 Made test items clickable on a History table

yana95:


src/main/resources/public/js/src/launches/historyGrid/HistoryItemCellView.js
```diff
@@ -123,8 +123,47 @@ define(function (require) {
                 }
             }
         },
+        getUrl: function () {
```
ArtsiomYeliseyenka: Try to cunstruct url with using pathNames field of the item model(this.model) instead of current url parsing.

PR #691, opened 2018-03-29
Bunch of fixes: eslint error, package meta, deprecated dependency

Guria:
I guess this component needs rework but now it causes linting errors for other PRs


Guria:
added #690 and #692

PR #692, opened 2018-03-29
Remove deprecated dependency

Guria:
I got an error with react-addons-css-transition-group within clean install from scratch.
It seems that Notification component isn't fully implemented yet.
So I just removed deprecated module usage for now and added todo item with link to migration guide.


Guria:
merged into #691

PR #694, opened 2018-03-30
EPMRPP-33812 History line. Make a launch number clickable

yana95:


src/main/resources/public/scss/launchPage/logLevel/log-history-line.scss
```diff
@@ -44,6 +44,9 @@
         span:first-child {
           display: none;
         }
+        .active-link{
+          color: #337ab7;
```
ArtsiomYeliseyenka: Use color constant from style guide.

PR #697, opened 2018-03-31
Eslint prettier

Guria:


Guria:
This will not force all codebase reformat but will enable reformatting staged file on precommit.
So we can merge it first, then wait until all current PRs is merged and then apply formatting to all codebase.

Guria:
Example of applied to codebase could be found here:
https://github.com/reportportal/service-ui/compare/eslintPrettier...prettierExmpl

app/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "arrowParens": "always",
```
Guria: options here is subject for discussion. See https://prettier.io/docs/en/options.html for details

Guria: updated according to discussion

PR #727, opened 2018-04-14
React. Launch level donut charts refactoring.

ArtsiomYeliseyenka:


app/src/pages/inside/launchesPage/launchSuiteGrid/defectStatistics/donutChart/donutChart.jsx
```diff
@@ -0,0 +1,96 @@
...
+          id: defect,
+          value: percents,
+          color: this.props.defectColors[defect],
+          offset: 100 - offset,
+        });
+        offset += percents;
+      }
+    });
+  }
+  chartData = [];
```
DiscoElevator: I think it's better to move it to the the state and initialize this state in the constructor instead of `componentWillMount` (which will be deprecated in the future).

PR #746, opened 2018-04-24
EPMRPP-33857. Added secs for launch duration when 1 min < time < 1 hour.

ArtsiomYeliseyenka:


app/src/common/utils/timeDateUtils.test.js
```diff
@@ -0,0 +1,91 @@
...
+const ONE_MINUTE = 60 * 1000;
+const ONE_HOUR = 3600 * 1000;
+const ONE_DAY = 24 * 3600 * 1000;
+
+describe('getDuration', () => {
+  test('should return sting', () => {
+    expect(typeof getDuration(NOW, NOW + ONE_HOUR)).toBe('string');
+  });
+
+  test('should return zero seconds', () => {
```
DiscoElevator: Should return "0s" in case of the same start and end

app/src/common/utils/timeDateUtils.test.js
```diff
@@ -0,0 +1,91 @@
...
+    expect(getDuration(NOW, NOW)).toBe('0s');
+  });
+
+  test('should contain seconds', () => {
+    expect(getDuration(NOW, NOW + ONE_SECOND * 20)).toMatch(/[0-9]+s+/);
```
DiscoElevator: I think it would be better to use strings instead of regular expressions, so anyone can look at tests and understand what this function can return in which case.

app/src/common/utils/timeDateUtils.test.js
```diff
@@ -0,0 +1,91 @@
...
+describe('getDuration', () => {
+  test('should return sting', () => {
+    expect(typeof getDuration(NOW, NOW + ONE_HOUR)).toBe('string');
+  });
+
+  test('should return zero seconds', () => {
+    expect(getDuration(NOW, NOW)).toBe('0s');
+  });
+
+  test('should contain seconds', () => {
```
DiscoElevator: Should contain seconds in which case?
The same question also applies to next tests

app/src/common/utils/timeDateUtils.test.js
```diff
@@ -0,0 +1,91 @@
...
+});
+
+describe('dateFormat', () => {
+  test('should return sting', () => {
+    expect(typeof dateFormat(0)).toBe('string');
+  });
+  test('should match match date pattern "@@@@-@@-@@ @@:@@:@@"', () => {
+    expect(dateFormat(0)).toMatch(/^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2}$/);
+  });
+  test('should contain UTC', () => {
```
DiscoElevator: What if 'UTC' will be at the beginning of a string? In the middle? Is it a valid case?

PR #770, opened 2018-05-11
Feature/hide notif onclick

IhorSyerkov:


app/src/components/main/notification/notificationContainer/notificationItem/notification.jsx
```diff
@@ -8,32 +8,38 @@ const cx = classNames.bind(styles);
 
 const messages = defineMessages({
   successLogin: { id: 'Notification.successLogin', defaultMessage: 'Signed in successfully' },
+  failureLogin: { id: 'Notification.failureLogin', defaultMessage: 'Wrong username or password' },
```
ArtsiomYeliseyenka: As I see on our backbone version, in case of failure login attempt we show "An error occurred while connecting to server: " + error message returned in response from WS

app/src/controllers/auth/sagas.js
```diff
@@ -0,0 +1,52 @@
...
+      params: {
+        grant_type: 'password',
+        username: payload.login,
+        password: payload.password,
+      },
+      method: 'POST',
+    });
+  } catch (e) {
+    const error = (e.response && e.response.data && e.response.data.message) || e.message;
+    yield put(showNotification({ messageId: 'failureDefault', type: 'error', values: { error } }));
```
DiscoElevator: Please use notification type constants from 'controllers/notification/constants.js'

app/src/components/main/notification/notificationContainer/notificationItem/notification.jsx
```diff
@@ -1,39 +1,50 @@
...
     type: PropTypes.oneOf(['', 'error', 'info', 'success']),
+    onMessageClick: PropTypes.func.isRequired,
     intl: intlShape.isRequired,
+    values: PropTypes.object,
   };
   static defaultProps = {
     message: '',
     messageId: '',
-    type: '',
+    type: 'info',
```
DiscoElevator: Can we use type constants from 'controllers/notification'?

PR #774, opened 2018-05-16
EPMRPP-34460 Create permissions resolver with tests

yana95:


app/src/common/utils/permissions/permissions.test.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+  it('should return true if user has permission "ALL"', () => {
+    expect(canCreateUser('USER', 'CUSTOMER')).toBeTruthy();
+  });
+  it('should return true if owner has permission "ALL"', () => {
+    expect(canCreateUser('USER', 'CUSTOMER', true)).toBeTruthy();
+  });
+  it('should return true if owner has permission "OWNER"', () => {
+    expect(canCreateUser('USER', 'MEMBER', true)).toBeTruthy();
+  });
```
DiscoElevator: Please also add test for case with OWNER permission when user is not an owner

app/src/common/utils/index.js
```diff
@@ -40,3 +36,26 @@ export {
...
+  canEditOwnAccount,
+  canDeleteLaunch,
+  canForceFinishLaunch,
+  canStartAnalysis,
+  canDeleteTestItem,
+  canMoveToDebug,
+  canMergeLaunches,
+  canWorkWithFilters,
+  canReadData,
+} from 'common/utils/permissions';
```
DiscoElevator: As I see, this `index.js` grows bigger and bigger for no particlular reason. I don't think we should export everythink here, since `utils` is not a single module with interface, it's just a folder with different utils. I suggest removing this esport from here (other exports will be removed in future), and importing these functions directly from util file, for example:
`import { canMergeLaunches } from 'common/utils/permissions'`

This will make imports more specific.

@ArtsiomYeliseyenka what do you think?

ArtsiomYeliseyenka: +1

PR #800, opened 2018-06-07
EPMRPP-34251. Fix for edit launch permissions.

ArtsiomYeliseyenka:


app/src/controllers/user/selectors.js
```diff
@@ -8,6 +8,7 @@ export const userIdSelector = (state) => userInfoSelector(state).userId;
 export const settingsSelector = (state) => userSelector(state).settings;
 export const startTimeFormatSelector = (state) => settingsSelector(state).startTimeFormat;
 export const assignedProjectsSelector = (state) => userInfoSelector(state).assigned_projects || {};
+export const userAccountSelector = (state) => userInfoSelector(state).userRole || '';
```
DiscoElevator: userRoleSelector

app/src/common/utils/permissions/permissions.js
```diff
@@ -36,6 +36,7 @@ export const canDeleteUser = checkPermission(ACTIONS.DELETE_USER);
 export const canSeeMembers = checkPermission(ACTIONS.SEE_MEMBERS);
 export const canEditOwnAccount = checkPermission(ACTIONS.EDIT_OWN_ACCOUNT);
 export const canDeleteLaunch = checkPermission(ACTIONS.DELETE_LAUNCH);
+export const canEditLaunch = checkPermission(ACTIONS.EDIT_LAUNCH); // requires userRole, projectRole, isOwner.
```
DiscoElevator: What is the purpose of this comment?

app/src/controllers/user/selectors.js
ArtsiomYeliseyenka: misprint from my side. Better name is userAccountRoleSelector

app/src/common/utils/permissions/permissions.js
ArtsiomYeliseyenka: as i heard not all these check requires all 3 arguments. With a comment beside I'll know immediatly how many arguments I should pass for correct permission calculation.

PR #815, opened 2018-06-14
Improve email validation to handle domains with more than one dot.

kleptog:
This is really simple change. Rather silly limitation.


kleptog:
Pull request on other branch

PR #822, opened 2018-06-20
EPMRPP-35129 REACT. Launch export functions

yana95:


app/src/pages/inside/launchesPage/launchSuiteGrid/hamburger/hamburger.jsx
```diff
@@ -172,36 +171,24 @@ export class Hamburger extends Component {
...
-                <GhostButton
-                  tiny
-                  onClick={() => {
-                    onExportHTML(launch);
-                  }}
-                >
-                  HTML
-                </GhostButton>
-              </div>
+              <a
```
ArtsiomYeliseyenka: Use GhostButtons on open required window.location in onClick callback or add style for your links in case when button pressed.

app/src/pages/inside/common/launchSuiteGrid/hamburger/hamburger.jsx
```diff
@@ -166,39 +169,23 @@ export class Hamburger extends Component {
...
             </div>
             <div className={cx('export-buttons')}>
               <div className={cx('export-button')}>
-                <GhostButton
-                  tiny
-                  onClick={() => {
-                    onExportPDF(launch);
-                  }}
-                >
+                <GhostButton tiny onClick={() => this.onExportLaunch('pdf')}>
```
DiscoElevator: I'd make a class method for every export type so we don't have to make new function on every render.
For example:
```
exportAsPDF = () => this.exportLaunch('pdf');

render() {
  //...
  <GhostButton tiny onClick={this.exportAsPDF}>
  // ...
}
```

ArtsiomYeliseyenka: +1

PR #827, opened 2018-06-21
Improve email validation to handle domains with more than one dot.

kleptog:
This is really simple change. Rather silly limitation.

See also reportportal/service-ui#815


DzmitryHumianiuk:
@kleptog thank you for contribution.
any help is helpful.

PR #838, opened 2018-06-28
EPMRPP-35153 REACT. Filter options on filter page

yana95:


app/src/pages/inside/filtersPage/filterGrid/filterOptions/filterOptions.jsx
```diff
@@ -0,0 +1,185 @@
...
+    )} ${this.fotmatTime(time.end)} ${time.dynamic && intl.formatMessage(messages.dynamic)}`;
+    return `${optionName} ${intl.formatMessage(messages.from)} ${condition}`;
+  };
+
+  optionsToString = () => {
+    const { intl } = this.props;
+    let optionName;
+    let condition;
+    const result = [];
+    this.props.entities.forEach((entity) => {
```
DiscoElevator: `const result = this.props.entities.map( // ...`

PR #845, opened 2018-07-10
EPMRPP-35139. Added breadcrumbs functionality

PavelZaletski:
Added breadcrumbs functionality


app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -46,33 +47,50 @@ const messages = defineMessages({
...
   };
 
   render() {
-    const { formatMessage } = this.props.intl;
+    const {
+      activeProject,
+      intl: { formatMessage },
+    } = this.props;
+
+    const breadcrumbs = [
```
ArtsiomYeliseyenka: I think i will be better to represent breadcrumb config like array of objects:
`
[
  {
    title: formatMessage(messages.pageTitle),
    to: { type: PROJECT_DASHBOARD_PAGE, payload: { projectId: activeProject } },
  },
  {
    title: formatMessage(messages.subPageTitle),
  },
]
`
And then check object property 'to'. If it's exists - draw link, no - draw plain text.

PavelZaletski: makes sense. Will do

app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -58,24 +64,68 @@ export class DashboardItemPage extends Component {
...
   };
 
   static defaultProps = {
-    dashboardId: undefined,
+    dashboardItems: [],
   };
 
   state = {
     isFullscreen: false,
+    dashboardName: '',
```
ArtsiomYeliseyenka: It's not needed to keep dashboardName in local state.
You have all access to dashboard items received with props. 
You can get active dashboard name just filtering dashboardItems array by id with dashboardId prop.

app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -58,24 +64,68 @@ export class DashboardItemPage extends Component {
...
+    const dashboard = dashboardItems.find((item) => item.id === dashboardId);
+
+    if (dashboard) {
+      dashboardName = dashboard.name;
+    }
+
+    return dashboardName;
+  };
+
+  getTitle = () => {
```
ArtsiomYeliseyenka: getBreadcrumbs is better name. Please fix all following cases.

app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -85,9 +135,9 @@ export class DashboardItemPage extends Component {
 
   render() {
     const { formatMessage } = this.props.intl;
+
     return (
-      <PageLayout title={formatMessage(messages.pageTitle)}>
-        <AddDashboardButton />
+      <PageLayout Title={this.getTitle()}>
```
ArtsiomYeliseyenka: Please do not define attributes with a capital letter in the beginning. Fix all following cases.

app/src/pages/inside/common/dashboardHeader/dashboardHeader.jsx
```diff
@@ -0,0 +1,22 @@
...
+import { PageBreadcrumbs } from '../pageBreadcrumbs';
+import styles from './dashboardHeader.scss';
+
+const cx = classNames.bind(styles);
+
+export const DashboardHeader = ({ breadcrumbs }) => (
+  <div className={cx('dashboard-header')}>
+    <PageBreadcrumbs data={breadcrumbs} />
+    <AddDashboardButton />
+  </div>
```
ArtsiomYeliseyenka: this component is unnecessary.  For example, in DashboardItemPage: 
```
<PageLayout breadcrumbs={getBreadCrumbs()}>
   <AddDashboardButton /> // absolutely  positioned, with a top property below zero.
   ...
   ...
</PageLayout>
```

app/src/pages/inside/filtersPage/filtersPage.jsx
```diff
@@ -143,7 +149,7 @@ export class FiltersPage extends Component {
       ...rest
     } = this.props;
     return (
-      <PageLayout title={intl.formatMessage(messages.filtersPageTitle)}>
+      <PageLayout Title={this.getTitle()}>
```
ArtsiomYeliseyenka: <PageLayout breadcrumbs={this.getBreadcrumbs()}>. Please rewrite all following cases in the same style

app/src/pages/inside/common/dashboardHeader/dashboardHeader.jsx
DiscoElevator: I disagree with this opinion since there can be many such absolutely positioned elements, and it will make markup less readable:
1. It's difficult to understand that this button is actually outside of the page content. These components are not a part of a page content, so it's strange to see them there. It's a title/header/ or whatever we may call it, but not a content.
2. It's difficult to maintain positions of many absolutely positioned components between each other.

ArtsiomYeliseyenka: On other side that button functionality related to page content, not layout

ArtsiomYeliseyenka: It's correct. But DashboardItemPage returns PageLayout which divides page into header and content. In header we keep page title or breadcrumbs, which are responsible for displaying of user location. All other info and functionality are placed into content block. All interactions with dashboards are placed in content block. So I can't see any reason to button of adding a new dashboard into header, which is not responsible for that.

DiscoElevator: I suggest to do some refactoring of the `PageLayout` component and its' usages:
- Instead of the `title` prop, let's make a component `PageHeader` (move and refactor existing `PageTitle` to a new file)
- Also, move `PageContent` to a new file too (suggested name is `PageSection`)

So, pageLayout will have 3 exported components: `PageLayout`, `PageHeader`, `PageContent`.
And the usage will be like this:
```
<PageLayout>
  <PageHeader>
    {/* Some title, for example, in this case may be `DashboardHeader` */}
  </PageHeader>
  <PageSection>
    {/* page content */}
  </PageSection>
</PageLayout>
```

Also, speaking of `DashboardHeader` itself, is it better to keep this component of just place breadcrumbs and the button inside the `PageHeader` directly and give an opportunity to control layout (title on the left, buttons on the right)? Maybe we need some prop whick adds a class with `display: flex; justify-content: space-between;`? @ArtsiomYeliseyenka what do you think?

ArtsiomYeliseyenka: Confirm solution of dividing PageLayout on a three components.
About positioning breadcrambs and controls in PageHeader: I suggest to define  space-between by default, because at the moment, there is no cases when such block positioned in other way. If in future we will face differ requirement, it will be easy to add modify-class prop.

PR #846, opened 2018-07-10
EPMRPP-35240 Bug with comma in time zona

yana95:


DzmitryHumianiuk:
https://github.com/reportportal/reportportal/issues/440

PR #847, opened 2018-07-12
EPMRPP-35130: nested routes

DiscoElevator:


app/src/controllers/filter/withFilter.jsx
```diff
@@ -15,41 +15,37 @@ const debounce = (callback, time) => {
...
+export const withFilter = (WrappedComponent) =>
+  connectRouter(
+    (query) => ({
+      filter: query[FILTER_KEY],
+    }),
+    {
+      updateFilter: (filter) => ({ [FILTER_KEY]: filter }),
+    },
+  )(
+    class FilterWrapper extends PureComponent {
```
ArtsiomYeliseyenka: Please rewrite with functional component

PR #850, opened 2018-07-17
EPMRPP-35229. Add 'Add new dashboard' button to dashboard page

PavelZaletski:
Add 'Add new dashboard' button to dashboard page


app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -73,6 +74,7 @@ export class DashboardItemPage extends Component {
     const { formatMessage } = this.props.intl;
     return (
       <PageLayout title={formatMessage(messages.pageTitle)}>
+        <AddDashboardButton />
```
DiscoElevator: I think it's better to create a new component for title and pass this component to a `title` prop. It will help to avoid absolute positioning and also will be useful for adding breadcrumbs

PavelZaletski: Will be implemented in scope of EPMRPP-35139

PR #851, opened 2018-07-23
EPMRPP-35263 Duration fluctuation on History Line

yana95:


src/main/resources/public/scss/launchPage/logLevel/log-history-line.scss
```diff
@@ -62,6 +65,39 @@
...
+          background-image: url('/img/launch/ic_info.svg');
+          background-size: contain;
+        }
+        .growth-duration{
+          display: inline-block;
+          position: relative;
+          margin-right: 3px;
+          padding: 0 2px;
+          border-radius: 100px;
+          background-color: lighten($COLOR--white, 20%);
```
ArtsiomYeliseyenka: Do we need to use color function here? There is no required color variable?

PR #855, opened 2018-07-29
EPMRPP-35331 The most time-consuming test cases widget

yana95:


src/main/resources/public/js/src/newWidgets/widgets/mostTimeConsumingTestCases/MostTimeConsumingTestCasesTableView.js
```diff
@@ -0,0 +1,64 @@
...
+            var contentData = this.model.getContent();
+            this.$el.html(Util.templates(this.template, contentData));
+            _.each(contentData.result, function (item) {
+                var itemView = new TestCasesItemView({
+                    model: new Epoxy.Model(item),
+                    url: this.getFilterByUIDRedirectLink(contentData.lastLaunch[0].id, item.uniqueId)
+                });
+                $('[data-js-items-container]', this.$el).append(itemView.$el);
+            }, this);
+            this.scroller = Util.setupBaronScroll($('[data-js-scroll]', this.$el));
```
ArtsiomYeliseyenka: please add onBeforeDestroy method and dispose this.scroller.baron(). See CumulativeTrandChartView.js as example

ArtsiomYeliseyenka:
Also in table view, launch name block shouldn't be scrolled

PR #869, opened 2018-08-03
EPMRPP-34323: ignore/include in auto-analysis modals

DiscoElevator:


DiscoElevator:
Will be ready for review after rebasing on #854

PR #909, opened 2018-08-31
EPMRPP-35718 Duration fluctuation. Add system tool tip

yana95:


src/main/resources/public/js/src/core/util.js
```diff
@@ -592,7 +592,7 @@ define(function (require, exports, module) {
             });
         },
 
-        timeFormat: function (start, end) {
+        timeFormat: function (start, end, threeDigitsAfterComa) {
```
DiscoElevator: threeDigitsAfterCo**m**ma

PR #924, opened 2018-09-11
EPMRPP-35523: Fix query params reset

DiscoElevator:


app/src/common/utils/omit.js
```diff
@@ -0,0 +1,5 @@
+export const omit = (object, keys) =>
```
ultd39: Please, add tests for this function.

PR #928, opened 2018-09-11
Epmrpp 33462 launches comparison chart

PavelZaletski:
Launches Comparison Chart


app/src/pages/inside/dashboardItemPage/newWidgets/c3chart/c3chart.jsx
```diff
@@ -0,0 +1,79 @@
...
+    if (config.unloadBeforeLoad) {
+      this.unloadData();
+    }
+
+    this.loadNewData(config.data);
+  }
+
+  render() {
+    const className = this.props.className ? ` ${this.props.className}` : '';
+    const style = this.props.style ? this.props.style : {};
```
ultd39: This check is unnecessary - `style` prop will be object if it's not provided because you declared that in defaultProps.
I think, you can just use `style` and `className` directly from props.

PR #930, opened 2018-09-13
EPMRPP-35176 This a fix for hardcoded 7s in duration field.

gyurcigyurma:
This a fix for hardcoded 7s in duration field.
Maybe it would be handy to solve a default case if props has no/undefined for start_time/end_time. Any suggestion is welcomed.

https://jiraeu.epam.com/browse/EPMRPP-35176

### DIFF
--- a/app/src/pages/inside/common/infoLine/infoLine.jsx
+++ b/app/src/pages/inside/common/infoLine/infoLine.jsx
@@ -6,6 +6,7 @@ import styles from './infoLine.scss';
 import { BarChart } from './barChart';
 import { Duration } from './duration';
 import { DefectTypeBlock } from './defectTypeBlock';
+import { getDuration } from '../../../../common/utils/timeDateUtils';
 
 const cx = classNames.bind(styles);
 const messages = defineMessages({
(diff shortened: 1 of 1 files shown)


app/src/pages/inside/common/infoLine/infoLine.jsx
```diff
@@ -33,6 +33,9 @@ export class InfoLine extends Component {
     const passed = executions.passed / executions.total * 100;
     const failed = executions.failed / executions.total * 100;
     const skipped = executions.skipped / executions.total * 100;
+    const endTime = this.props.data.end_time;
+    const startTime = this.props.data.start_time;
+    const duration = (endTime - startTime) / 1000;
```
DiscoElevator: It's better to use the util function `getDuration` from `common/utils/timeDateUtils.js`

gyurcigyurma: Thank you, I just spotted this util later. I'll fix.

app/src/pages/inside/common/infoLine/infoLine.jsx
```diff
@@ -6,6 +6,7 @@ import styles from './infoLine.scss';
 import { BarChart } from './barChart';
 import { Duration } from './duration';
 import { DefectTypeBlock } from './defectTypeBlock';
+import { getDuration } from '../../../../common/utils/timeDateUtils';
```
DiscoElevator: It is better to use absolute import here:
`import { getDuration } from 'common/utils/timeDateUtils';`

PR #940, opened 2018-09-17
UI Dev Guide. 8 - CSS.

ArtsiomYeliseyenka:


app/docs/8-CSS.md
```diff
@@ -0,0 +1,23 @@
...
+
+'/src/common/css/fonts/' - fonts.
+
+'/src/common/css/variables/' - color, font and screen breakpoint variables.
+
+As far as usually components has small size, it's preferable to do not use css selectors nesting (exclude pseudo elements, pseudo classes and state-classes).
+
+It's allowed to use flex-box properties.
+
+We are strongly recommend to follow [CSS properties order guide](https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/).
```
DiscoElevator: I think it's better to add such list here so we can easily maintain it.

app/docs/8-CSS.md
```diff
@@ -0,0 +1,23 @@
+# CSS
+
+On our project we are using [SCSS syntax](https://sass-scss.ru/guide/) and [CSS modules](https://github.com/css-modules/css-modules).
+
+'/src/common/css/common.scss' - commons styles.
```
DiscoElevator: common styles

app/docs/8-CSS.md
```diff
@@ -0,0 +1,23 @@
+# CSS
+
+On our project we are using [SCSS syntax](https://sass-scss.ru/guide/) and [CSS modules](https://github.com/css-modules/css-modules).
```
DiscoElevator: https://sass-lang.com/guide
To be more consistent with other links (which are mainly lead to the English versions by default)

app/docs/8-CSS.md
```diff
@@ -0,0 +1,56 @@
...
+
+'/src/common/css/fonts/' - fonts.
+
+'/src/common/css/variables/' - color, font and screen breakpoint variables.
+
+As far as usually components has small size, it's preferable to do not use css selectors nesting (exclude pseudo elements, pseudo classes and state-classes).
+
+It's allowed to use flex-box properties.
+
+Main element of component should have class name the same as component's name, but in lowercase and dash-notation (ComponentName / component-name).
```
DiscoElevator: Speaking of this requirement... Since we use css-modules, I propose to make this as not required, but recommended. Also, we already have complex cases which don't follow this rule, such as`GridRow`.

PR #948, opened 2018-09-20
EPMRPP-35502 No number of run of the launch is in breadcrumbs fix

gyurcigyurma:
https://jiraeu.epam.com/browse/EPMRPP-35502

### DIFF
--- a/app/src/controllers/testItem/selectors.js
+++ b/app/src/controllers/testItem/selectors.js
@@ -49,7 +49,12 @@ const isListView = (query, namespace) => {
 export const isListViewSelector = (state, namespace) =>
   isListView(pagePropertiesSelector(state), namespace);
 
-const itemTitle = (item) => `${item.name} ${item.number ? `#${item.number}` : ''}`;
+const itemTitleFormatter = (item) => {
+  if (item.number || item.number === 0) {
+    return `${item.name} #${item.number}`;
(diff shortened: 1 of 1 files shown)


app/src/controllers/testItem/selectors.js
```diff
@@ -49,6 +49,8 @@ const isListView = (query, namespace) => {
 export const isListViewSelector = (state, namespace) =>
   isListView(pagePropertiesSelector(state), namespace);
 
+const itemTitle = (item) => `${item.name} ${item.number ? `#${item.number}` : ''}`;
```
ArtsiomYeliseyenka: `${item.name}${item.number && ` #${item.number}`}`;

gyurcigyurma: I see your point but in this case if the nested level has no number, then return an undefined. 
image
@ypetya has a good point: Is it possible that item.number == 0 ? If yes then probably we need an even better solution.

DiscoElevator: I think a formatter function will be enough. I don't think that item.number can be 0, but it would be good to cover this case too.

gyurcigyurma: @DiscoElevator  Could you please check it again? Thank you.

PR #968, opened 2018-10-01
EPMRPP-34389 fixed empty selectedlaunches-array at page start

RavicVI:


app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -182,6 +182,10 @@ export class LaunchesPage extends Component {
     deleteItemsAction: () => {},
   };
 
+  componentWillMount(){
```
DiscoElevator: 1. I think it's better to do this in `componentWillUnmount`.
2. Please add this fix to other pages: suites, tests, steps.

PR #971, opened 2018-10-01
EPMRPP-35554: LogsGridToolbar implementation

ultd39:


app/src/pages/inside/logsPage/logsGridToolbar/utils/logLevel.js
```diff
@@ -0,0 +1,41 @@
...
+const getLogLevelById = (logLevelId) => LOG_LEVELS.find((logLevel) => logLevel.id === logLevelId);
+
+const getLogLevelFromStorage = () => getLogLevelById(getStorageItem(LOG_LEVEL_STORAGE_KEY));
+
+const getDefaultLogLevel = () => getLogLevelById(DEFAULT_LOG_LEVEL);
+
+export const getLogLevel = () =>
+  getLogLevelFromStorage() || getDefaultLogLevel() || LOG_LEVELS[LOG_LEVELS.length - 1];
+
+export const setLogLevel = (logLevel) => setStorageItem(LOG_LEVEL_STORAGE_KEY, logLevel.id);
```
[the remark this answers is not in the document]
ultd39: Fixed. Now this value stored in `${userId}_settings.logFilteringLevel`. And I also moved `withAttachments` flag to `${userId}_settings.logFilteringLevelWithAttachments`.

ultd39:
All issues resolved, but I'm not sure about solutions.

app/src/pages/inside/logsPage/logsGridToolbar/logsGridToolbar.jsx
```diff
@@ -0,0 +1,176 @@
...
+    children: () => {},
+    onChangeLogLevel: () => {},
+    onChangeWithAttachments: () => {},
+  };
+
+  state = {
+    markdownMode: false,
+    consoleView: false,
+    withAttachments: getWithAttachments(this.props.userId),
+    logLevel: getLogLevel(this.props.initialLogLevel, this.props.userId),
```
DiscoElevator: I think that we shouldn't copy prop `initialLogLevel` to the state. The case: if we change URL, this change won't be reflected on the page.
Let's make URL the main source of truth here. You can also cache default value to reduce localStorage usage on every render.

ultd39: Yes you right. Fixed.

PR #972, opened 2018-10-02
EPMRPP-35858 UI: Demo instance

yana95:


src/main/resources/public/js/src/main-view.js
```diff
@@ -43,12 +43,13 @@ define(function (require) {
             this.sidebarView = new Sidebar({
                 tpl: 'tpl-sidebar',
                 projectUrl: config.project.projectId,
-                currentPage: this.contextName
+                currentPage: this.contextName,
+                isDemo: true,
```
DiscoElevator: Let's hardcode it on the model level

PR #957, opened 2018-09-24
EPMRPP-35547 React. LogItemActivity tab

ultd39:

PR #983, opened 2018-10-08
EPMRPP-35548 Attachment modal

norbert-csaba-herczeg-epam:
Take over feature branch.


app/src/common/img/rotate-inline.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="20" height="20" viewBox="0 0 20 20">
+  <path fill="#999" fill-rule="evenodd" d="M6.52143322,5.79548129 L1,11.2899506 L6.52143322,16.7844199 L12.0428664,11.2899506 L6.52143322,5.79548129 L6.52143322,5.79548129 Z M3.4052504,11.2899506 L6.51717614,8.19322193 L9.63335895,11.2899506 L6.52143322,14.3866792 L3.4052504,11.2899506 L3.4052504,11.2899506 Z M16.7554544,5.98187809 C15.2569621,4.49493999 13.2987051,3.74935279 11.3361911,3.74935279 L11.3361911,1 L7.72192988,4.59661097 L11.3361911,8.18898564 L11.3361911,5.44386915 C12.8602259,5.44386915 14.3885177,6.024241 15.5507006,7.18074841 C17.8793236,9.49799953 17.8793236,13.2513533 15.5507006,15.5686044 C14.3885177,16.7251122 12.8602259,17.3054836 11.3361911,17.3054836 C10.5103175,17.3054836 9.688701,17.1317957 8.92242654,16.7928929 L7.65381659,18.0553071 C8.79471413,18.6822782 10.0633241,19 11.3361911,19 C13.2987051,19 15.2569621,18.2544137 16.7554544,16.7674747 C19.7481819,13.7893622 19.7481819,8.95999059 16.7554544,5.98187809 L16.7554544,5.98187809 Z"/>
```
ultd39: Please avoid hardcode color in svg. You can set `fill` rule where you use this icon.

norbert-csaba-herczeg-epam: @ultd39 I believe that the SVG was imported as-is from the site build / provided assets. Could you please clarify your suggestion?

ultd39: Actually I never use imported svg as-is because in most cases it contains a lot of unnecessary elements, which appear during creating image or icon. For remove it I use some of svg optimizers ([svgomg](https://jakearchibald.github.io/svgomg/) for example) and change a code manually after that. As result I have optimized version which much less than original.

About colors. Not so long time ago I had a case where I need to use existing icon but in other color and of course it impossible when color is hardcoded in fill property.

Please [see my pull request](https://github.com/reportportal/service-ui/pull/957/files#diff-f08d85472049981f1a441c39c35fd70d) as example: here I removed fill hardcode and optimized it.

DiscoElevator: Many of our icons can have multiple states, so if we used them as is, we would have different svg files for every state (default, hover, disabled). And the only difference in these files would be is the `fill` attribute value.
I think having one file and controlling its' color by css is better in this case.

norbert-csaba-herczeg-epam: @ultd39 @DiscoElevator Thanks, I'll modify the PR to adhere to your suggestions!

norbert-csaba-herczeg-epam: @ultd39 done.

norbert-csaba-herczeg-epam:
@DiscoElevator I believe that every request has been fulfilled. Let me know if there is anything else that needs to be addressed!

app/src/components/main/modal/modalLayout/modalFooter/modalFooter.scss
```diff
@@ -1,3 +1,10 @@
+.rotate-icon {
```
DiscoElevator: I can't find any usages of this class. Looks like this style can be removed

app/src/pages/inside/logsPage/modals/attachmentModal/attachmentCodeModal.jsx
```diff
@@ -0,0 +1,38 @@
...
+    } = this.props;
+    const cancelButton = {
+      text: intl.formatMessage(messages.close),
+      onClick: (closeModal) => closeModal(),
+    };
+
+    return (
+      <ModalLayout title={intl.formatMessage(messages.title)} cancelButton={cancelButton}>
+        <form>
+          <Highlight {...language}>{content}</Highlight>
```
DiscoElevator: Looks like you're trying do apply destructuring on string

app/src/pages/inside/logsPage/modals/attachmentModal/attachmentImageModal.jsx
```diff
@@ -0,0 +1,69 @@
...
+    const rotationA = this.state.rotationAmount;
+    this.setState({
+      rotationAmount: rotationA - 90,
+    });
+  };
+
+  generateRotationCommand = (amount) => `rotate(${amount}deg)`;
+
+  renderCustomButton = () => (
+    <div onClick={this.rotateImageHandler} left="true" className={cx('attachment-image-rotate')}>
```
DiscoElevator: `left="true"` -> `left`

app/src/pages/inside/logsPage/modals/attachmentModal/attachmentImageModal.jsx
```diff
@@ -0,0 +1,69 @@
...
+    const command = this.generateRotationCommand(this.state.rotationAmount);
+    const style = { transform: command };
+
+    return (
+      <ModalLayout
+        title={intl.formatMessage(messages.title)}
+        okButton={this.renderOkButton(intl, messages)}
+        customButton={this.renderCustomButton(intl, messages)}
+      >
+        <div>
```
DiscoElevator: Is this div necessary?

app/src/pages/inside/logsPage/modals/attachmentModal/attachmentImageModal.jsx
```diff
@@ -0,0 +1,69 @@
...
+          <img style={style} id="attachment-image" alt="attachment" src={image} />
```
DiscoElevator: Is `id` necessary?

app/.eslintignore
```diff
@@ -5,3 +5,4 @@ localization/messages
 localization/translated/en.json
 package.json
 package-lock.json
+src/pages/inside/logsPage/modals/attachmentModal/har/WithZipJs.jsx
```
DiscoElevator: I think it's better to add eslint directive to the file itself instead of this line

norbert-csaba-herczeg-epam:
@DiscoElevator Updated according to your observations.

PR #990, opened 2018-10-11
EPMRPP-33151 React. Launches. History view

yana95:
Signed-off-by: Yana_Zaitsava <Zaitsava_Yana@epam.com>


app/src/controllers/itemsHistory/selectors.js
```diff
@@ -0,0 +1,18 @@
+import { itemsSelector } from 'controllers/testItem';
+
+const domainSelector = (state) => state.itemsHistory || {};
+export const itemsHistorySelector = (state) => domainSelector(state).items;
+export const historySelector = (state) => domainSelector(state).history;
+export const visibleItemsCountSelector = (state) => domainSelector(state).visibleItemsCount;
+export const loadingSelector = (state) => domainSelector(state).loading;
+export const historyItemsSelector = (state) => {
```
DiscoElevator: It's better to memoize this selector to avoid unnecessary rerenders.
```
import {createSelector} from 'reselect';
export const historyItemsSelector = createSelector(itemsSelector, (testItems) => {
 // ...
});
```

app/src/controllers/itemsHistory/selectors.js
```diff
@@ -0,0 +1,18 @@
...
+  const launchesToRender = [];
+
+  itemsSelector(state).forEach((item) => {
```
DiscoElevator: Can be done by using `reduce`

PR #992, opened 2018-10-11
EPMRPP-35563 React. Log level. Copy/Send defect type modal

AmsterGet:


app/src/common/urls.js
```diff
@@ -83,6 +83,7 @@ export const URLS = {
     `${urlBase}project/${activeProject}/emailconfig`,
   suite: (activeProject, suiteId) => `${urlBase}${activeProject}/item/${suiteId}`,
 
+  testItemsUpdateIssues: (activeProject) => `${urlBase}${activeProject}/item`,
```
DiscoElevator: Can we use `testItem` url below?

AmsterGet: I'll try to use it

app/src/pages/inside/logsPage/modals/copySendDefectModal/copySendDefectModal.jsx
```diff
@@ -0,0 +1,134 @@
...
+      {
+        issue: {
+          ...itemForCopy.issue,
+          externalSystemIssues: itemForCopy.issue.externalSystemIssues || [],
+        },
+        test_item_id: lastHistoryItem.id,
+      },
+    ];
+
+    fetch(URLS.testItem(activeProject, ''), {
```
DiscoElevator: It's better to rewrite url function like this:
`testItem: (activeProject, id = '') => //...`
and then usage in this case will be:
`URLS.testItem(activeProject)`

AmsterGet: Sorry for my inattention. Fixed

PR #1001, opened 2018-10-17
EPMRPP-33239 Launches. Log level. Item info. Attachments

norbert-csaba-herczeg-epam:


app/src/controllers/attachments/sagas.js
```diff
@@ -0,0 +1,81 @@
...
+  ATTACHMENT_IMAGE_FETCH_ERROR,
+} from './constants';
+
+function fetchImageData({ projectId, binaryId }) {
+  return new Promise(() => ({
+    image: URLS.getFileById(projectId, binaryId),
+  }));
+}
+
+function fetchRealData({ projectId, binaryId }) {
```
DiscoElevator: Coul you please clarify what does it mean 'real data' in this context? Why not `fetchData`, `fetchAttachment` or similar?

app/src/controllers/attachments/sagas.js
```diff
@@ -0,0 +1,81 @@
...
+}
+
+/* IMAGE */
+
+function* openImageModalsWorker({ data: { projectId, binaryId } }) {
+  try {
+    const data = yield call(fetchImageData, { projectId, binaryId });
+    yield put(showModalAction({ id: 'attachmentImageModal', data }));
+  } catch (error) {
+    yield put({ type: ATTACHMENT_IMAGE_FETCH_ERROR, error });
```
DiscoElevator: 1. Please use action creators.
2. `payload: error`. If you want to mark this action as an error action, just set `error: true` (this is a common practice for FSA: https://github.com/redux-utilities/flux-standard-action#actions)

app/src/controllers/attachments/sagas.js
```diff
@@ -0,0 +1,81 @@
...
+  GET_ATTACHMENT_HAR,
+  ATTACHMENT_HAR_FETCH_ERROR,
+  GET_ATTACHMENT_BINARY,
+  ATTACHMENT_BINARY_FETCH_ERROR,
+  GET_ATTACHMENT_IMAGE,
+  ATTACHMENT_IMAGE_FETCH_ERROR,
+} from './constants';
+
+function fetchImageData({ projectId, binaryId }) {
+  return new Promise(() => ({
```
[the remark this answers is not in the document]
norbert-csaba-herczeg-epam: @DiscoElevator nice catch, ill add unit tests in return ;-)

app/src/controllers/attachments/sagas.js
norbert-csaba-herczeg-epam: @DiscoElevator that `catch` isn't even relevant, so I'll swap it's internals to a simple error log. `ATTACHMENT_IMAGE_FETCH_ERROR` is never consumed.

app/src/controllers/attachments/sagas.js
norbert-csaba-herczeg-epam: I'll rename it to `fetchData` as suggested. It was meant to distinguis actual calls which turn to a real service eventaually (not like the image type since it's returning a promise to comply with the API).

app/src/common/img/launch/ic_info.svg
```diff
@@ -0,0 +1,9 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="16" height="16" viewBox="0 0 16 16">
+  <defs>
+    <path id="ic_info-a" d="M7.33333333,10 L8.66666667,10 L8.66666667,11.3333333 L7.33333333,11.3333333 L7.33333333,10 Z M7.33333333,4.66666667 L8.66666667,4.66666667 L8.66666667,8.66666667 L7.33333333,8.66666667 L7.33333333,4.66666667 Z M7.99333333,1.33333333 C4.31333333,1.33333333 1.33333333,4.32 1.33333333,8 C1.33333333,11.68 4.31333333,14.6666667 7.99333333,14.6666667 C11.68,14.6666667 14.6666667,11.68 14.6666667,8 C14.6666667,4.32 11.68,1.33333333 7.99333333,1.33333333 L7.99333333,1.33333333 Z M8,13.3333333 C5.05333333,13.3333333 2.66666667,10.9466667 2.66666667,8 C2.66666667,5.05333333 5.05333333,2.66666667 8,2.66666667 C10.9466667,2.66666667 13.3333333,5.05333333 13.3333333,8 C13.3333333,10.9466667 10.9466667,13.3333333 8,13.3333333 L8,13.3333333 Z"/>
+  </defs>
+  <g fill="none" fill-rule="evenodd" transform="matrix(1 0 0 -1 0 16)">
+    <polygon points="0 0 16 0 16 16 0 16"/>
+    <use fill="#999999" xlink:href="#ic_info-a"/>
```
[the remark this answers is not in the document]
norbert-csaba-herczeg-epam: @DiscoElevator I forgot to go through these, will check them. In the meantime I'll remove unused icons as well, beacuase I noticed that there is a few of those too.

app/src/pages/inside/logsPage/logItemInfo/logItemInfoTabs/attachments/README.md
```diff
@@ -0,0 +1,46 @@
...
+
+## Input props
+
+- **attachments**: _array_, required
+- **projectId**: _string_, required
+- **openImageModal**: _function_
+- **openHarModal**: _function_
+- **openBinaryModal**: _function_
+
+The componenct receives `attachments` which is an array of attachments in the following
```
DiscoElevator: typo in 'component'

app/src/pages/inside/logsPage/logItemInfo/logItemInfoTabs/attachments/index.js
```diff
@@ -0,0 +1,3 @@
+export { Attachments } from './attachments';
+export { AttachmentsContainer } from './attachmentsContainer';
```
DiscoElevator: `export { AttachmentsContainer as Attachments } from './attachmentsContainer'`

The idea here is that we use only connected component outside. Presentational layer is needed only for testing/storybook in this case. That's why we don't want to export it from here.

app/src/pages/inside/logsPage/logItemInfo/logItemInfoTabs/attachments/attachments.scss
```diff
@@ -0,0 +1,73 @@
...
+    width: 80%;
+    &.hideMainArea :global .carousel-slider {
+      display: none;
+    }
+  
+    .preview-container {
+      line-height: 365px;
+      cursor: pointer;
+  
+      .preview-image {
```
DiscoElevator: No need to make such deep nesting in this case. And we don't need to reflect HTML structure in CSS rules in most cases. All class names are unique.
The only case for nesting is when we need to make some rules more specific.

norbert-csaba-herczeg-epam: @DiscoElevator I had to keep the `global` tree since we need to override the default style, and I didn't wanted to use `!important` in a top level class.

norbert-csaba-herczeg-epam:
@DiscoElevator Updated with your suggestions.

app/src/controllers/attachments/actionCreators.js
```diff
@@ -1,16 +1,16 @@
 import { GET_ATTACHMENT_IMAGE, GET_ATTACHMENT_HAR, GET_ATTACHMENT_BINARY } from './constants';
 
-export const openImageModal = (payload) => ({
+export const openImageModalActin = (payload) => ({
```
DiscoElevator: Acti**o**n

norbert-csaba-herczeg-epam: @DiscoElevator fixed

norbert-csaba-herczeg-epam:
@DiscoElevator updated the `AttachmentsContainer` export to the suggested alias.

PR #1013, opened 2018-10-26
EPMRPP-36893 reproducible builds

norbert-csaba-herczeg-epam:
Dependency versions have been fixed.

The contents of `package-lock.json` may still slightly be modified on local `npm install`, see: [package-lock.json and optional packages](https://github.com/npm/npm/issues/17722). This means that additional meta information may be added, or removed which does not affect actual dependency versions.

I also had to fix a few components because they didn't pass the linter rules.

The `Makefile` has been updated to use `npm ci` instead of `npm install`. Detailed information can be found in the [official docs](https://docs.npmjs.com/cli/ci#description).


norbert-csaba-herczeg-epam:
@DiscoElevator eslint ignore-s were added inline because I didn't wanted to add a global exception to the rules since not all Labels will be rendered dynamically.

app/package.json
```diff
@@ -21,120 +21,120 @@
...
+    "html-react-parser": "0.4.5",
+    "intl": "1.2.5",
+    "moment": "2.22.1",
+    "promise-polyfill": "8.1.0",
+    "perf-cascade": "2.5.4",
+    "prop-types": "15.6.1",
+    "qhistory": "1.0.0",
+    "qs": "6.5.2",
+    "rc-slider": "8.6.3",
+    "react": "16.3.2",
```
DiscoElevator: Please also updare react to the latest version

norbert-csaba-herczeg-epam:
@DiscoElevator updated.

PR #1015, opened 2018-10-29
EPMRPP-36911 React application fails to load

norbert-csaba-herczeg-epam:
The `core-js` transitive dependency has been hoisted with an older version (`v1.2.7`) from multiple versions which caused the app to crash at runtime. I couldn't figure it out why NPM chose the oldest version.

Since there were multiple references to `core-js` in the `package-lock.json` file both with `dev` flag and without, I chose to add it as a simple dependency.


norbert-csaba-herczeg-epam:
@DiscoElevator This is somewhat tied to #1013 (packages) so in case either gets merged, I believe the other one will need an update.

PR #1022, opened 2018-11-01
EPMRPP-36738. Fix filter query value for testItem pages & Add interrupted test type for history line

AmsterGet:


app/src/controllers/testItem/sagas.js
```diff
@@ -96,7 +96,9 @@ function* fetchTestItems() {
         'filter.eq.launch': launchId,
         'filter.eq.parent': !noChildFilter ? parentId : undefined,
         'filter.level.path': sizePath,
-        'filter.under.path': noChildFilter ? parentId : undefined,
+        'filter.under.path': noChildFilter
+          ? itemIds.filter((item) => item !== launchId.toString()).join('.')
```
DiscoElevator: Is it necessary to use `toString` here?

AmsterGet: `itemIds` is represented as an array of strings, but `launchId` may be a `Number` according to new API

DiscoElevator: Maybe we should convert all ids to number in selectors to reflect API changes?

AmsterGet: We get `itemIds` from `location` in `state`
Should I convert this items to numbers in appropriate selector?
image

AmsterGet: Done

PR #1033, opened 2018-11-09
EPMRPP-36796-log-level-debug-level-transfers-to-launch-log-level

tamas-levente-szucs-epam:
screenshot 15


app/src/controllers/testItem/utils.js
```diff
@@ -28,9 +29,17 @@ export const calculateLevel = (data) =>
 
 export const getQueryNamespace = (levelIndex) => `item${levelIndex}`;
 
-export const getNextPage = (currentLevel, debugMode) => {
-  if (currentLevel === launchLevels.LEVEL_STEP) {
+export const getNextPage = (currentLevel, debugMode, currentType) => {
+  if (
```
DiscoElevator: Can we just do the following:
```
if (currentLevel === launchLevels.LEVEL_STEP) {
  debugMode ? PROJECT_USERDEBUG_LOG_PAGE : PROJECT_LOG_PAGE;
}
```

tamas-levente-szucs-epam: If i understand it clearly, this debugmode working only on the previous level... before this problem.

DiscoElevator:
Closed in favor of #1260

PR #1034, opened 2018-11-09
EPMRPP-37037: fix launch level defect type filters selection

DiscoElevator:


app/src/pages/inside/common/filterEntitiesGroups/launchLevelEntities.jsx
```diff
@@ -360,7 +361,9 @@ export class LaunchLevelEntities extends Component {
             title: `${this.props.intl.formatMessage(messages[`${defectTypeRef}_title`])} ${
```
AmsterGet: I think in this method we also need to add value check for each item like in `getStaticEntities`
```
value:
           STATS_SKIPPED in filterValues
             ? filterValues[STATS_SKIPPED]
             : {
                 value: '',
                 condition: CONDITION_GREATER_EQ,
               },
```

PR #1044, opened 2018-11-16
EPMRPP-33456 fix styles for tooltips

andrei-bulagin:
use common styles as default tooltipWrapper theme

### DIFF
--- a/app/src/components/widgets/charts/launchesDurationChart/tooltip.jsx
+++ b/app/src/components/widgets/charts/launchesDurationChart/tooltip.jsx
@@ -1,22 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { duration } from 'moment';
+import classNames from 'classnames/bind';
 
 import { isValueInterrupted } from './chartUtils';
+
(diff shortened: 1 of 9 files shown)


app/src/components/widgets/charts/launchesDurationChart/tooltip.jsx
```diff
@@ -1,22 +1,27 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { duration } from 'moment';
+import classNames from 'classnames/bind';
 
 import { isValueInterrupted } from './chartUtils';
 
+import styles from './launchesDurationChart.scss';
```
DiscoElevator: Please make a new file `launchDurationTooltip.scss` and put corresponding styles there.
Also, this file should be renamed according to the component's name.
```
launchesDurationChart
  |-launchDurationTooltip
    |-launchDurationTooltip.jsx
    |-launchDurationTooltip.scss
    |-index.js
```

andrei-bulagin: Done

PR #1060, opened 2018-11-26
EPMRPP-36979 REACT. Filter bar is displayed only on Launches view

andrei-bulagin:
- add Top filter bar to Suite, Step and Test pages
- move LaunchFiltersToolbar to common components


app/src/pages/inside/common/launchFiltersSection/launchFiltersSection.jsx
```diff
@@ -0,0 +1,42 @@
...
+      activeFilterId,
+      activeFilter,
+      activeFilterEntities,
+      onSelectFilter,
+      onRemoveFilter,
+      onChangeFilter,
+      onResetFilter,
+    }) => (
+      <FilterEntitiesContainer
+        level={level}
```
DiscoElevator: This section contains only launch level filters. No need to receive it from props.
`level={LEVEL_LAUNCH}`

andrei-bulagin: Removed

PR #1061, opened 2018-11-27
EPMRPP-36942 Fix missed header on Admin / Server Settings

andrei-bulagin:


app/src/pages/admin/serverSettingsPage/serverSettingsPage.jsx
```diff
@@ -1,12 +1,35 @@
...
+export class ServerSettingsPage extends Component {
+  static propTypes = {
+    intl: intlShape.isRequired,
+  };
+
+  render() {
+    return (
+      <PageLayout>
+        <PageHeader
+          breadcrumbs={[
```
DiscoElevator: Just a little suggestion: make method `getBreadcrumbs` which returns this array. Just to make JSX look less bloated.

andrei-bulagin: Done

PR #1063, opened 2018-11-28
EPMRPP-37076 Fix filter fields

andrei-bulagin:


app/src/controllers/filter/constants.js
```diff
@@ -26,7 +26,7 @@ export const DEFAULT_FILTER = {
   entities: [],
   type: 'launch',
   orders: [
-    { is_asc: false, sorting_column: 'start_time' },
-    { is_asc: false, sorting_column: 'number' },
+    { isAsc: false, sortingColumn: 'start_time' },
```
DiscoElevator: maybe `startTime`?

andrei-bulagin: Yes, it should be another fix for Launch fields (in separate Pull Request)

PR #1065, opened 2018-11-29
EPMRPP-35466: all/latest launches selector

DiscoElevator:


app/src/controllers/launch/selectors.js
```diff
@@ -24,3 +31,15 @@ export const queryParametersSelector = createQueryParametersSelector({
 });
 
 export const debugModeSelector = (state) => domainSelector(state).debugMode || false;
+
+const createLaunchesLinkSelector = (filterId) =>
+  createSelector(projectIdSelector, payloadSelector, (activeProject, payload) => ({
```
AmsterGet: Are you really need `projectIdSelector` here?

app/src/pages/inside/launchesPage/launchFiltersToolbar/allLatestDropdown/allLatestDropdown.scss
```diff
@@ -0,0 +1,118 @@
...
+  font-size: 13px;
+  text-align: center;
+  text-transform: uppercase;
+
+  @media (max-width: $SCREEN_SM_MAX) {
+    display: none;
+  }
+}
+
+.value-short {
```
AmsterGet: image
May be we need to add some `padding` here

app/src/controllers/launch/selectors.js
DiscoElevator: ohh, really! It's already in the payload

PR #1076, opened 2018-12-05
Add stylelint. Fix stylelint errors.

DiscoElevator:
Things to do in future (not in the scope of this PR):
1. remove warnings
2. replace colors marked with `stylelint-disable-line` with appropriate ones

### DIFF
--- a/app/src/components/main/modal/modalLayout/modalLayout.scss
+++ b/app/src/components/main/modal/modalLayout/modalLayout.scss
@@ -20,7 +20,7 @@
   display: block;
   width: 100%;
   height: 100%;
-  background-color: #000000;
+  background-color: #000000; /* stylelint-disable-line color-no-hex */
   opacity: 0.3;
   transition: opacity 300ms ease-in;
   z-index: 1;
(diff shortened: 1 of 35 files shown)


app/src/components/main/modal/modalLayout/modalLayout.scss
```diff
@@ -20,7 +20,7 @@
   display: block;
   width: 100%;
   height: 100%;
-  background-color: #000000;
+  background-color: #000000; /* stylelint-disable-line color-no-hex */
```
ultd39: Why not short format like `#000`?

ultd39:
Great work! I think I'll also use your solution for style linting in TDM project :)

app/src/components/main/modal/modalLayout/modalLayout.scss
DiscoElevator: The most important question it why there is a color which is not in the palette.  All lines marked with `stylelint-disable-line` will be investigated in the future.

PR #1079, opened 2018-12-06
EPMRPP-36884  REACT. Step view. Start time filter options have incorrect names

istvan-kelemen-epam:
dates

### DIFF
--- a/app/src/components/filterEntities/entityItemStartTime.jsx
+++ b/app/src/components/filterEntities/entityItemStartTime.jsx
@@ -34,7 +34,7 @@ const presets = [
     },
   },
   {
-    label: 'Last week',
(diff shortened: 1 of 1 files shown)


app/src/components/filterEntities/entityItemStartTime.jsx
```diff
@@ -34,7 +34,7 @@ const presets = [
     },
   },
   {
-    label: 'Last week',
+    label: 'Last 7 days',
```
DiscoElevator: Please also add localization

DiscoElevator: Will be implemented in another task

PR #1088, opened 2018-12-10
EPMRPP-35247: React: Long filter name is overlapped the delete icon mobile view

istvan-kelemen-epam:
localhost_3000_


istvan-kelemen-epam:
https://jiraeu.epam.com/browse/EPMRPP-35247

PR #1091, opened 2018-12-12
EPMRPP-37186 React. Adjust Email notification component

AmsterGet:


app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/forms/notificationRuleForm/validation.js
```diff
@@ -0,0 +1,47 @@
...
+    ruleOne.launchNameRule.sort().every((launch) => ruleTwo.launchNameRule.includes(launch)) &&
+    ruleOne.recipients.sort().every((recipient) => ruleTwo.recipients.includes(recipient))
+  );
+};
+
+export const validate = ({ rules }) => {
+  const errors = {};
+  const rulesArrayErrors = [];
+
+  rules.forEach((item, index, arr) => {
```
DiscoElevator: This function is difficult to understand (see comments below). I'm not sure, but will my version be correct too?
```
const rulesErrors = rules.map((rule, index) => {
  if (rules.length > 1) {
    const hasDuplicates = arr.find((item, i) => compareRules(item, rule) && i !== index);
    if (hasDuplicates) {
      return { confirmed: 'hasDuplicates' };
    }
  }
  if (rule.recipients.length === 0 && !rule.informOwner) {
    return {
      recipients: 'recipientsHint',
      confirmed: 'error',
    };
  }
  return null;
}).filter(Boolean);
```

app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/notificationsTab.jsx
```diff
@@ -19,33 +19,34 @@ const cx = classNames.bind(styles);
   (state) => ({
     projectRole: activeProjectRoleSelector(state),
     userRole: userAccountRoleSelector(state),
-    emailEnabled: projectEmailEnabledSelector(state),
-    emailCases: projectEmailCasesSelector(state),
+    enabled: notificationIntegrationSelector(state).enabled,
```
DiscoElevator: Please add corresponding selector to incapsulate name and position of this property.

app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/notificationsTab.jsx
```diff
@@ -55,38 +56,51 @@ export class NotificationsTab extends Component {
...
+
   isAbleToEditForm = () =>
     canConfigreEmailNotifications(this.props.userRole, this.props.projectRole);
-  isAbleToSubmit = ({ emailCases }) => emailCases.every(({ confirmed }) => confirmed);
-  convertEmailCaseForSubmission = (obj) => {
-    const { informOwner, launchNames, recipients, sendCase, tags } = obj;
+
+  isAbleToSubmit = ({ rules }) => rules.every(({ confirmed }) => confirmed);
+
+  convertRulesForSubmission = (obj) => {
```
DiscoElevator: `convertRuleForSubmission ` - looks like this function converts only one rule

app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/forms/notificationRuleForm/validation.js
AmsterGet: But these errors in different if blocks can be found at the same time. So, I use your solution with a few updates.

app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/notificationsTab.jsx
AmsterGet: Fixed

app/src/pages/inside/settingsPage/settingsTabs/notificationsTab/forms/notificationRuleForm/validation.js
AmsterGet: I just change the order of these conditions according to their priority

PR #1096, opened 2018-12-14
EPMRPP-37291: Fix All launches and a Filter tab highlighting

Shyscarecrow:


Shyscarecrow:
Closed in favor of #1100

PR #1101, opened 2018-12-14
EPMRPP-34186: React: No special view for case with no results found

akos-toth-epam:


app/src/pages/inside/filtersPage/filtersPage.scss
```diff
@@ -0,0 +1,47 @@
...
+  margin: 0 0 10px;
+  color: $COLOR--gray-47;
+  align-items: center;
+  justify-content: center;
+  font-family: $FONT-SEMIBOLD;
+  font-size: 16px;
+  font-weight: 600;
+  line-height: 22px;
+
+  span {
```
ultd39: Please avoid to use tag selectors.

app/src/pages/inside/filtersPage/filtersPage.scss
```diff
@@ -0,0 +1,47 @@
...
+    max-width: 200px;
+    font-family: $FONT-REGULAR;
+    font-weight: bold;
```
ultd39: @DiscoElevator, as I remember we shouldn't use font-weight because we have special bold font for that?

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -207,6 +220,18 @@ export class LaunchesPage extends Component {
     this.props.unselectAllLaunchesAction();
   }
 
+  getNoItemMessage = () => (
```
ultd39: Better to rename this method to `renderNoItemMessage` because it return some markup but not a value or something.

app/src/pages/inside/membersPage/membersPage.jsx
```diff
@@ -99,6 +112,28 @@ export class MembersPage extends Component {
...
+              }),
+            )}
+          </p>
+          <p className={cx('filter-not-found-hint')}>
+            {this.props.intl.formatMessage(messages.checkQuery)}
+          </p>
+        </div>
+      );
+    }
+    return output;
```
ultd39: Is some cases this function returns empty string or node? I guess check `(this.props.filter !== null && this.props.filter !== '')` should be somewhere outside and method should always return node.

app/src/pages/inside/membersPage/membersPage.scss
```diff
@@ -0,0 +1,47 @@
...
+  margin: 0 0 10px;
+  color: $COLOR--gray-47;
+  align-items: center;
+  justify-content: center;
+  font-family: $FONT-SEMIBOLD;
+  font-size: 16px;
+  font-weight: 600;
+  line-height: 22px;
+
+  span {
```
ultd39: Please avoid to use tag selectors.

istvan-kelemen-epam:
New PR:
https://github.com/reportportal/service-ui/pull/1129

DiscoElevator:
Closed in favor of #1129

PR #1117, opened 2019-01-04
fix switcherOn localization in BE

richard-jakab-epam:


app/localization/translated/be.json
```diff
@@ -245,7 +245,7 @@
   "DemoDataTabForm.preloaderInfo": "Генераванне дадзеных пачалося. Працэс можа заняць некалькі хвілін, калі ласка, пачакайце.",
   "DisplayFilter.display": "Адлюстраваць на запусках:",
   "DisplayFilter.showOnLaunchesSwitcherOff": "Выкл",
-  "DisplayFilter.showOnLaunchesSwitcherOn": "Вкл",
+  "DisplayFilter.showOnLaunchesSwitcherOn": "УКЛ",
```
DiscoElevator: Please write as `Укл`. Transform to uppercase is done by the corresponding CSS.

PR #1125, opened 2019-01-09
pencil icon visibility management sovled (EPMRPP-37380-edit-icon-show)

marton-horvath-epam:


app/src/pages/inside/dashboardItemPage/widgetWizardModal/widgetControls/controls/filtersControl/filtersItem/filtersItem.jsx
```diff
@@ -41,13 +43,20 @@ export class FiltersItem extends PureComponent {
           onChange={onChange}
           circleAtTop
         >
-          <FilterName search={search} filter={filter} userId={userId} showDesc={false} />
+          <FilterName
+            search={search}
+            filter={filter}
+            userId={userId}
+            showDesc={false}
+            editable={false}
```
ultd39: It's isn't mistake? `editable` here always should be `false`?

marton-horvath-epam: Yes, you are right, i have corrected it. Thanks for the comment :)

app/src/pages/inside/dashboardItemPage/widgetWizardModal/widgetControls/controls/filtersControl/filtersItem/filtersItem.jsx
```diff
@@ -41,13 +43,20 @@ export class FiltersItem extends PureComponent {
...
+            editable={editable}
+          />
           <FilterOptions entities={filter.conditions} sort={filter.orders}>
-            {userId === filter.owner && (
-              <span className={cx('pencil-icon')} onClick={onEdit}>
-                {Parser(PencilIcon)}
-              </span>
-            )}
+            {userId === filter.owner &&
+              editable && (
```
DiscoElevator: I don't see any place where this flag is passed to the component. Are all filters unavailable for edit in this modal now?
Looks like a mistake.

marton-horvath-epam: I have checked already and i have seen that filtersItem only called as part of FiltersControl. The FiltersControl called only from the widgetControl components, so based on the usage (we dont need the edit icon) i set the "editable" option visibility false as default. I think it is more logical, cause an item can be occurred at many place, but in an editable form it should be occurred only one place. So to avoid more code (give editable value as an extra parameter) i prefer my solution as default the "filtersItem" is not editable.

DiscoElevator: >  so based on the usage (we dont need the edit icon)
If current user is the owner of this filter, this filter should be editable.

The bug is that we have 2 "edit" icons: one in the name row (`filterName.jsx`), and one in the options row (in this file). Your change in the `filterName.jsx` is enough to hide the unneeded icon. Don't hide the second one - it works correctly.

marton-horvath-epam: OK, seems that part wasn't clear for me, thanks, i will correct it.

app/src/pages/inside/dashboardItemPage/widgetWizardModal/widgetControls/controls/filtersControl/filtersItem/filtersItem.jsx
```diff
@@ -18,6 +18,7 @@ export class FiltersItem extends PureComponent {
     filter: PropTypes.object.isRequired,
     onChange: PropTypes.func.isRequired,
     onEdit: PropTypes.func.isRequired,
+    editable: PropTypes.bool,
```
DiscoElevator: I think this prop is unnecessary now.
`editable` state is controlled by the component itself (see https://github.com/reportportal/service-ui/pull/1125/files#diff-232b5b52b9b43ec5f2915609e54ad5dbR54)

Just setting `editable={false}` on line 51 (https://github.com/reportportal/service-ui/pull/1125/files#diff-232b5b52b9b43ec5f2915609e54ad5dbR51) won't be a mistake.

app/src/pages/inside/dashboardItemPage/widgetWizardModal/widgetControls/controls/filtersControl/filtersItem/filtersItem.jsx
DiscoElevator: Not a mistake. We need to hide "edit" icon for this component unconditionally only in this case.

PR #1127, opened 2019-01-10
EPMRPP-37597: The attribute icon is no more shown for launch without attributes

Shyscarecrow:


app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.jsx
```diff
@@ -7,7 +7,7 @@ const cx = classNames.bind(styles);
 
 export const AttributesBlock = ({ attributes }) => (
   <div className={cx('attributes-block')}>
-    <div className={cx('attributes-icon')} />
+    {attributes.filter(notSystemAttributePredicate) && <div className={cx('attributes-icon')} />}
```
DiscoElevator: `attributes.some(notSystemAttributePredicate)`
`filter` returns an array which is always converted to boolean as `true`.

PR #1128, opened 2019-01-10
Epmrpp 37340 extra share icon on filters

tamas-deli-epam:
Extra icon removed.

image

https://jiraeu.epam.com/browse/EPMRPP-37340

### DIFF
--- a/app/src/pages/inside/filtersPage/filterGrid/filterName/filterName.jsx
+++ b/app/src/pages/inside/filtersPage/filterGrid/filterName/filterName.jsx
@@ -4,7 +4,6 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Parser from 'html-react-parser';
 
-import ShareIcon from 'common/img/share-icon-inline.svg';
 import PencilIcon from 'common/img/pencil-icon-inline.svg';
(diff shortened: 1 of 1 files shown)


AmsterGet:
Please, change the base branch to `reportportal:v5`

app/src/pages/inside/filtersPage/filterGrid/filterName/filterName.jsx
```diff
@@ -71,12 +70,6 @@ export class FilterName extends Component {
           >
             {Parser(this.getHighlightName(filter.name))}
           </span>
-          {filter.share &&
```
DiscoElevator: Should it be deleted?
I still see this icon in the designs: https://zpl.io/VOo11R8
However, on RP 4 (port 8080) there are only globe icons in this modal. Please clarify with BA/UX designer which icons should be visible on the step 2 of the widget creation modal.

DiscoElevator: I've just discussed this question with the team: these icons should be hidden only on filters page. In other places one of these icons should be visible (depending on owner).
Maybe it's better to introduce a prop which hides both icons.

DiscoElevator:
Closed in favor of #1252

PR #1133, opened 2019-01-14
EPMRPP-37586 OX axis visibility fix in launchesDuration chart

richard-jakab-epam:
EPMRPP-37586 -  React: No OX axis is shown on the widget

ox

### DIFF
--- a/app/src/components/widgets/charts/launchesDurationChart/launchesDurationChart.scss
+++ b/app/src/components/widgets/charts/launchesDurationChart/launchesDurationChart.scss
@@ -7,7 +7,7 @@
         .c3-axis-y-label {
           font-size: 12px;
         }
-        path {
+        &.c3-axis-y path.domain {
           display: inline;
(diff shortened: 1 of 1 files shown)


DiscoElevator:
Are there any task for this? If yes, please specify task id in the PR name and in the commit message.

app/src/components/widgets/charts/launchesDurationChart/launchesDurationChart.scss
```diff
@@ -7,6 +7,10 @@
         .c3-axis-y-label {
           font-size: 12px;
         }
+        path {
```
AmsterGet: Maybe it would be better if we indicate which `path` we should highlight? F.e. by `.domain` selector

AmsterGet: Maybe it would be better if we indicate which `path` we should highlight? F.e. by `.domain` selector

PR #1141, opened 2019-01-16
EPMRPP-37006: Admin part. Event monitoring. Layout

Shyscarecrow:


app/src/controllers/administrate/events/constants.js
```diff
@@ -0,0 +1,3 @@
+export const FETCH_EVENTS = 'fetchEvents';
+export const NAMESPACE = 'events';
+export const DEFAULT_PAGE_SIZE = 50;
```
DiscoElevator: Is it used now? If not, please remove.

PR #1144, opened 2019-01-21
EPMRPP-36981 - REACT. No filters are added by clicking on filter icons on the left of column headers 

richard-jakab-epam:
EPMRPP-36981  -  No filters are added by clicking on filter icons on the left of column headers

### DIFF
--- a/app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
+++ b/app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
@@ -1,5 +1,6 @@
 import { Component } from 'react';
 import PropTypes from 'prop-types';
+import isEqual from 'fast-deep-equal';
 import { connect } from 'react-redux';
 import {
   launchFiltersSelector,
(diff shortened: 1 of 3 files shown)


app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -81,28 +86,37 @@ export class LaunchFiltersContainer extends Component {
...
+      this.props.createFilter();
+    }
   };
 
   handleFilterRemove = (filter) => {
     if (filter.id >= 0) {
       this.props.hideFilterOnLaunchesAction(filter);
     }
     this.props.removeLaunchesFilterAction(filter.id);
+    this.fetchLaunches()();
```
DiscoElevator: You're calling fetch and immediately canceling it? Looks like this line does nothing.
Also, is it necessary to update launches list here? It is already triggered when we remove a filter.

app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -81,28 +86,37 @@ export class LaunchFiltersContainer extends Component {
...
+      this.fetchLaunches(newFilter);
+    }
+
+    if (this.props.activeFilter) {
+      this.updateFilter(
+        this.props.activeFilterId,
+        Object.keys(conditions).map((key) => ({ ...conditions[key] })),
+      );
+    } else {
+      this.props.createFilter();
```
DiscoElevator: Please pass conditions here to avoid inconsistent state (these conditions are not reflected in redux state, only someshere in component's state).

DiscoElevator:
Closed in favor of #1230

PR #1145, opened 2019-01-21
bugfix/EPMRPP-37573 User photo is not applied

AmsterGet:


app/src/controllers/user/actionCreators.js
```diff
@@ -13,6 +14,12 @@ const fetchUserSuccessAction = (user) => ({
   payload: user,
 });
 
+export const setPhotoTimeStampAction = (timeStamp) => (dispatch) =>
```
DiscoElevator:
```
export const setPhotoTimeStampAction = (timeStamp) => ({
    type: SET_PHOTO_TIME_STAMP,
    payload: timeStamp,
  });
```

app/src/controllers/user/constants.js
```diff
@@ -1,12 +1,14 @@
...
 export const SET_ACTIVE_PROJECT = 'setActiveProject';
 export const SET_START_TIME_FORMAT = 'setStartTimeFormat';
+export const SET_PHOTO_TIME_STAMP = 'setPhotoTimeStamp';
 
 export const START_TIME_FORMAT_RELATIVE = 'relative';
 export const START_TIME_FORMAT_ABSOLUTE = 'absolute';
 
 export const SETTINGS_INITIAL_STATE = {
   startTimeFormat: START_TIME_FORMAT_RELATIVE,
+  photoTimeStamp: null,
```
DiscoElevator: Maybe it should be initalized as `Date.noew()`?

AmsterGet: Fixed

app/src/controllers/user/actionCreators.js
AmsterGet: Fixed

PR #1150, opened 2019-01-23
EPMRPP-37758: Event monitoring. Fix pagination

Shyscarecrow:


app/src/controllers/administrate/events/selectors.js
```diff
@@ -10,4 +11,5 @@ export const loadingSelector = (state) => domainSelector(state).loading || false
 
 export const querySelector = createQueryParametersSelector({
   defaultPagination: DEFAULT_PAGINATION,
+  namespace: NAMESPACE,
```
DiscoElevator: Namespace for query parameters are used for nested routes.
This bug should be fixed in another way.
The problem is that after switching page we get `store.location['page'size'] === undefined` (the key still exist in the store, but with value `undefined` - it is strange), which is not replaced by default value in the selector. We need to investigate the reason of this behavior and fix it.

DiscoElevator:
Closed in favor of #1163

PR #1165, opened 2019-02-01
EPMRPP-37404_Request_on_GET_test_items_has_comma_at_the_beginning

richard-jakab-epam:
EPMRPP-37404
React: Request on GET test items has comma (,) at the beginning in case of filtering by status


app/src/components/filterEntities/entityDropdown.jsx
```diff
@@ -34,6 +34,14 @@ export class EntityDropdown extends Component {
 
   render() {
     const { value, onRemove, removable, entityId, smallSize, title, meta } = this.props;
+    let dropDownValue;
```
DiscoElevator: Please move this logic to a class method `getValue`

PR #1170, opened 2019-02-05
EPMRPP-34391: tag shows long launch name now

csigataraj:
fixed long name bug

### DIFF
--- a/app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
+++ b/app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
@@ -1,4 +1,5 @@
 .selected-item {
+  min-height: 20px;
   background-color: $COLOR--white;
   border: 1px solid $COLOR--gray-80;
   font-size: 12px;
--- a/app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
+++ b/app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
@@ -1,5 +1,4 @@
 .selected-item {
-  // height: 20px;
   background-color: $COLOR--white;
   border: 1px solid $COLOR--gray-80;
   font-size: 12px;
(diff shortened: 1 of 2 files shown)


app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
```diff
@@ -1,5 +1,5 @@
 .selected-item {
-  height: 20px;
+  // height: 20px;
```
AmsterGet: Please, remove unnecessary comments

csigataraj: removed

app/src/pages/inside/common/selectedItems/selectedItems.scss
```diff
@@ -3,6 +3,8 @@
   padding-top: 15px;
   padding-left: 15px;
   padding-right: 15px;
+  word-break: break-all;
+  word-break: break-word;
```
DiscoElevator: Let's use `@supports` directive to make sure every browser will use correct style:
```
  word-break: break-all;

    word-break: break-word
  }
```

app/src/pages/inside/common/selectedItems/selectedItem/selectedItem.scss
```diff
@@ -1,5 +1,4 @@
 .selected-item {
-  height: 20px;
```
DiscoElevator: Maybe it's better to leave this as `min-height`?

csigataraj: have put it back as min-height

app/src/pages/inside/common/selectedItems/selectedItems.scss
csigataraj: replaced it as suggested

app/src/pages/inside/common/selectedItems/selectedItems.scss
```diff
@@ -3,6 +3,11 @@
   padding-top: 15px;
   padding-left: 15px;
   padding-right: 15px;
+  word-break: break-all;
+
+  @supports (word-break: break-word) {
+    word-break: break-word
```
DiscoElevator: Please add missing semicolon

csigataraj: added

PR #1171, opened 2019-02-05
EPMRPP-35499: [REACT] Body of item's page is not loaded after removing item

istvan-kelemen-epam:


app/src/pages/inside/testItemPage/notFound.jsx
```diff
@@ -0,0 +1,27 @@
...
+  },
+  notFound: {
+    id: 'LaunchesPage.notFound',
+    defaultMessage: 'Launch is not found',
+  },
+});
+
+export const NotFound = injectIntl(({ intl: { formatMessage } }) => (
+  <div className={cx('launch-not-found')}>
+    <p className={cx('launch-not-found-text')}>
```
AmsterGet: You can use here `NoItemMessage` component from `components/main/noItemMessage`

istvan-kelemen-epam: It has only one message parameter, but the requirement is 2 texts.

image-epmrpp-35499

istvan-kelemen-epam: Fixed

PR #1178, opened 2019-02-06
Epmrpp 37903 no tooltip appears for the filter desc icon

tamas-deli-epam:
Tooltip added to the filters on the launches page.


DiscoElevator:
Please change branch from `master` to `v5`

tamas-deli-epam:
I will open a new pull requset because the target branch is the master.

PR #1180, opened 2019-02-07
EPMRPP-36486: Tags are not clickable now

tamas-deli-epam:


tamas-deli-epam:
Jira ID: EPMRPP-36486
Link: https://jiraeu.epam.com/browse/EPMRPP-36486

PR #1183, opened 2019-02-11
V5 fix for EPMRPP-37182: Too many redirects err in console view.

tritseratops:
fix for EPMRPP-37182: Too many redirects err in console view.


DiscoElevator:
Closed in favor of #1188

PR #1194, opened 2019-02-14
EPMRPP-37946: Implement adding attributes to filters by click on attr…

DiscoElevator:
…ibute


DiscoElevator:
> The following unexpected state of the filter after clicking on the attribute (see the screenshot below).
> image

Should be fixed in #1144

PR #1199, opened 2019-02-15
EPMRPP-37980 moved tooltip message above the input field

csigataraj:
fixed the bug where the tooltip message overshadows the input field

### DIFF
--- a/app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
+++ b/app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
@@ -23,10 +23,6 @@
   &.type-top {
     bottom: 100%;
   }
-
-  & .type-top {
-    top: -30px;
-  }
 }
 
 .hint-content {
@@ -46,6 +42,7 @@
 
   &.type-top {
     transform: translateY(-100%);
+    top: -30px;
   }
 
   &.static-hint {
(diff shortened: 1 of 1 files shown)


app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
```diff
@@ -11,6 +11,7 @@
   left: 0;
   width: 100%;
   height: 0;
+  top: 0;
```
DiscoElevator: This change breaks other fields with hints. Check login page for examples.

app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
```diff
@@ -23,6 +23,10 @@
   &.type-top {
     bottom: 100%;
   }
+
+  & .type-top {
```
DiscoElevator: There's no point in using nested selector because there's already a special class for target element (`.hint-content`). Please move this style (without selector) below the line 48.

csigataraj: But wouldn't that affect they bottom type ones as well? (Same issue as the first iteration of this PR)

app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
csigataraj: fixed

app/src/components/fields/fieldErrorHint/fieldErrorHint.scss
csigataraj: fixed

PR #1205, opened 2019-02-20
fix for EPMRPP-35606 React: Item name in modals should be enclosed in…

tritseratops:
…to single quotes


tritseratops:
Added  fixes for EPMRPP-35606 and EPMRPP-35148

PR #1213, opened 2019-02-26
EPMRPP-37975: All Users. Fix Add user button icon

Shyscarecrow:


app/src/pages/admin/allUsersPage/usersToolbar/actionPanel/actionPanel.scss
```diff
@@ -7,5 +7,10 @@
       width: 20px;
       height: 20px;
     }
+    &:last-child {
+      svg {
```
DiscoElevator: Please find another way instead of manipulating button styles from outside. Maybe it can be fixed inside of the button component or inside of the svg file.

PR #1218, opened 2019-02-27
bugfix/EPMRPP-38119 Widget modal. Criteria for widget

AmsterGet:


app/src/components/inputs/inputDropdown/inputDropdown.jsx
```diff
@@ -75,6 +77,13 @@ export class InputDropdown extends Component {
     this.node = node;
   };
 
+  getOptionChangeHandler = (option) =>
+    (!option.disabled &&
```
DiscoElevator: I think that using `if..else` will be more verbose, but more readable here.

AmsterGet: Fixed

PR #1227, opened 2019-03-04
EPMRPP-37342: fixed, disable drop-zone while uploading

mykhailo-kalashnikov-epam:


app/src/pages/inside/launchesPage/modals/launchImportModal/launchImportModal.jsx
```diff
@@ -162,7 +162,15 @@ export class LaunchImportModal extends Component {
       ),
     };
   };
-
+  get validFiles() {
```
DiscoElevator: I think it's better to use functions instead of field getters:
- it will be harder to make a mistake in name (calling unexisting function will thow an exception)
- using functions for computed data is less confusing (when accessing field it's not clear if this field is readonly, or not)

app/src/pages/inside/launchesPage/modals/launchImportModal/launchImportModal.jsx
```diff
@@ -162,7 +162,10 @@ export class LaunchImportModal extends Component {
       ),
     };
   };
-
+  validFiles = () => this.state.files.filter(({ valid }) => valid);
```
DiscoElevator: getValidFiles

app/src/pages/inside/launchesPage/modals/launchImportModal/launchImportModal.jsx
```diff
@@ -162,7 +162,10 @@ export class LaunchImportModal extends Component {
...
+  uploadInProgress = () => this.validFiles().some(({ isLoading }) => isLoading);
```
DiscoElevator: isUploadInProgress

app/src/pages/inside/launchesPage/modals/launchImportModal/launchImportModal.jsx
```diff
@@ -162,7 +162,10 @@ export class LaunchImportModal extends Component {
...
+  uploadFinished = () =>
```
DiscoElevator: isUploadFinished

PR #1229, opened 2019-03-05
EPMRPP-38172: Fix long user name

Shyscarecrow:


app/src/pages/admin/allUsersPage/allUsersGrid/nameColumn/nameColumn.scss
```diff
@@ -20,6 +20,16 @@
     margin-top: 15px;
   }
 }
+.name {
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
DiscoElevator: On v4 if user name is too long, it breaks into multiple rows. Do we need to change this behavior in v5?

Shyscarecrow: Our design team recommends to cut long names instead of breaking into rows

PR #1230, opened 2019-03-05
EPMRPP-36981: No filters are added by clicking on filter icons on the left of column headers

DiscoElevator:


app/src/controllers/launch/sagas.js
```diff
@@ -24,12 +24,15 @@ function* fetchLaunchesWithParams({ payload }) {
   );
 }
 
+const notEmptyConditionsPredicate = ({ value }) =>
```
AmsterGet: It is similar with [this](https://github.com/reportportal/service-ui/pull/1230/files#diff-31c78ba55f47f72427159c7b26a8fdf3R18)
May be better to create `.utils.js` and hold it there also with [this](https://github.com/reportportal/service-ui/pull/1230/files#diff-31c78ba55f47f72427159c7b26a8fdf3R20)

DiscoElevator: fixed

PR #1237, opened 2019-03-08
EPMRPP-37670 Add sorting in filter on Launches page

vitalii-chornyi:


app/src/pages/inside/common/launchFiltersToolbar/filtersActionBar/filtersActionBar.jsx
```diff
@@ -3,6 +3,7 @@ import { FormattedMessage } from 'react-intl';
 import classNames from 'classnames/bind';
 import { SharedMessage } from './sharedMessage';
 import { FilterControls } from './filterControls';
+import { FiltersSorting } from '../../../../inside/dashboardItemPage/modals/common/widgetControls/controls/filtersControl/filtersSorting/filtersSorting';
```
DiscoElevator: I think it can be moved upper since it is used in 2 places now

PR #1238, opened 2019-03-11
EPMRPP-35499: Fix error when navigating to deleted item

DiscoElevator:


DiscoElevator:
Also fixes EPMRPP-35498

app/src/pages/inside/testItemPage/notFound.jsx
```diff
@@ -5,19 +5,25 @@ import styles from './notFound.scss';
 
 const cx = classNames.bind(styles);
 const messages = defineMessages({
-  notFound: {
-    id: 'LaunchesPage.notFound',
+  launchNotFound: {
+    id: 'LaunchesPage.launchNotFound',
     defaultMessage: 'Launch is not found',
   },
   notFoundDescription: {
```
ElenaRomanchuk: It looks like you forgot translate this message and customise it for both of use case.

DiscoElevator: It's the same behavior as in v4

PR #1245, opened 2019-03-11
EPMRPP-35140: Fix token storage

DiscoElevator:


app/src/components/containers/initialDataContainer/initialDataContainer.jsx
```diff
@@ -44,6 +44,7 @@ export class InitialDataContainer extends Component {
   };
 
   componentDidMount() {
+    this.props.resetTokenAction();
```
AmsterGet: After success login and page refreshing we will reset token in `localStorage`.
It isn't expected (see the screenshot below).
image

DiscoElevator: fixed

PR #1257, opened 2019-03-14
EPMRPP-37314: Add Administrative - Projects page with data loading

margalencia:
image

1. Page layout: Search field, Export button, Change view buttons, Pagination bar (see in `pages/admin/projectsPage/projects/projects.jsx`)
2. Integration with the backend


app/src/controllers/administrate/projects/actionCreators.js
```diff
@@ -0,0 +1,16 @@
+import { getStorageItem, setStorageItem } from 'common/utils';
+import { FETCH_PROJECTS, TABLE_VIEW, USER_VIEW, TOGGLE_PROJECTS_VIEW } from './constants';
+
+export const fetchProjectsAction = (params) => ({
+  type: FETCH_PROJECTS,
+  payload: params,
+});
+
+export const toggleViewAction = (type) => (dispatch) => {
+  const userView = getStorageItem(USER_VIEW) || TABLE_VIEW;
```
DiscoElevator: Isn't grid should be default view?

app/src/pages/admin/projectsPage/projects/projects.scss
```diff
@@ -0,0 +1,32 @@
...
+.search {
+  width: 400px;
+}
+
+.buttons {
+  display: flex;
+}
+
+.toolbar-button {
+  margin-left: 10px;
```
DiscoElevator: Accrording to the design, view toggling buttons shouldn't have space between them

app/src/controllers/administrate/projects/selectors.js
```diff
@@ -0,0 +1,14 @@
+import { createQueryParametersSelector } from './../../pages';
+import { DEFAULT_PAGINATION } from './constants';
+import { administrateDomainSelector } from '../selectors';
+
+const domainSelector = (state) => administrateDomainSelector(state).projects || {};
+
+export const projectsPaginationSelector = (state) => domainSelector(state).pagination;
+export const projectsSelector = (state) => domainSelector(state).projects;
+export const loadingSelector = (state) => domainSelector(state).loading || false;
+export const viewTypeSelector = (state) => domainSelector(state).viewType;
```
DiscoElevator: I think `viewMode`sounds better

app/src/controllers/administrate/projects/actionCreators.js
margalencia: on Dashboard Table View is the default, probably should be the same here?

app/src/pages/admin/projectsPage/projects/projects.scss
margalencia: But on Dashboard, there is a space, probably should be the same here?
image

app/src/controllers/administrate/projects/actionCreators.js
DiscoElevator: I think it's better to clarify this with BA

app/src/pages/admin/projectsPage/projects/projects.scss
DiscoElevator: I think it needs to be clarified with the UX team. Let's leave it as is for now, and if there shouldn't be space, fix it in another task

PR #1283, opened 2019-03-20
EPMRPP-38377: React. Project members. User invite. User search doesn'…

vitalii-chornyi:
.

### DIFF
--- a/app/src/common/urls.js
+++ b/app/src/common/urls.js
@@ -195,6 +195,10 @@ export const URLS = {
   statisticsServerSettings: () => `${urlBase}settings/analytics`,
   events: (projectId) => `${urlBase}${projectId}/activity`,
   allUsers: () => `${urlBase}user/all`,
+  searchUsers: (term) =>
+    `${urlBase}user/all${getQueryParams({
+      'filter.cnt.email': term,
+    })}`,
   exportUsers: () =>
     `${urlBase}user/export${getQueryParams({
       view: 'csv',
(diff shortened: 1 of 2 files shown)


app/src/common/urls.js
```diff
@@ -195,6 +195,10 @@ export const URLS = {
   statisticsServerSettings: () => `${urlBase}settings/analytics`,
   events: (projectId) => `${urlBase}${projectId}/activity`,
   allUsers: () => `${urlBase}user/all`,
+  searchUsers: (term) =>
+    `${urlBase}user/all${getQueryParams({
+      'filter.cnt.email': term,
```
DiscoElevator: We need to search by email and login. Maybe there is a special filter for this or a way to combine two filters.

app/src/components/inputs/inputUserSearch/inputUserSearch.jsx
```diff
@@ -36,15 +36,8 @@ const newOptionCreator = (option) => ({
   userLogin: option.label,
 });
 const promptTextCreator = (label) => label;
-const makeURL = (input, isAdmin, projectId) => {
-  let url;
-  if (!isAdmin) {
-    url = URLS.projectUserSearchUser(projectId, input);
-  } else {
-    url = URLS.projectAdminSearchUser(input);
```
DiscoElevator: Also, please remove `URLS.projectAdminSearchUser` if it's not used anymore

PR #1288, opened 2019-03-21
EPMRPP-36947 React. Admin part. All users. Button and Modal Invite user

vitalii-chornyi:


app/src/pages/inside/membersPage/modals/inviteUserModal/inviteUserModal.jsx
```diff
@@ -80,6 +85,10 @@ export class InviteUserModal extends Component {
     activeProject: '',
     isAdmin: false,
   };
+  constructor() {
```
DiscoElevator:
```
constructor(props) {
  super(props);
  // ...
}
```

app/src/pages/inside/membersPage/modals/inviteUserModal/inviteUserModal.jsx
```diff
@@ -125,6 +135,13 @@ export class InviteUserModal extends Component {
               </FieldErrorHint>
             </FieldProvider>
           </ModalField>
+          {data.isProjectSelector ? (
```
DiscoElevator: `{data.isProjectSelector && ( ... )}`

app/src/pages/inside/membersPage/modals/inviteUserModal/inviteUserModal.jsx
```diff
@@ -125,6 +148,18 @@ export class InviteUserModal extends Component {
               </FieldErrorHint>
             </FieldProvider>
           </ModalField>
+          {data.isProjectSelector && (
+            <ModalField label="Project" name="project" labelWidth={LABEL_WIDTH}>
+              <InputTagsSearch
+                minLength={1}
+                async
+                uri={this.projectSearchUrl}
+                value={this.state.projectFormattedValue}
```
DiscoElevator: You don't need to store value in state as it comes from the redux-form. You can specify parse/format functions to convert value and pass them to `FieldProvider` as props. (Also, you need to wrap `InputTagsSearch` with `FieldProvider` to connect it to the form, see lines 141-149 for reference)

PR #1307, opened 2019-03-27
EPMRPP-33467 React. Dashboard. Widget. cumulative trend chart

vitalii-chornyi:


app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
```diff
@@ -0,0 +1,242 @@
...
+      this.width = newWidth;
+    }
+  };
+
+  renderContents = (d) => {
+    const index = d[0].index;
+    const groupName = this.dataGroupNames[index];
+    const columns = this.config.data.columns;
+    const itemsData = [];
+    columns.forEach((column) => {
```
margalencia: maybe `map` is better here? `const itemsData = columns.map`

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
```diff
@@ -0,0 +1,242 @@
...
+      'preview-view': isPreview,
+    });
+    return (
+      <div className={classes}>
+        {this.state.isConfigReady && (
+          <C3Chart config={this.config} onChartCreated={this.onChartCreated}>
+            {!this.props.isPreview && (
+              <Legend
+                items={Object.getOwnPropertyNames(this.config.data.colors)}
+                onClick={this.onClick}
```
margalencia: probably descriptive method names will be more readable? for example `onClick={this.toggleChart}, onMouseOver={this.focusChart}`

app/src/components/widgets/charts/cumulativeTrendChart/generateConfig.js
```diff
@@ -0,0 +1,42 @@
...
+
+const getTypeSum = (items, type) =>
+  items.map((item) => Number(item.values[type])).reduce((acc, item) => acc + item);
+
+export const generateChartDataParams = (widget) => {
+  const chartDataColumns = [];
+  const dataGroupNames = [];
+  const contentFields = widget.contentParameters.contentFields;
+
+  contentFields.forEach((field) => {
```
margalencia: also `map` here

app/src/components/widgets/charts/cumulativeTrendChart/tooltip.jsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const CumulativeTrendTooltip = ({ groupName, itemsData }) => (
+  <React.Fragment>
+    <div className="tooltip-header">
```
margalencia: you use className without `cx` function, will the component name prefix be added to the class name in resulted html? what if another component have the same class name?

app/src/components/widgets/charts/cumulativeTrendChart/generateConfig.js
```diff
@@ -0,0 +1,42 @@
...
+    chartDataColumns.push([field]);
+  });
+
+  Object.keys(widget.content.result)
+    .sort()
+    .forEach((resultId) => {
```
margalencia: probably `map`here?

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.scss
```diff
@@ -0,0 +1,99 @@
...
+        color: $COLOR--gray-47;
+        overflow: hidden;
+
+        .item-wrapper {
+          clear: both;
+          position: relative;
+          height: 25px;
+          line-height: 25px;
+
+          .color-mark {
```
margalencia: do you need nesting in css?

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
vitalii-chornyi: If we change logic of event handler in future do we need to update method name to something more appropriate like onClick={this.toggleEntireWidget} ?

margalencia: I didn't get for sure what do you mean, but my point is that function name usually reflects what it does, not when it is called. If the logic changes the name will probably change. The names I've provided are just to illustrate what I mean, because I don't know for sure what `this.chart.toggle(id)`

vitalii-chornyi: I encourage descriptive names for methods as well and I understood your point but I disagree with your approach in this case. And I provided method name just for example either. I prefer to distinguish event handlers like onClick, onSubmit,.. from class methods with  business logic so you can change it anytime without updating a template. In another words onClick is always onClick and it doesn't depend on what do we do on it.

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
vitalii-chornyi: fixed

app/src/components/widgets/charts/cumulativeTrendChart/generateConfig.js
vitalii-chornyi: fixed

app/src/components/widgets/charts/cumulativeTrendChart/tooltip.jsx
vitalii-chornyi: fixed

app/src/components/widgets/charts/cumulativeTrendChart/generateConfig.js
vitalii-chornyi: fixed

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.scss
vitalii-chornyi: fixed

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
```diff
@@ -0,0 +1,241 @@
...
+
+@injectIntl
+export class CumulativeTrendChart extends Component {
+  static propTypes = {
+    intl: intlShape.isRequired,
+    widget: PropTypes.object.isRequired,
+    isPreview: PropTypes.bool,
+    height: PropTypes.number,
+    container: PropTypes.instanceOf(Element).isRequired,
+    observer: PropTypes.object,
```
AmsterGet: Please provide props to manage widget legend and use it inside of your component.
See f.e. `widgets/charts/launchStatisticsChart/launchStatisticsChart.jsx` lines 63, 64.

vitalii-chornyi: done

app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.stories.jsx
```diff
@@ -0,0 +1,54 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Report Portal.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import { storiesOf } from '@storybook/react';
+import { host } from 'storybook-host';
+import { withReadme } from 'storybook-readme';
+import 'c3/c3.css';
```
AmsterGet: It's already imported at `.storybook/config.js`

vitalii-chornyi: Removed import c3.css. storiesOf, host and withReadme  have been used in this file

PR #1309, opened 2019-03-27
EPMRPP-33317: Defect Types - add page layout

margalencia:
image


app/src/pages/common/settingsPage/defectTypesTab/defectTypesTab.jsx
```diff
@@ -0,0 +1,99 @@
...
+            {intl.formatMessage(messages.abbreviationCol)}
+          </div>
+          <div className={cx('header-cell', 'color-cell')}>
+            {intl.formatMessage(messages.colorCol)}
+          </div>
+          <div className={cx('header-cell', 'diagram-cell')}>
+            {intl.formatMessage(messages.diagramCol)}
+          </div>
+        </div>
+        {[PRODUCT_BUG, SYSTEM_ISSUE, AUTOMATION_BUG, NO_DEFECT, TO_INVESTIGATE].map((groupName) => (
```
AmsterGet: Each call of `render` method will create the new array with the same items.
I think it's better to move this array definition to a class field, f.e. `groupNames`

app/src/pages/common/settingsPage/defectTypesTab/defectTypePropType.js
```diff
@@ -0,0 +1,10 @@
+import PropTypes from 'prop-types';
+
+export const defectTypePropType = {
```
AmsterGet: I think it's better to write `defectTypePropType = PropTypes.shape(...)` here and remove the same construction in the components where it was used.

app/src/pages/common/settingsPage/defectTypesTab/defectTypesTab.jsx
margalencia: done

PR #1312, opened 2019-03-28
EPMRPP-38393: Reset page number on changing launch filter

DiscoElevator:
Sorting will be implementet in #1237

### DIFF
--- a/app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
+++ b/app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
@@ -16,6 +16,7 @@ import { fetchLaunchesWithParamsAction, fetchLaunchesAction } from 'controllers/
 import { debounce } from 'common/utils';
 import { hideFilterOnLaunchesAction } from 'controllers/project';
 import { isEmptyValue } from 'common/utils/isEmptyValue';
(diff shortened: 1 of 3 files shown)


app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -90,16 +93,20 @@ export class LaunchFiltersContainer extends Component {
...
-      this.updateFilter(this.props.activeFilterId, addFilteringFieldToConditions(conditions));
+      this.updateFilter(this.props.activeFilterId, conditionsWithFilteringField);
     } else {
       this.props.createFilter({
-        conditions: addFilteringFieldToConditions(conditions),
+        conditions: conditionsWithFilteringField,
       });
     }
+
+    this.props.onChange(conditions);
```
AmsterGet: I see that this handler doesn't have any arguments.
[Code line](https://github.com/reportportal/service-ui/pull/1312/files#diff-cd9d7a96a10c68f93eec371ce8b5b725R354)

DiscoElevator: Right. Argument is not used for now. I just think it's a common pattern to pass new value to `onChange` no matter if someone plans to use it or not.

PR #1315, opened 2019-03-29
EPMRPP-36933 Log page local storage sync fixes

AmsterGet:
Task - https://jiraeu.epam.com/browse/EPMRPP-36933


app/src/controllers/log/index.js
```diff
@@ -17,3 +23,14 @@ export {
   previousLogLinkSelector,
   retryLinkSelector,
 } from './selectors';
+export {
```
DiscoElevator: Are all this functions should be exported? They use each other and I think that the list should look like this:
```
  getWithAttachments,
  setWithAttachments,
  setLogLevel,
  getLogLevel,
  getLogViewMode,
  setLogViewMode,
```

AmsterGet: Fixed. But also we need `getLogLevelById` util to take corresponding log level object with `id` and `label`

PR #1316, opened 2019-03-29
EPMRPP-33317: Subtypes chart + Add subtype button

margalencia:
image


app/src/pages/common/settingsPage/defectTypesTab/defectTypesGroup.jsx
```diff
@@ -16,27 +22,88 @@ ColorMarker.propTypes = {
...
-            <span className={cx('defect-type-name')}>{longName}</span>
-          </span>
-        </div>
-        <div className={cx('abbr-cell')}>{shortName}</div>
-        <div className={cx('color-cell')}>
-          <ColorMarker color={color} />
+    const { group, intl } = this.props;
+
+    return (
+      <div>
```
DiscoElevator: Is this div necessary? You can use `Fragment` instead

app/src/pages/common/settingsPage/defectTypesTab/defectTypesTab.scss
```diff
@@ -65,3 +74,21 @@
 .defect-type-name {
   margin-left: 5px;
 }
+
+.defect-type-chart {
+  position: absolute !important;
```
DiscoElevator: Why do you use `!important` here?

margalencia: I have to use it to overwrite the chart's style, it has `position: relative`.
I need to remove the chart from the normal flow so it doesn't impact the first row height.

app/src/pages/common/settingsPage/defectTypesTab/defectTypesGroup.jsx
margalencia: yep, for now, let's live it as it is, I will fix it in the next PR, I've restructured the files, created subcomponents

DiscoElevator:
изображение

margalencia:
will be fixed in the next pr 
image

PR #1310, opened 2019-03-28
EPMRPP-37395. Admin. Projects. Actions in menu.

ElenaRomanchuk:

PR #1317, opened 2019-04-01
EPMRPP-37317 Admin. Projects. Assign user on redirect to project.

ElenaRomanchuk:
EPMRPP-37317. Admin. Projects. Assign user on redirect to project.
! Based on EPMRPP-37395. Please, check and merge PR https://github.com/reportportal/service-ui/pull/1310 at first.


app/src/pages/admin/projectsPage/projectMenu/projectMenu.jsx
```diff
@@ -0,0 +1,185 @@
...
+    id: 'ProjectsPage.modal.deleteModalConfirmationText',
+    defaultMessage: 'Are you sure you want to delete the project <b>{name}</b>?',
+  },
+  deleteModalSubmitButtonText: {
+    id: 'ProjectsPage.modal.deleteModalSubmitButtonText',
+    defaultMessage: 'Delete',
+  },
+  deleteError: {
+    id: 'ProjectsPage.deleteError',
+    defaultMessage: 'An error occurred during deleting the project',
```
margalencia: I have the same messages for the bulk `delete` operation in Grid view in `projectsPage/projectsToolbar/projectsToolbar.jsx`, maybe we can use a shared mesage config for them?

ElenaRomanchuk: Is it a good idea to extract all messages into separate file for common use in Projects?

DiscoElevator: I think it's a good idea

ElenaRomanchuk: Done!

app/src/pages/common/modals/confirmModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+    return (
+      <ModalLayout
+        title={title}
+        okButton={{
+          text: confirmText,
+          onClick: (closeModal) => {
+            confirmModal();
+            closeModal();
+            onConfirm();
+            hideModal();
```
DiscoElevator: Is it necessary to use `hideModal` after `closeModal`? As far as I see it is called automatically after closing a modal.

app/src/pages/common/modals/confirmModal.jsx
```diff
@@ -0,0 +1,46 @@
...
+import { connect } from 'react-redux';
+import { ModalLayout, withModal } from 'components/main/modal';
+import { hideModalAction, confirmModalAction } from 'controllers/modal';
+
+@withModal('confirmModal')
+@connect(null, {
+  confirmModal: confirmModalAction,
+  hideModal: hideModalAction,
+})
+export class ConfirmModal extends Component {
```
DiscoElevator: I think `ConfirmationModal` sounds better (so no confusion between `confirmModal` as an action and `ConfirmModal` as a component)

app/src/pages/common/modals/confirmModal.jsx
ElenaRomanchuk: `hideModal` is called in `onExited` event of `CSSTransition` component in `modalLayout`. Unfortunately, it doesn't work properly when you perform an asynchronous redirect to another page.

DiscoElevator: Maybe dispatching `hideModalAction` in saga would look less confusing.

ElenaRomanchuk: I suggest dispatch it in confirmModalAction. Would it be OK?

DiscoElevator: You can't dispatch two actions in one (there's a task to remove all thunk usage in favor of saga - no need to keep both libraries).
Somewhere [here](https://github.com/reportportal/service-ui/pull/1317/files#diff-4b647578726008404857fb737592a00cR11), I think:
```
if (confirmed) {
  yield put(hideModalAction());
}
```

PR #1329, opened 2019-04-03
EPMRPP-38511: part of EPMRPP-33317. Edit subtype form + color picker

margalencia:
Still need to implement:

1. Close color picker when click outside
2. Position input fields according to the design
3. Show locator tooltip on defects name

image

image

### DIFF
--- a/app/package.json
+++ b/app/package.json
@@ -49,6 +49,7 @@
     "react": "16.6.0",
     "react-autosuggest": "9.3.4",
     "react-c3js": "0.1.20",
+    "react-color": "2.17.0",
     "react-copy-to-clipboard": "5.0.1",
     "react-custom-scrollbars": "4.2.1",
     "react-datepicker": "1.4.1",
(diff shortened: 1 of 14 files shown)


app/src/components/main/colorPicker/colorPicker.jsx
```diff
@@ -0,0 +1,129 @@
...
+                  ref={ref}
+                  style={style}
+                  data-placement={placement}
+                  onClick={(e) => e.stopPropagation()}
+                >
+                  <div className={cx('title')}>{intl.formatMessage(messages.pickSwatch)}:</div>
+                  <CirclePicker
+                    color={hex}
+                    onChange={onChange}
+                    colors={this.COLORS}
```
ElenaRomanchuk: May be it would be better to setup color picker settings as constants, which would be better to put in a separate file.

margalencia: done

app/package.json
```diff
@@ -49,6 +49,7 @@
     "react": "16.6.0",
     "react-autosuggest": "9.3.4",
     "react-c3js": "0.1.20",
+    "react-color": "^2.17.0",
```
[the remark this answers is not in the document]
margalencia: done

PR #1336, opened 2019-04-05
EPMRPP-38403. Admin. Projects. Improvements.

ElenaRomanchuk:


app/src/pages/admin/projectsPage/projectStatisticButton/projectStatisticButton.jsx
```diff
@@ -0,0 +1,26 @@
+import PropTypes from 'prop-types';
+import { NavLink } from 'redux-first-router-link';
+import React from 'react';
+import { PROJECT_DETAILS_PAGE } from 'controllers/pages';
+import { Icon } from 'components/main/icon/icon';
+
+export const ProjectStatisticButton = ({ projectName, onClick }) => (
+  <NavLink
```
DiscoElevator: I think it can be `Link`.
`NavLink` is used in cases when you need to change link style accorging to a current route (for example, sidebar buttons).

PR #1337, opened 2019-04-05
EPMRPP-36894 REACT. Boundary lines cross over the tooltip

vitalii-chornyi:


app/src/common/css/variables/display.scss
```diff
@@ -0,0 +1,2 @@
+$Z-INDEX-DEFAULT: 1;
```
DiscoElevator: I think it's better to name this file `z-index-variables.scss` as more descriptive

vitalii-chornyi: done

PR #1344, opened 2019-04-08
EPMRPP-38547. Ascending sort when switching the sort field.

ElenaRomanchuk:


app/src/controllers/sorting/withSortingURL.jsx
```diff
@@ -39,6 +39,8 @@ export const withSortingURL = ({
       const fields = oldFields.length > 0 ? oldFields : defaultFields;
       if (fields.includes(field)) {
         direction = direction === SORTING_ASC ? SORTING_DESC : SORTING_ASC;
+      } else {
+        direction = oldDirection ? SORTING_ASC : defaultDirection;
```
DiscoElevator: I think there should be `direction = defaultDirection;`

ElenaRomanchuk: No. I discussed this with BA and she approved current behavior.

PR #1356, opened 2019-04-11
EPMRPP-38460. Launches page. Fix breadcrumbs overflow

ultd39:
**Before:**
localhost_3000_ (1)
**Afrer:**
localhost_3000_

**Before:**
localhost_3000_ (3)
**Afrer:**
localhost_3000_ (2)

### DIFF
--- a/app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
+++ b/app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
@@ -5,6 +5,8 @@
   padding-right: 5px;
   font-size: 12px;
   color: $COLOR--charcoal-grey;
+  word-break: break-word;
+  line-height: 18px;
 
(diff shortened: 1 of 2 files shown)


app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
```diff
@@ -5,6 +5,8 @@
   padding-right: 5px;
   font-size: 12px;
   color: $COLOR--charcoal-grey;
+  word-break: break-word;
```
DiscoElevator: This value doesn't work in firefox. Please add fallback `break-all`:
```
word-break break-all;

@supports(word-break: break-word) {
  word-break: break-word;
}
```

ultd39: You right! Thanks!

PR #1362, opened 2019-04-11
EPMRPP-37398 [REACT] All launches are loaded on one page despite limi…

vitalii-chornyi:
…tation per page


app/src/controllers/pagination/constants.js
```diff
@@ -6,5 +6,5 @@ export const SIZE_KEY = 'page.size';
 
 export const DEFAULT_PAGINATION = {
   [PAGE_KEY]: 1,
-  [SIZE_KEY]: 50,
+  [SIZE_KEY]: 20,
```
DiscoElevator: According to [this ticket](https://jiraeu.epam.com/browse/EPMRPP-35245), default page size should be 50

PR #1378, opened 2019-04-16
EPMRPP-37759, EPMRPP-38541. Redirect. Saving browser history on redirect

ElenaRomanchuk:
Exclude `redirect` function in ordinal page navigation, saving history of page visits.

### DIFF
--- a/app/src/common/utils/routingUtils.js
+++ b/app/src/common/utils/routingUtils.js
@@ -32,3 +32,5 @@ export const mergeQuery = (oldQuery, paramsToMerge) => {
     {},
   );
 };
+
+export const redirectAction = (linkAction) => linkAction;
--- a/app/src/common/utils/routingUtils.js
+++ b/app/src/common/utils/routingUtils.js
@@ -32,5 +32,3 @@ export const mergeQuery = (oldQuery, paramsToMerge) => {
     {},
   );
 };
-
-export const redirectAction = (linkAction) => linkAction;
--- a/app/src/components/widgets/charts/launchExecutionAndIssueStatistics/issueStatisticsChart.jsx
+++ b/app/src/components/widgets/charts/launchExecutionAndIssueStatistics/issueStatisticsChart.jsx
@@ -24,12 +24,12 @@ import { injectIntl, intlShape } from 'react-intl';
 import * as d3 from 'd3-selection';
 import classNames from 'classnames/bind';
 import { connect } from 'react-redux';
-import { redirect } from 'redux-first-router';
 import ReactDOMServer from 'react-dom/server';
 import { defectLinkSelector } from 'controllers/testItem';
 import { defectTypesSelector } from 'controllers/project';
 import { activeProjectSelector } from 'controllers/user';
 import { TEST_ITEM_PAGE } from 'controllers/pages';
+import { redirectAction } from 'common/utils/routingUtils';
 import { TooltipWrapper } from '../common/tooltip';
 import { C3Chart } from '../common/c3chart';
 import chartStyles from './launchExecutionAndIssueStatistics.scss';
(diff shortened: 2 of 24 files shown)


app/src/common/utils/routingUtils.js
```diff
@@ -32,3 +32,5 @@ export const mergeQuery = (oldQuery, paramsToMerge) => {
     {},
   );
 };
+
+export const redirectAction = (linkAction) => linkAction;
```
DiscoElevator: I don't think it's necessary since this function does nothing.

app/src/components/widgets/charts/launchExecutionAndIssueStatistics/issueStatisticsChart.jsx
```diff
@@ -50,7 +50,7 @@ const getResult = (widget) => widget.content.result[0] || widget.content.result;
     getDefectLink: (params) => defectLinkSelector(state, params),
   }),
   {
-    redirect,
+    redirect: redirectAction,
```
DiscoElevator: I think `redirect` will be confusing in this way. My suggestion is:
`navigate: (link) => link`

app/src/controllers/administrate/projects/actionCreators.js
```diff
@@ -53,3 +56,12 @@ export const confirmAssignToProject = (project) => ({
   type: CONFIRM_ASSIGN_TO_PROJECT,
   payload: project,
 });
+
+export const redirectToProjectSectionAction = (projectName, section) => ({
```
DiscoElevator: navigate...

app/src/controllers/pages/actionCreators.js
```diff
@@ -19,5 +18,5 @@ export const updatePagePropertiesAction = (properties) => (dispatch, getState) =
     meta: { query: newQuery },
   };
 
-  dispatch(redirect(updatedAction));
+  dispatch(updatedAction);
```
DiscoElevator: Thet's tricky. For example, if I add/remove different filters, by pressing 'Back' button I will go through the history of my filter edits. I don't think it's convenient since there may be many edits, and in this case I won't get back to the previous page.
I think this behavior should be discussed with BA.

app/src/common/utils/routingUtils.js
ElenaRomanchuk: It just wraps linkAction into action creator function. I can replace it by the same construction in @connect for each component like this:
```@connect(
null, 
{
 redirect: (linkAction) =>(linkAction)
}```
or
```@connect(
null,
(dispatch) => {
 redirect: (linkAction) => dispatch(linkAction)
}```
or I can write navigate action creator for each link.

DiscoElevator: `@connect( null, { navigate: (linkAction) =>(linkAction) })`

app/src/components/widgets/charts/launchExecutionAndIssueStatistics/issueStatisticsChart.jsx
ultd39: I agree. Now it isn't actually a redirect.

ElenaRomanchuk: Agree. I've fixed it.

app/src/common/utils/routingUtils.js
ElenaRomanchuk: Fixed.

app/src/controllers/pages/actionCreators.js
ElenaRomanchuk: I've discussed this behavior with BA. She approved it. Changing query should affect browser history.

PR #1379, opened 2019-04-16
EPMRPP-37192 REACT. Arrow and Refresh buttons are displayed incorrect…

vitalii-chornyi:
…ly on log view if step name is long

### DIFF
--- a/app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
+++ b/app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
@@ -5,6 +5,7 @@
   padding-right: 5px;
   font-size: 12px;
   color: $COLOR--charcoal-grey;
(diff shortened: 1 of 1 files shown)


app/src/components/main/breadcrumbs/breadcrumb/breadcrumb.scss
```diff
@@ -5,6 +5,7 @@
   padding-right: 5px;
   font-size: 12px;
   color: $COLOR--charcoal-grey;
+  min-width: 15px;
```
AmsterGet: Do these changes do what the task requires?

vitalii-chornyi: they don't but the rule 'word-break' does. It was rebased from another commit in v5 and merged here so only one line is remaining

PR #1381, opened 2019-04-16
EPMRPP-38613: part of EPMRPP-33317, add defect type's tooltip with locator. Refactor withHoverable tooltip to use react-popper

margalencia:
image

image

image

### DIFF
--- a/app/src/components/main/tooltips/hoverableTooltip/hoverableTooltip.scss
+++ b/app/src/components/main/tooltips/hoverableTooltip/hoverableTooltip.scss
@@ -1,75 +1,149 @@
-.hover-rect {
-  position: relative;
-  display: inline-block;
-  width: 100%;
-  height: 100%;
-  vertical-align: top;
+$ARROW_HEIGHT: 6px;
+$ARROW_WIDTH: 16px;
 
-  &:hover {
-    .hoverable-tooltip {
-      visibility: visible;
-      opacity: 1;
-    }
-  }
+.tooltip-trigger {
+  display: inline-block;
 }
 
-.hoverable-tooltip {
-  position: absolute;
-  visibility: hidden;
-  padding-top: 5px;
-  opacity: 0;
-  transition: visibility 0ms, opacity 300ms ease;
-  z-index: $Z-INDEX-POPUP;
-
-  &::before {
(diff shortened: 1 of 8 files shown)


app/src/components/main/tooltips/hoverableTooltip/withHoverableTooltip.jsx
```diff
@@ -1,65 +1,83 @@
...
+            ReactDOM.createPortal(
+              <Popper placement={data.placement} modifiers={data.modifiers}>
+                {({ placement, ref, style, arrowProps }) => (
+                  <div
+                    className={cx('tooltip', {
+                      'no-mobile': data.noMobile,
+                      'desktop-only': data.desktopOnly,
+                    })}
+                    ref={ref}
+                    style={{ ...style, ...{ width: data.width } }}
```
DiscoElevator: `                    style={{ ...style, width: data.width }}`

app/src/components/main/tooltips/hoverableTooltip/hoverableTooltip.scss
```diff
@@ -1,75 +1,149 @@
...
 }
 
-.hoverable-tooltip {
-  position: absolute;
-  visibility: hidden;
-  padding-top: 5px;
-  opacity: 0;
-  transition: visibility 0ms, opacity 300ms ease;
-  z-index: $Z-INDEX-TOOLTIP;
+/* stylelint-disable */
```
DiscoElevator: I think it's better to turn off rules for specific strings instead of turning off linter for almost the whole file.

margalencia: It doesn't work with separate lines, because as I understood correctly, the linter runs on resulting css-file, and there are a lot of lines with attribute selector in it, so I sill get errors

DiscoElevator: Try `/* stylelint-disable-line selector-max-attribute */`

margalencia: done with `/* stylelint-disable selector-max-attribute */`

PR #1386, opened 2019-04-17
EPMRPP-38296: React. Launches tab. There is no asterisk sign for not …

mykhailo-kalashnikov-epam:
…saved filters

### DIFF
--- a/app/src/controllers/filter/selectors.js
+++ b/app/src/controllers/filter/selectors.js
@@ -25,9 +25,12 @@ export const unsavedFilterIdsSelector = createSelector(
   userFiltersSelector,
   (filters, savedFilters) =>
     filters
-      .map((filter) => {
-        const savedFilter = savedFilters.find((item) => item.id === filter.id);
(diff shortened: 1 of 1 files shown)


app/src/controllers/filter/selectors.js
```diff
@@ -20,14 +18,6 @@ export const activeFilterSelector = createSelector(
...
-  launchFiltersSelector,
-  userFiltersSelector,
-  (filters, savedFilters) =>
-    filters
-      .map((filter) => {
-        const savedFilter = savedFilters.find((item) => item.id === filter.id);
-        return savedFilter && !isEqual(filter, savedFilter) ? savedFilter.id : null;
-      })
-      .filter(Boolean),
+export const unsavedFilterIdsSelector = createSelector(launchFiltersSelector, (filters) =>
```
AmsterGet: What do we get if a previously saved filter with a positive `id` is in edit mode with unsaved changes?

mykhailo-kalashnikov-epam: Good point.

PR #1394, opened 2019-04-18
EPMRPP-38621: Fix z-index for popups

DiscoElevator:


app/src/common/css/variables/z-index-variables.scss
```diff
@@ -1,2 +1,2 @@
 $Z-INDEX-DEFAULT: 1;
-$Z-INDEX-TOOLTIP: 10;
\ No newline at end of file
+$Z-INDEX-POPUP: 10;
```
vitalii-chornyi: Why did you remove z-index for tooltip ?
Can't we have they both on page, tooltip and popup ? Which of them should be on top?

DiscoElevator: It was just renamed. Also, tooltips will be changed in another PR: #1381

vitalii-chornyi: That's not about naming. I meant we need to distinguish about 4 different groups of views: default, dropdowns/tooltips/menus, popups and highest - modal window. And use for them z-index values like 1,10,100,1000 to don't interfere each other.

DiscoElevator: Totally agree. I'll look into that if I get some time.

PR #1402, opened 2019-04-22
EPMRPP-37146. Search on dashboards page

ultd39:
300 items by default and Not found message.

### DIFF
--- a/app/src/pages/inside/dashboardPage/dashboardPage.jsx
+++ b/app/src/pages/inside/dashboardPage/dashboardPage.jsx
@@ -218,7 +218,9 @@ export class DashboardPage extends Component {
             filter={filter}
             onFilterChange={onFilterChange}
           />
-          {dashboardItems.length > 0 ? (
+          {filter && dashboardItems.length === 0 ? (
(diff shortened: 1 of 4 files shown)


app/src/pages/inside/dashboardPage/dashboardPage.jsx
```diff
@@ -213,14 +218,18 @@ export class DashboardPage extends Component {
...
           />
-          <DashboardList
-            dashboardItems={dashboardItems}
-            gridType={gridType}
-            userInfo={userInfo}
-            onDeleteItem={this.onDeleteDashboardItem}
-            onEditItem={this.onEditDashboardItem}
-            onAddItem={this.onAddDashboardItem}
-          />
+          {dashboardItems.length > 0 ? (
```
DiscoElevator: Please add filter existence check, because in case project doesn't have any dashboards, there should be another view.

ultd39: Yout right! Thank you! I've added `filter` condition.

PR #1405, opened 2019-04-22
EPMRPP-38610. Redirect on last page in case of absence any items on current page.

ElenaRomanchuk:


app/src/controllers/pagination/withPagination.jsx
```diff
@@ -50,6 +50,12 @@ export const withPagination = ({
       updatePagination: () => {},
     };
 
+    componentDidUpdate() {
+      const totalPages = this.props.totalPages || 1;
+      if (this.props.page > totalPages) {
+        this.changePaginationOptions({ page: totalPages });
```
AmsterGet: Should we redirect to the first page (as described in the task - https://jiraeu.epam.com/browse/EPMRPP-38610)?

ElenaRomanchuk: As we discussed with BA, in case of absencens any items on a page we should redirect user on the last existed page. I' ll leave comment in the task.

AmsterGet: Ok

PR #1418, opened 2019-04-24
EPMRPP-36975. LaunchFiltersToolbar. Fix filters overlapping

ultd39:
**Before:**
Filters_before
**After:**
Filters_after


ultd39:
I have no idea. @DiscoElevator?

DiscoElevator:
I think it's better to clarify this with our UX team

ultd39:
Ok, I used the same approach as in the old version:
image

PR #1421, opened 2019-04-24
EPMRPP-35244. FiltersPage. Show update/delete confirmation

ultd39:
Just notification for update/delete actions.


app/src/pages/inside/filtersPage/filtersPage.jsx
```diff
@@ -150,7 +162,14 @@ export class FiltersPage extends Component {
     fetch(URLS.filter(this.props.activeProject, filter.id), {
       method: 'put',
       data: filter,
-    }).then(this.props.fetchFiltersAction);
+    })
+      .then(this.props.fetchFiltersAction())
```
DiscoElevator: `.then(this.props.fetchFiltersAction)`

ultd39: Oh, what a stupid mistake! Thank you!

PR #1432, opened 2019-04-26
EPMRPP-35421: add To Investigate chart, display To Investigate sub types in defect type selector, add To Investigate tooltip

margalencia:
image

image

image

image


app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.jsx
```diff
@@ -39,51 +42,59 @@ export class DefectTypeTooltip extends Component {
...
+              <span className={cx('value')}>{data.total}</span>
+            </DefectLink>
+            {/**
+             * Display defect types in a proper order,
+             * the first is the system type (Product Bug, To investigate, etc.)
+             * that is way they are sorted by ID
+             * Don't display defect sub types with zero defect's amount,
+             * except system types (i.e. with index=0)
+             */
+            Object.keys(data)
```
DiscoElevator: I think it's better to move this code as an util function.

app/src/pages/inside/common/launchSuiteGrid/defectStatistics/donutChart/donutChart.scss
```diff
@@ -15,6 +15,11 @@
   line-height: 18px;
   word-break: keep-all;
 }
+.total-to-investigate {
+  background-color: $COLOR--to-investigate-counter;
+  color: $COLOR--white;
+  border-color: $COLOR--white !important;
```
DiscoElevator: Why `!important`?

margalencia: because it is set as inline style depending on defect type color, red - for Product bug, etc.
see above in donutChart.jsx in line 130

there is a different style for To investigate counter 
image

app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.jsx
margalencia: it could be used somewhere else?
you mean move all this sequence of map, sort, filter.. ?

PR #1433, opened 2019-04-26
Epmrpp 38612

ElenaRomanchuk:


app/src/pages/inside/common/modals/deleteItemsModal/deleteItemsModal.scss
```diff
@@ -2,5 +2,6 @@
   line-height: 22px;
   b {
     font-family: $FONT-SEMIBOLD;
+    word-break: break-all;
```
DiscoElevator: I think we should add better breaking here:

```
word-break: break-all;

@supports(word-break: break-word) {
  word-break: break-word;
}
```

ElenaRomanchuk: Thanks a lot!

PR #1438, opened 2019-04-29
EPMRPP-37189 - REACT. Incorrect navigation between test items by arro…

mykhailo-kalashnikov-epam:
…w buttons on log view, if there are several pages of steps

### DIFF
--- a/app/src/controllers/log/selectors.js
+++ b/app/src/controllers/log/selectors.js
@@ -12,6 +12,7 @@ import {
 import { DEFAULT_PAGINATION } from 'controllers/pagination';
 import { itemsSelector } from 'controllers/testItem';
 import { debugModeSelector } from 'controllers/launch';
+import { stepPaginationSelector } from 'controllers/step';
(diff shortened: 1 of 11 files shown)


app/src/controllers/log/selectors.js
```diff
@@ -194,3 +195,25 @@ export const retryLinkSelector = createSelector(
     },
   }),
 );
+
+export const disablePrevItemLinkSelector = createSelector(
+  stepPaginationSelector,
```
DiscoElevator: Parent level can be any of levels. I suggest adding special selector which calls appropriate pagination selector (use `itemsSelector` as an example)

app/src/controllers/log/selectors.js
```diff
@@ -196,8 +199,21 @@ export const retryLinkSelector = createSelector(
   }),
 );
 
+export const paginationSelector = (state) => {
```
DiscoElevator: Please move it to testItem selectors

PR #1440, opened 2019-04-29
EPMRPP-38745. Fix project members page for just created project.

ElenaRomanchuk:


ElenaRomanchuk:
Should be fixed other case in the same place.

PR #1454, opened 2019-05-03
EPMRPP-36908: fix - React: Markdown options are not applied

Andrew-Omelchenko:
Fixes `Refused to load the image '<URL>' because it violates the following Content Security Policy directive: "img-src 'self' data: <URL> stats.g.doubleclick.net *.epam.com *.uservoice.com"
` console message and its consequences.

### DIFF
--- a/main.go
+++ b/main.go
@@ -62,7 +62,7 @@ func main() {
 			},
             "font-src": {"'self'", "fonts.googleapis.com", "fonts.gstatic.com"},
             "media-src": {"'self'", "*.saucelabs.com"},
-			"img-src":    {"'self'", "data:", "www.google-analytics.com", "stats.g.doubleclick.net", "*.epam.com", "*.uservoice.com", "*.saucelabs.com"},
+			"img-src": {"*", "'self'", "data:"},
 			"object-src": {"'self'"},
 		}
 


app/src/index.tpl.html
```diff
@@ -3,6 +3,9 @@
     <head>
       <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
       <meta charset="utf-8">
+      <meta http-equiv="Content-Security-Policy"
+        content="img-src * 'self' data: https:;"
```
AmsterGet: I think it will be better manage it in our `main.go` file in section 'Content security policies'.

Andrew-Omelchenko: Ok. I understand your comment. But... I've just made an attempt to clarify the root cause of the problem with loading images. To show ones with addresses like in `https://i.imgur.com/sZlktY7.png` (url taken from EPMRPP-36908), we ought to relax current security policy, prescribed in main.go file. Will it be correct? Or should I request permission from somebody, who is responsible to approve changes in security policy.

AmsterGet: Good comment. Let's discuss it on scrum with the whole team after the holidays.

AmsterGet: I clarify this question with [Andrei Varabyeu](https://github.com/avarabyeu). You can add `*` source to the `img-src` directive in `main.go` file.

Andrew-Omelchenko: Fixed.

PR #1458, opened 2019-05-07
EPMRPP-38828 - React. Launch. Incorrect filter behaviour on open by url

mykhailo-kalashnikov-epam:


app/src/common/utils/waitForSelector.js
```diff
@@ -0,0 +1,10 @@
+import { select, take } from 'redux-saga/effects';
+
+export function* waitForSelector(selector) {
```
margalencia: Could you please explain in more details what is the purpose of this generator?

mykhailo-kalashnikov-epam: Since the project settings request does not call from launch page Saga, I cannot know when response got, the only indicator is the active filter selector, this generator waits for any valid selector value and is interrupted.

PR #1469, opened 2019-05-08
EPMRPP-34392: fix toolbar on top when at least one launch/launch suite is selected

margalencia:


app/src/pages/inside/launchesPage/LaunchToolbar/launchToolbar.scss
```diff
@@ -2,3 +2,9 @@
   background-color: $COLOR--white-two;
   border-bottom: 1px solid $COLOR--gray-91;
 }
+
+.sticky-toolbar {
+  position: sticky;
+  top: 0;
+  z-index: 2;
```
vitalii-chornyi: Why 2 not 3?

ElenaRomanchuk: The z-index value constants are introduced in z-index-variables.sccs. I suggest to use them instead of hard coded number.

margalencia: fixed, I've added new variable to `z-index-variables`, 2 is to be displayed above the all table elements

PR #1473, opened 2019-05-10
EPMRPP-35191 - React: Lazy-load for widget on dashboard page

mykhailo-kalashnikov-epam:


app/src/components/main/scrollWrapper/scrollWrapper.jsx
```diff
@@ -141,6 +142,7 @@ export class ScrollWrapper extends Component {
     if (scrollLeft !== lastViewScrollLeft) {
       this.scrollbars.thumbHorizontal.style.opacity = 1;
     }
+    forceCheck();
```
DiscoElevator: Maybe we can somehow utilize `onLazyLoad` function? I mean, move this check there or inside the condition on line 139. What do you think?

PR #1475, opened 2019-05-10
EPMRPP-37440: incorrect tooltip for defect details when no defect subtypes are added

Andrew-Omelchenko:


app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.jsx
```diff
@@ -76,25 +91,12 @@ export class DefectTypeTooltip extends Component {
...
-                  className={cx('item')}
-                >
-                  <div className={cx('name')}>
-                    <div className={cx('circle')} style={{ backgroundColor: color }} />
-                    {longName}
-                  </div>
-                  <span className={cx('value')}>{data[locator]}</span>
-                </DefectLink>
-              ))}
+            Object.keys(data).length > 2 && (
```
margalencia: According to the defect's description 
> Defect type's statistic shouldn't contain total info if there are no defect subtypes on the project

image

You should leave `Product Bug` not `Total Product Bugs`

Andrew-Omelchenko: Fixed.

app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.jsx
```diff
@@ -53,48 +53,49 @@ export class DefectTypeTooltip extends Component {
 
     const defectConfig = projectConfig.subTypes && projectConfig.subTypes[type.toUpperCase()];
 
+    const filteredBodyData = Object.keys(data)
```
AmsterGet: I think it would be better to move this sequence to the `getFilteredBodyData` function and call it here.

app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.scss
```diff
@@ -3,11 +3,11 @@
 .total-item {
   display: flex;
   justify-content: space-between;
-  box-sizing: border-box;
+  text-decoration: none;
```
AmsterGet: Which purpose of this swap?
Please see our [docs](https://github.com/reportportal/service-ui/blob/v5/app/docs/8-CSS.md)

Andrew-Omelchenko: Fixed.

app/src/pages/inside/common/defectTypeTooltip/defectTypeTooltip.jsx
Andrew-Omelchenko: Fixed.

PR #1490, opened 2019-05-14
EPMRPP-38703. React. Extended disabling Save validation rule

twinfreaks:


DiscoElevator:
Looks like not actual - the case is not reproduced on DEV and there's already check for empty filter (`this.isNoFilterValues()`).
Will reopen this PR and recheck in case of the ticket reopening.

PR #1494, opened 2019-05-14
EPMRPP-38631 - React. Performance. WebPack re-configuration

mykhailo-kalashnikov-epam:


app/package.json
```diff
@@ -10,27 +10,25 @@
...
     "test": "jest --notify",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
     "build:lang:en": "webpack --display none --env.dev && babel localization/translator.js | node",
     "manage:translations": "webpack --display none --env.dev && babel-node ./localization/translatorRunner.js",
     "manage:translations:test": "webpack --display none --env.dev && babel-node ./localization/translatorRunnerCheck.js",
     "precommit": "lint-staged",
-    "postcommit": "git update-index --again"
+    "postcommit": "git update-index --again",
+    "storybook": "NODE_ENV=development start-storybook -p 9001 -c .storybook",
```
AmsterGet: Is it variable necessary to run storybook? It seems that `start-storybook` script doesn't use `NODE_ENV` environment variable

app/package.json
```diff
@@ -10,27 +10,25 @@
     "format:js": "npm run eslint -- --fix",
     "format:scss": "npm run stylelint -- --fix",
     "format": "npm run format:js && npm run format:scss && prettier --write */**/*.md",
-    "dev": "webpack-dev-server --port 3000",
-    "build": "webpack -p --env.production",
-    "storybook": "start-storybook -p 9001 -c .storybook",
-    "storybook:build": "build-storybook -c .storybook -o .storybook_app",
+    "dev": "NODE_ENV=development webpack-dev-server --port 3000",
```
AmsterGet: I recommend to provide webpack `--env` option: `--env.NODE_ENV=development` and use it in the config file

app/package.json
```diff
@@ -10,27 +10,25 @@
...
+    "build": "webpack -p",
     "test": "jest --notify",
     "test:watch": "jest --watch",
     "test:coverage": "jest --coverage",
     "build:lang:en": "webpack --display none --env.dev && babel localization/translator.js | node",
```
AmsterGet: Also please update this scripts with proper `--env` options

app/webpack.config.js
```diff
@@ -0,0 +1,12 @@
+const merge = require('webpack-merge');
+const baseConfig = require('./webpack/base.config');
+const prodConfig = require('./webpack/prod.config');
+const devConfig = require('./webpack/dev.config');
+
+module.exports = () => {
+  const mode = process.env.NODE_ENV;
```
AmsterGet: With `--env.NODE_ENV=development` webpack option you will get access to `NODE_ENV` as a property of function parameters object

app/package.json
mykhailo-kalashnikov-epam: production mode used by default.

PR #1495, opened 2019-05-14
EPMRPP-34334. Grid. Sticky header

ultd39:
Not so DRY but we have a different component for a checkbox.


app/src/components/main/grid/gridHeader/checkboxHeaderCell/checkboxHeaderCell.scss
```diff
@@ -1,11 +1,14 @@
 .checkbox-header-cell {
   display: table-cell;
-  position: relative;
+  position: sticky;
+  top: 0;
+  z-index: 2;
```
margalencia: I've added a variable to `variables/z-index-variables.scss` for sticky toolbar z-index, you can use it too

ultd39: Thank you. Fixed.

PR #1496, opened 2019-05-14
EPMRPP-38814. Events monitoring. Update activity types and object types in filter.

ElenaRomanchuk:


app/src/pages/admin/projectEventsPage/eventsGrid/eventsGrid.scss
```diff
@@ -1,55 +1,50 @@
...
-  width: 15%;
-  vertical-align: top;
-  font-size: 13px;
-  font-family: $FONT-SEMIBOLD;
-  @media (max-width: $SCREEN_XS_MAX) {
-    overflow: hidden;
-    text-overflow: ellipsis;
-  }
-}
-.action-col:first-letter,
```
ElenaRomanchuk: Actually I removed only uppercase for the first letter (because now I use common translation for all values: in grid and in filter list).

PR #1498, opened 2019-05-15
EPMRPP-38788: invitation link redirect

Andrew-Omelchenko:


Andrew-Omelchenko:
There is was decided to keep all functionality as for 4.0 version, redirection to the login page won't be developed for now

PR #1500, opened 2019-05-15
EPMRPP-38319: set default level to Suite Level

margalencia:
Set level to Suite when current level is "" to display Suites toolbar


app/src/controllers/testItem/sagas.js
```diff
@@ -121,7 +122,7 @@ function* fetchTestItems({ payload = {} }) {
   );
   const dataPayload = yield take(createTestItemActionPredicate(NAMESPACE));
   const currentLevel = yield select(levelSelector);
-  const level = calculateLevel(dataPayload.payload.content) || currentLevel;
+  const level = calculateLevel(dataPayload.payload.content) || currentLevel || LEVEL_SUITE;
```
DiscoElevator: 1. Please rebase, this line was updated with other navigation fix.
2. This fix won't work for the case when we open a link with wrong item id. There should be special error screen for that (#1238).

PR #1507, opened 2019-05-16
EPMRPP-38772 - React: Discard, Clone, Save filter management icons sh…

mykhailo-kalashnikov-epam:
…ould be active in case at least one filtering option is filled in


app/src/pages/inside/common/launchFiltersToolbar/launchFiltersToolbar.jsx
```diff
@@ -13,7 +13,7 @@ import {
   changeActiveFilterAction,
 } from 'controllers/filter';
 import { changeLaunchDistinctAction, launchDistinctSelector } from 'controllers/launch';
-import { isEmptyObject } from 'common/utils';
+import { isEmptyObject, isEmptyValue } from 'common/utils';
```
DiscoElevator: `isEmptyValue` is not exported from `index.js` so we get `undefined` here.

PR #1504, opened 2019-05-16
EPRRPP-32108. React. Added validation for unique name for newly created widgets

twinfreaks:

PR #1509, opened 2019-05-16
EPMRPP-37474-Added validation for unique name for newly created dashboards

twinfreaks:


twinfreaks:
> Is [this](https://github.com/reportportal/service-ui/pull/1504) PR still actual?

Yes, it is another PR, another bug

PR #1510, opened 2019-05-17
EPMRPP-37238 SideBar fix, added customer isActive function to overrid…

tritseratops:
…e default behavior that did not work with hash in url


tritseratops:
will create another with black-jcj and package-lock.json

PR #1512, opened 2019-05-17
EPMRPP-37238 SideBar fix, added customer isActive function to overrid…

tritseratops:
…e default behavior that did not work with hash in url, changed redux-first-router version to "0.0.20-next" and redux-first-router-link to "2.1.1"


tritseratops:
Created another with 1 commit
 https://github.com/reportportal/service-ui/pull/1531

PR #1515, opened 2019-05-17
EPMRPP-38244 - React: Send request on GET status of launches are in progress every 5 seconds

KatsiarynaTatarynovich:
In case at least one launch is in progress status, the request on
GET http://host:port/api/v1/

{project_name}
/launch/status?ids=<launchId>
should be send every 5 seconds.

### DIFF
--- a/app/src/pages/inside/launchesPage/launchesPage.jsx
+++ b/app/src/pages/inside/launchesPage/launchesPage.jsx
@@ -3,6 +3,7 @@ import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { injectIntl, intlShape, defineMessages } from 'react-intl';
 import track from 'react-tracking';
+import isEqual from 'fast-deep-equal';
 import {
   LAUNCHES_PAGE,
   LAUNCHES_PAGE_EVENTS,
@@ -218,10 +219,23 @@ export class LaunchesPage extends Component {
     highlightItemId: null,
   };
 
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return prevState.prevLaunches !== nextProps.launches
+      ? {
+          prevLaunches: nextProps.launches,
+          launchesInProgress: nextProps.launches
+            .filter((item) => item.status === 'IN_PROGRESS')
+            .map((item) => item.id),
+        }
+      : null;
+  }
+
   state = {
     highlightedRowId: null,
     isGridRowHighlighted: false,
     isSauceLabsIntegrationView: false,
+    prevLaunches: [],
+    launchesInProgress: [],
   };
 
   componentDidMount() {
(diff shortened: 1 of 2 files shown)


app/src/pages/inside/common/durationBlock/durationBlock.jsx
```diff
@@ -82,13 +83,26 @@ export class DurationBlock extends Component {
...
   static defaultProps = {
     type: '',
     status: '',
+    id: null,
+    activeProject: '',
     itemNumber: null,
   };
+
+  componentDidMount() {
+    this.intervalId = setInterval(() => this.fetchLaunch(this.props.id), 5000);
```
DiscoElevator: I think it's better to do this on the page level. Moreover, I see in the api definition that this endpoint can work with an array of ids. Looks like you don't have to send a separate request for every launch.

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -214,7 +214,20 @@ export class LaunchesPage extends Component {
...
   };
 
+  componentDidUpdate() {
+    clearInterval(this.intervalId);
+
+    this.arrIdLaunchInProgress = this.props.launches
+      .filter((item) => item.status === 'IN_PROGRESS')
+      .map((item) => item.id);
+
+    if (this.arrIdLaunchInProgress.length) {
```
DiscoElevator: I think you should also check if launches list has changed. Otherwise, timeout will be reset even on launch selection.

```
if (/* check if we still have launches in progress and array has been changed and there's no active timeout (this.timeout === null) */) {
  // ...
}
```

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -214,7 +214,20 @@ export class LaunchesPage extends Component {
     deleteItemsAction: () => {},
   };
 
+  componentDidUpdate() {
+    clearInterval(this.intervalId);
+
+    this.arrIdLaunchInProgress = this.props.launches
```
DiscoElevator: I think it's better to place this in state via `computeDerivedStateFromProps` (only if launches list has changed)

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -214,7 +214,20 @@ export class LaunchesPage extends Component {
     deleteItemsAction: () => {},
   };
 
+  componentDidUpdate() {
+    clearInterval(this.intervalId);
```
DiscoElevator:
```
if (/* there's no launches in progress and this.intervalId not empty */) {
  // clear interval
  this.intervalId = null;
}
```

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -335,6 +348,22 @@ export class LaunchesPage extends Component {
     });
   };
 
+  fetchLaunchStatus = (arrIdLaunchInProgress) => {
+    if (!this.arrIdLaunchInProgress.length) {
+      clearInterval(this.intervalId);
```
DiscoElevator: It's done in `componentDidUpdate`

app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -218,10 +219,23 @@ export class LaunchesPage extends Component {
     highlightItemId: null,
   };
 
+  static getDerivedStateFromProps(nextProps, prevState) {
+    return prevState.launches !== nextProps.launches
+      ? {
+          launches: nextProps.launches,
```
DiscoElevator: `prevLaunches`

PR #1518, opened 2019-05-20
Fix/EPMRPP-37764: clickable area firefox

Andrew-Omelchenko:


app/src/pages/inside/dashboardItemPage/widgetsGrid/widgetsGrid.jsx
```diff
@@ -211,6 +211,7 @@ export class WidgetsGrid extends Component {
               isDraggable={this.props.isModifiable}
               isResizable={this.props.isModifiable}
               draggableHandle=".draggable-field"
+              useCSSTransforms={false}
```
AmsterGet: I read [here](https://github.com/STRML/react-grid-layout) that this approach can decrease our performance in other browsers (see the screenshot below).
image
@DiscoElevator what you do think? Can we adjust this fix only for FireFox?

Andrew-Omelchenko: Fixed with adding the ability to disable `useCSSTransforms` feature in Firefox browser only.

PR #1530, opened 2019-05-21
EPMRPP-38914 - React. Dropdown menu is located incorrectly in list view for Edge browser

KatsiarynaTatarynovich:


app/src/components/buttons/dotsMenuButton/dotsMenuButton.jsx
```diff
@@ -130,22 +131,30 @@ export class DotsMenuButton extends Component {
...
+                    className={cx('menu', {
+                      opened: this.state.opened,
+                    })}
+                  >
+                    {this.renderMenuItems(items)}
+                  </div>
+                );
+              }}
+            </Popper>,
+            document.querySelector('#dots-menu-root'),
```
DiscoElevator: I think "#popover-root" can be used here

PR #1553, opened 2019-05-27
EPMRPP-38847. Regex editor with syntax highlighting.

ElenaRomanchuk:
chrome_JOSTTTCXQ2

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -573,7 +573,7 @@
         },
         "style-loader": {
           "version": "0.20.3",
-          "resolved": "http://registry.npmjs.org/style-loader/-/style-loader-0.20.3.tgz",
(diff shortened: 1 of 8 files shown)


app/src/common/css/variables/colors.scss
```diff
@@ -43,3 +43,16 @@ $COLOR--organish: #fb7848;
 $COLOR--dirty-green: #5f722b;
 $COLOR--greenish-grey: #b8d0b0;
 $COLOR--go-to-stack-trace: #fafa89;
+
+$COLOR--REGEXP-symbol: #3c00c7;
```
AmsterGet: @DiscoElevator think that it is better to define these variables within component styles, because this is not our general RP style guide and will not be used anywhere else.

PR #1556, opened 2019-05-27
EPMRPP-38965 - REACT. Investigated Percentage Widget. Clickable Area doesn't transfer the user to step level

KatsiarynaTatarynovich:


app/src/components/widgets/charts/investigatedTrendChart/investigatedTrendChart.jsx
```diff
@@ -65,6 +85,16 @@ export class InvestigatedTrendChart extends Component {
       this.props.observer.unsubscribe('widgetResized', this.resizeChart);
   }
 
+  onChartClick = (data) => {
+    const { widget, statisticsLink } = this.props;
+    const id = this.isTimeline
```
AmsterGet: Be careful with timeline mode. In this case user should be redirected to the launches page with new filter (according to our 4th version).

app/src/components/widgets/charts/investigatedTrendChart/investigatedTrendChart.jsx
```diff
@@ -136,12 +175,43 @@ export class InvestigatedTrendChart extends Component {
...
 
     this.setState({
       isConfigReady: true,
     });
   };
 
+  timeLineModeClickHandler = (data) => {
+    const { widget } = this.props;
+
+    const arrResult = Object.keys(widget.content.result).map((item) => item);
```
AmsterGet: I think you can create common util function to get new filter, f.e. `getUpdatedFilterWithTime(chartFilter, itemDate)` and reuse it here and in the `launchStatisticsChart`

PR #1558, opened 2019-05-27
EPMRPP-35178: React: Status bar is not available in case no statistic…

KatsiarynaTatarynovich:


app/src/pages/inside/common/infoLine/barChart/barChart.scss
```diff
@@ -4,6 +4,7 @@
   width: 100px;
   height: 10px;
   border-radius: 100px;
+  background-color: $COLOR--skipped;
```
AmsterGet: According to our [4th](http://dev.epm-rpp.projects.epam.com:8080/ui/#amsterget_personal/launches/all%7Cpage.page=1&page.size=15&page.sort=start_time,number%2CDESC/5cecf8d1a3bbc40001cbfb57?page.page=1&page.size=15&page.sort=start_time%2CASC) version color `$COLOR--white` should be used.

PR #1561, opened 2019-05-27
EPMRPP-38852 - React: Navigation via Next/Previous buttons from log v…

mykhailo-kalashnikov-epam:
…iew does not work correctly for drill down via list view

### DIFF
--- a/app/src/controllers/testItem/sagas.js
+++ b/app/src/controllers/testItem/sagas.js
@@ -38,6 +38,7 @@ import {
   isLostLaunchSelector,
   createParentItemsSelector,
   itemsSelector,
+  breadcrumbsSelector,
 } from './selectors';
 import { calculateLevel } from './utils';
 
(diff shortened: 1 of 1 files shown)


app/src/controllers/testItem/sagas.js
```diff
@@ -118,10 +125,10 @@ function* fetchTestItems({ payload = {} }) {
   );
   const dataPayload = yield take(createFetchPredicate(NAMESPACE));
   let level;
-  if (!parentItem) {
```
DiscoElevator: Why did you revert this change?

mykhailo-kalashnikov-epam: This change is no longer necessary, SUITE_LEVEL defined as default, also if we go to log page from statistic - level changing to LEVEL_NOT_FOUND which broke navigation logic on logPage.

PR #1568, opened 2019-05-28
EPMRPP-39070 - React: History table grid is broken on STEP view of the launch

mykhailo-kalashnikov-epam:


app/src/pages/inside/historyPage/historyView/historyTable/historyTable.jsx
```diff
@@ -49,16 +67,111 @@ export class HistoryTable extends Component {
...
+        issue: filteredLaunchHistoryItem[0].issue,
+        defects: filteredLaunchHistoryItem[0].statistics.defects,
+        itemIds,
+      };
+    }
+    return itemProps;
   };
 
-  isTheBigDepth = () => this.props.history.length > OPTIMAL_HISTORY_DEPTH_FOR_RENDER;
+  getCorrespondHistoryItem = (historyItemProps, currentHistoryItem, launchId) => {
```
DiscoElevator: _corresponding_

app/src/pages/inside/historyPage/historyView/historyTable/historyTable.jsx
```diff
@@ -30,15 +36,27 @@ const messages = defineMessages({
...
   (state) => ({
     items: itemsHistorySelector(state),
     history: historySelector(state),
     visibleItemsCount: visibleItemsCountSelector(state),
+    link: (ownProps) => nameLinkSelector(state, ownProps),
   }),
-  { fetchItemsHistoryAction },
+  {
+    fetchItemsHistoryAction,
+    redirectToTestItem: redirect,
```
DiscoElevator: I think it shouldn't be redirect because we can't use "back" button in this case

PR #1570, opened 2019-05-28
EPMRPP-39092 - React: Update 'Configuration examples' with parameter 'attribute' on Profile page

KatsiarynaTatarynovich:
Expected:
'Tags' should be replaced with 'attributes'


app/src/pages/inside/profilePage/configExamplesBlock/tabsConfig.jsx
```diff
@@ -19,7 +19,7 @@ const javaConfig = (token, activeProject) => ({
       <div className={cx('options')}>
         <p>rp.enable = true</p>
         <p>rp.description = My awesome launch</p>
-        <p>rp.tags = TAG1;TAG2</p>
+        <p>rp.attributes = TAG1;TAG2</p>
```
AmsterGet: Please update it for all configuration examples. Also please clarify with QA team how we should show default values for this examples. Now attributes are array with the next structure:
```
{
key: string,
value: string,
system: boolean
}
```

PR #1571, opened 2019-05-28
EPMRPP-37001 Implemented percentage-of-investigations widget on statu…

twinfreaks:
…s page


app/src/components/widgets/charts/investigatedTrendChart/statusPageModeConfig.jsx
```diff
@@ -0,0 +1,142 @@
...
+import { TimelineTooltip } from './timelineTooltip';
+import { MESSAGES } from './common/constants';
+
+export const mapMinListLength = {
+  '1M': 30,
+  '3M': 12,
+  '6M': 24,
+};
+
+const renderTooltip = (itemData, intl) => (d, defaultTitleFormat, defaultValueFormat, color) => {
```
AmsterGet: Seems that it is duplicates with `renderTooltip` in the  `../timelineConfig`.
It will be better to create a common method for this two modes

app/src/components/widgets/charts/investigatedTrendChart/statusPageModeConfig.jsx
```diff
@@ -0,0 +1,142 @@
+import ReactDOMServer from 'react-dom/server';
+import { getItemColor } from 'components/widgets/charts/common/utils';
+import { TimelineTooltip } from './timelineTooltip';
+import { MESSAGES } from './common/constants';
+
+export const mapMinListLength = {
+  '1M': 30,
```
AmsterGet: Seems that this constants already used in the `pages/admin/projectStatusPage/constants.js` (`PERIOD_VALUES`).
Please move it upper (f.e. to `common/constants/statusPeriodValues.js`) and use everywhere across the PR

app/src/pages/admin/projectStatusPage/statusPageContent/widgets/investigated/investigated.jsx
```diff
@@ -0,0 +1,82 @@
...
+    interval: '3M',
+  };
+
+  state = {
+    containerEl: null,
+  };
+
+  componentDidMount() {
+    // eslint-disable-next-line react/no-did-mount-set-state
+    this.setState({ containerEl: this.myRef.current });
```
AmsterGet: Is this field will change its value across component life cycle?
I think it will be better to use class field property instead.

app/src/pages/admin/projectStatusPage/statusPageContent/widgets/investigated/investigated.jsx
```diff
@@ -0,0 +1,82 @@
...
+          <InvestigatedTrendChart
+            widget={this.prepareData(data, interval)}
+            interval={interval}
+            container={containerEl}
+            withoutBackground
+            onStatusPageMode
+          />
+        ) : (
+          <div className={cx('no-data-wrapper')}>
+            <NoItemMessage message={intl.formatMessage(messages.noDataMessage)} />
```
AmsterGet: Also please use our `components/widgets/noDataAvailable` instead (see the screenshot below from the 4th version)
image

app/src/pages/admin/projectStatusPage/statusPageContent/widgets/investigated/investigated.jsx
twinfreaks: This field changes only ones after the first render. There is only 1 alternative way to implement this: to create additional wrapper component around this component to pass only 1 property down.

DiscoElevator: Where does it change? Usually refs are not placed in state.

DiscoElevator: Where does it change? Usually refs are not placed in state.

twinfreaks: I don't have a ref at first render when I'm trying to pass ref to the inner component. Ref is being created after render is finished. Do you suggest to add an additional wrapper to pass updated ref as property down?

DiscoElevator: I think some flag would be better

PR #1575, opened 2019-05-29
EPMRPP-39090: all subtypes of group should be selected

Andrew-Omelchenko:


Andrew-Omelchenko:
Refactor `DefectStatistics` to class component

PR #1581, opened 2019-05-29
EPMRPP-38848. PA label on step and log view.

ElenaRomanchuk:
Added PA label (works as designed).
AA label doesn't consist corresponding icon because I didn't get it in time. So I'll create extra PR for adding icon later.

### DIFF
--- a/app/src/pages/inside/stepPage/stepGrid/defectType/patternAnalyzedLabel/patternAnalyzedLabel.jsx
+++ b/app/src/pages/inside/stepPage/stepGrid/defectType/patternAnalyzedLabel/patternAnalyzedLabel.jsx
@@ -1,0 +1,22 @@
+import classNames from 'classnames/bind';
+import PropTypes from 'prop-types';
+import { withTooltip } from 'components/main/tooltips/tooltip';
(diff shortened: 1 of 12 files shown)


app/src/pages/inside/stepPage/stepGrid/defectType/patternAnalyzedLabel/patternAnalyzedLabel.jsx
```diff
@@ -0,0 +1,23 @@
+import classNames from 'classnames/bind';
+import PropTypes from 'prop-types';
+import { withHoverableTooltip } from 'components/main/tooltips/hoverableTooltip';
+import { PatternAnalyzedTooltip } from './patternAnalyzedTooltip';
+import styles from './patternAnalyzedLabel.scss';
+
+const cx = classNames.bind(styles);
+
+export const PatternAnalyzedLabel = withHoverableTooltip({
```
AmsterGet: Should this tooltip be hoverable? I don't see any action in the body of the tooltip

ElenaRomanchuk: Fixed.

PR #1586, opened 2019-05-30
EPMRPP-39059. Pattern analysis tab. Access restrictions.

ElenaRomanchuk:


app/src/components/inputs/inputSwitcher/inputSwitcher.jsx
```diff
@@ -25,20 +25,26 @@ import styles from './inputSwitcher.scss';
...
-export const InputSwitcher = ({ children, value, onChange, onFocus, onBlur }) => {
+export const InputSwitcher = ({ children, value, onChange, onFocus, onBlur, readOnly }) => {
   const sliderClasses = cx({
     'switcher-slider': true,
     centered: !children,
     enabled: value,
   });
   const onChangeHandler = (e) => {
-    onChange(e.target.checked);
+    if (!readOnly) onChange(e.target.checked);
```
AmsterGet: I think we can block it by CSS using `readonly` class name on `label` with property `pointer-events: none`.

DiscoElevator: According to MDN, `pointer-events` usage on HTML elements is experimental and may be changed in the future. At this moment, it's a part of the SVG standard. I think we may leave JS check for `readonly`.

AmsterGet: You're right! This has some risks. So leave this approach.

PR #1591, opened 2019-05-31
EPMRPP-39111 - React: Request on GET launches is not sent on refreshi…

mykhailo-kalashnikov-epam:
…ng the page with no results


app/src/controllers/launch/sagas.js
```diff
@@ -40,7 +40,12 @@ function* fetchLaunchesWithParams({ payload }) {
 const notEmptyConditionsPredicate = ({ value }) => !isEmptyValue(value);
 
 function* fetchLaunches() {
+  const firstLoad = yield select(firstTimeLoad);
   const filterId = yield select(filterIdSelector);
+  if (firstLoad && filterId < 0) {
```
DiscoElevator: I think it's better to find filter with id < 0 (`launchFiltersSelector`). If filter doesn't exist, change active filter to ALL

PR #1592, opened 2019-05-31
EPMRPP: 39049 - defect types do not have standard order

Andrew-Omelchenko:


Andrew-Omelchenko:
On v4 "Overall Statistics" defect types are sorted. So, in this PR I sort them too
Screen Shot 2019-06-04 at 6 41 37 PM

PR #1600, opened 2019-06-03
EPMRPP-39150 Router links highligth fix

AmsterGet:


app/src/components/main/navLink/navLink.jsx
```diff
@@ -0,0 +1,92 @@
...
+import PropTypes from 'prop-types';
+import { NavLink } from 'redux-first-router-link';
+import { actionToPath, getOptions, history, selectLocationState } from 'redux-first-router';
+import matchPath from 'rudy-match-path';
+import { stripBasename } from 'rudy-history/PathUtils';
+
+@connect((state) => ({
+  location: selectLocationState(state),
+}))
+export default class NavLinkWrapper extends Component {
```
DiscoElevator: Exported component name should match file name. I suggest exporting it as `NavLinkWrapper` and export with renaming in `index.js`

AmsterGet: Fixed

PR #1609, opened 2019-06-05
EPMRPP-38965 - REACT. Investigated Percentage Widget. Clickable Area doesn't transfer the user to step level

KatsiarynaTatarynovich:


app/src/components/widgets/charts/investigatedTrendChart/investigatedTrendChart.jsx
```diff
@@ -45,7 +46,9 @@ const cx = classNames.bind(styles);
 @connect(
   (state) => ({
     projectId: activeProjectSelector(state),
-    statisticsLink: statisticsLinkSelector(state, {
+    defectTypes: defectTypesSelector(state),
+    getDefectLink: (params) => defectLinkSelector(state, params),
+    getStatisticsLink: statisticsLinkSelector(state, {
```
DiscoElevator: statisticsLink

app/src/components/widgets/charts/investigatedTrendChart/investigatedTrendChart.jsx
```diff
@@ -72,6 +77,8 @@ export class InvestigatedTrendChart extends Component {
 
   static defaultProps = {
     navigate: () => {},
+    getDefectLink: () => {},
+    getStatisticsLink: () => {},
```
DiscoElevator: 1. looks like it's an object (see line 68: `getStatisticsLink: PropTypes.object.isRequired`)
2. If it's required, you don't need default props

app/src/components/widgets/charts/investigatedTrendChart/investigatedTrendChart.jsx
```diff
@@ -122,6 +129,17 @@ export class InvestigatedTrendChart extends Component {
     this.chart.toggle(id);
   };
 
+  getDefectTypeLocators = (id) => {
+    const { defectTypes } = this.props;
+    const investigatedDefectType = ['AUTOMATION_BUG', 'NO_DEFECT', 'PRODUCT_BUG', 'SYSTEM_ISSUE'];
```
DiscoElevator: You can use constants from app/src/common/constants/defectTypes.js

PR #1613, opened 2019-06-05
EPMRPP-36999: Implemented launches quantity widget on Status page

twinfreaks:


app/src/common/constants/dateFormats.js
```diff
@@ -0,0 +1 @@
+export const DEFAULT_DATE_FORMAT = 'YYYY-MM-DD';
```
AmsterGet: Please rename it and move to the `common/constants/timeDateFormat.js`

PR #1623, opened 2019-06-07
EPMRPP-39049: reopened legend defects order

Andrew-Omelchenko:
the legend of the widgets

### DIFF
--- a/app/src/controllers/project/selectors.js
+++ b/app/src/controllers/project/selectors.js
@@ -1,6 +1,6 @@
 import { createSelector } from 'reselect';
 import { OWNER } from 'common/constants/permissions';
-import { DEFECT_TYPES_SEQUENCE } from 'common/constants/defectTypes';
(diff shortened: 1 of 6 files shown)


app/src/controllers/project/selectors.js
```diff
@@ -48,6 +48,22 @@ export const defectTypesSelector = createSelector(subTypesSelector, (subTypes) =
   ),
 );
 
+export const legendDefectTypesSelector = createSelector(subTypesSelector, (subTypes) => {
+  const PREFIX = 'statistics$defects$';
+  const result = [
+    'statistics$executions$total',
```
DiscoElevator: Execution statistics doesn't belong to defect types. Why is it here?

Andrew-Omelchenko: Maybe, misleading name for the selector was chosen. Something like `orderedContentFieldsSelector` will be more appropriate in this case. In legend of some widgets execution statistics is used as well as defect types statistics.

PR #1641, opened 2019-06-11
EPMRPP-39292. Prevent opening during editing launch

ElenaRomanchuk:


ElenaRomanchuk:
Actually, I dislike this decision, but I don't have any other ideas.

PR #1660, opened 2019-06-14
EPMRPP-38910: only relevant fields with zero values

Andrew-Omelchenko:
relevant fields with zero values should be displayed

### DIFF
--- a/app/src/components/widgets/charts/overallStatisticsChart/overallStatisticsPanel.jsx
+++ b/app/src/components/widgets/charts/overallStatisticsChart/overallStatisticsPanel.jsx
@@ -1,7 +1,7 @@
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import classNames from 'classnames/bind';
-import { orderedContentFieldsSelector } from 'controllers/project';
+import { orderedDefectFieldsSelector } from 'controllers/project';
 import { TotalStatistics } from './totalStatistics';
 import { OverallDefects } from './overallDefects';
 import styles from './overallStatistics.scss';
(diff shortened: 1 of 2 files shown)


Andrew-Omelchenko:
Fixed.

app/src/components/widgets/charts/overallStatisticsChart/overallStatisticsPanel.jsx
```diff
@@ -26,17 +26,33 @@ export class OverallStatisticsPanel extends React.PureComponent {
...
+      .map((key) => ({ key, value: values[key] || 0 }));
   };
 
-  render() {
-    const { widget, isPreview } = this.props;
+  getTotals = () => {
+    const { widget } = this.props;
     const values = widget.content.result[0].values;
+    const fields = widget.contentParameters.contentFields.filter(
+      (field) => field.indexOf('executions') !== -1,
```
DiscoElevator: I think you should also filter values by their selected/unselected status to support the following case from EPMRPP-38910 (fields should be hidden ONLY in case when user deselects them in the widget settings). This case now is broken by changes in this PR.

Andrew-Omelchenko: When user unselects some values, those values will automatically disappear from `widget.contentParameters.contentFields` array and don't be displayed on the widget, so I think it's redundant to additionally filter them. If server doesn't return some values, but user selects them to display, widget shows zeroes (line 43).

DiscoElevator: In case when field exists in `contentFields`, but doesn't exists in `values` it shows incorrectly:
image

JSX conditions like `{passed && ...` in case `passed = 0` render as `0`.

Andrew-Omelchenko:
Oops, sorry. I missed this case. Fixed.

PR #1675, opened 2019-06-18
EPMRPP-38839: incorrect hyphenation of launch name

Andrew-Omelchenko:


app/src/pages/inside/common/itemInfo/itemInfo.scss
```diff
@@ -30,15 +30,15 @@
     font-size: 16px;
   }
 }
-.name{
-  word-break: break-all;
+.name {
+  word-break: break-word;
```
DiscoElevator: Not every browser supports this value.
```
  word-break: break-all;

    word-break: break-word;
  }
```

Andrew-Omelchenko: Fixed.

PR #1676, opened 2019-06-18
EPMRPP-39353: incorrect time values when filtering launches

Andrew-Omelchenko:


app/src/components/filterEntities/entityItemStartTime.jsx
```diff
@@ -39,12 +33,9 @@ const presets = [
     value: {
       start: moment()
         .startOf('day')
-        .subtract(1, 'weeks')
+        .subtract(6, 'days')
```
DiscoElevator: Why 6 days?

app/src/components/filterEntities/entityItemStartTime.jsx
```diff
@@ -53,18 +44,15 @@ const presets = [
     value: {
       start: moment()
         .startOf('day')
-        .subtract(1, 'months')
+        .subtract(29, 'days')
```
DiscoElevator: Why 29 days? Months have different day count.

app/src/components/filterEntities/entityItemStartTime.jsx
Andrew-Omelchenko: Expected:
"filter.btw.start_time":"-1440;362;+0300" ('1440' is number of minutes of the previous day,'362' is number of minutes till the current moment of filtering and +0300 is the user's timezone)
Launches with start time in that range are shown. It was for Last 2 days. So for Last 7 days we should send "-1440*6;number of minutes till the current moment of filtering;+0300"

app/src/components/filterEntities/entityItemStartTime.jsx
Andrew-Omelchenko: In dropdown user sees not 1 month, but `30 days`.

DiscoElevator:
Waiting for clarification with QA and BA

Andrew-Omelchenko:
Fixed. Exactly as in v4.

PR #1677, opened 2019-06-18
EPMRPP-38601: Move user credentials to request body

DiscoElevator:


DiscoElevator:
Waiting for WS to investigate possibility of changing content type to json

PR #1681, opened 2019-06-18
EPMRPP-39298-fixed-add-edit-modal-share-button-layout

twinfreaks:


app/src/pages/inside/dashboardPage/modals/addEditModal/addEditModal.scss
```diff
@@ -47,3 +47,7 @@
     }
   }
 }
+
+.share-switcher-wrapper > div > label {
```
DiscoElevator: What if component's layout will change? I suggest changing `BigSwitcher` root element style from `display: flex` to `display:inline-flex`

twinfreaks: I'm afraid that this could affect some other places where BigSwitcher is used

DiscoElevator: It's better to fix the root cause of a problem than try to fix many similar local bugs

app/src/pages/inside/dashboardPage/modals/addEditModal/addEditModal.jsx
```diff
@@ -196,6 +196,7 @@ export class AddEditModal extends Component {
           <ModalField
             label={intl.formatMessage(messages.dashboardShareLabel)}
             labelWidth={labelWidth}
+            className={cx('share-switcher-wrapper')}
```
DiscoElevator: This class is not needed anymore

PR #1694, opened 2019-06-21
EPMRPP-37335 REACT No pop-up description of the saved filter fixed

tritseratops:
added 2 more icon with tooltips elements


tritseratops:
Found that for newly saved filters schema does not work.

PR #1709, opened 2019-06-26
EPMRPP-38589 Fixed delete widget modal name overflow

twinfreaks:


AmsterGet:
Please also provide `word-break: break-word` property if it supports by user's browser:
 ``` word-break: break-all;

    word-break: break-word;
  }
```

PR #1712, opened 2019-06-27
Epmrpp 39195 create widget filter by default

twinfreaks:


twinfreaks:
Need to create new PR with squashed commits

PR #1713, opened 2019-06-27
EPMRPP-39195 fixed default filter when widget was created from the la…

twinfreaks:
…unches page


DiscoElevator:
Closed in favor of #1714

PR #1725, opened 2019-07-03
EPMRPP-38645-Launch-statistics-on-chartjs

vitalii-chornyi:


app/src/components/widgets/charts/launchStatisticsChart/chartJsConfig.js
```diff
@@ -0,0 +1,110 @@
...
+          isTimeLine
+            ? tooltipItem[0].label
+            : `${result[tooltipItem[0].index].name} ${tooltipItem[0].label}`,
+
+        label(tooltipItem, data) {
+          const dataset = data.datasets[tooltipItem.datasetIndex];
+          const label = messages[dataset.label]
+            ? formatMessage(messages[dataset.label])
+            : dataset.label;
+          const value = -(-dataset.data[tooltipItem.index]);
```
DiscoElevator: `const value = Number(dataset.data[tooltipItem.index]);`

vitalii-chornyi: Fixed

vitalii-chornyi: 1,2 are fixed

vitalii-chornyi:
> 1. I see the legend in widget wizard
>    изображение
> 2. widget in preview mode works as non-preview (has tooltips, chick handlers)
> 3. Defect types sorting in legend and the widget is not correct:
>    изображение
>    DEV version:
>    изображение
> 4. Areas have no tooltips and not clickable (tooltips are shown only at chart points)

1,2 are fixed

PR #1727, opened 2019-07-04
EPMRPP-38255. Log with attachments for launches and test items.

ElenaRomanchuk:


ElenaRomanchuk:
I have just found a way to reuse common logs code base. So, i'll redo this request.

PR #1745, opened 2019-07-08
EPMRPP-39203. Single modal for edit and view test item details for step.

ElenaRomanchuk:
image


app/src/pages/inside/stepPage/modals/testItemDetailsModal/testItemDetailsModal.jsx
```diff
@@ -44,105 +56,221 @@ const messages = defineMessages({
...
+                {Parser(IconDuplicate)}
+              </CopyToClipboard>
             )}
-        </div>
-        {item.parameters && (
-          <div className={cx('row')}>
-            <TestParameters parameters={item.parameters} />
           </div>
+        </ModalField>
+        {item.tags &&
```
DiscoElevator: Tags are not used anymore, so please remove this code (they were replaced with attributes)

app/src/pages/inside/stepPage/modals/testItemDetailsModal/testItemDetailsModal.scss
```diff
@@ -1,16 +1,26 @@
...
+  justify-content: space-between;
   padding-top: 10px;
   padding-bottom: 10px;
-  position: relative;
+  font-family: $FONT-SEMIBOLD;
+  font-size: 14px;
+}
+
+.id {
+  word-wrap: break-word;
```
DiscoElevator: Doesn't work in some browsers (i.e. Firefox, IE).
Please also add a fallback:
```
  word-break: break-all;

    word-break: break-word;
  }
```

app/src/pages/inside/stepPage/modals/testItemDetailsModal/testItemDetailsModal.scss
```diff
@@ -1,16 +1,26 @@
...
+}
+
+.location {
+  display: flex;
+  justify-content: space-between;
+  white-space: pre-line;
+  word-break: break-word;
```
DiscoElevator: The same as above: fallback needed

PR #1746, opened 2019-07-08
EPMRPP-39499: REACT. Update for description isn't start from the new line

Andrew-Omelchenko:


app/src/pages/inside/common/modals/editItemsModal/editItemsModal.jsx
```diff
@@ -257,7 +257,7 @@ export class EditItemsModal extends Component {
     if (descriptionAction !== DESCRIPTION_LEAVE) {
       data.description = {
         action: descriptionAction,
-        comment: description || '',
+        comment: `${DESCRIPTION_UPDATE ? '\n' : ''}${description}` || '',
```
DiscoElevator: In case of empty description, it will add an unnecessary empty line.
By the way, I'm not sure if it should be done on UI or WS side. Need to discuss.

DiscoElevator: Decided to do this on UI

PR #1753, opened 2019-07-08
EPMRPP-39446 Investigated percentage of launches legend fix

AmsterGet:
Issue with disappearing bar charts (based on c3.js) has been fixed as a part of this task using Page visibility API.


app/src/components/widgets/charts/investigatedTrendChart/launchModeConfig.jsx
```diff
@@ -14,25 +14,27 @@ const renderTooltip = (itemData, intl) => (d, defaultTitleFormat, defaultValueFo
...
     name: item.name,
     number: item.number,
     startTime: item.startTime,
   }));
+  const groups = Object.keys(sortedResult[0].values);
 
-  const columns = DEFECT_TYPES.map((type) => {
+  const columns = groups.map((type) => {
     const values = sortedResult.map((item) => item.values[type] || 0);
+    colors[type] = COLORS[type];
```
DiscoElevator: `map` should not have any side-effects.

AmsterGet: Fixed

PR #1782, opened 2019-07-16
EPMRPP-39750 Launches page selected items fix

AmsterGet:


app/src/pages/inside/launchesPage/launchesPage.jsx
```diff
@@ -271,7 +271,6 @@ export class LaunchesPage extends Component {
 
   componentWillUnmount() {
     clearInterval(this.intervalId);
-    this.props.unselectAllLaunchesAction();
```
DiscoElevator: I prefer keeping this. If we leave launches page, it feels natural to clear state we don't need (also, it is common behavior for test item pages)

AmsterGet: Fixed

PR #1799, opened 2019-07-19
EPMRPP-39759. Fix attachments loading when navigating the history line.

ElenaRomanchuk:


app/src/controllers/log/attachments/sagas.js
```diff
@@ -31,9 +31,8 @@ function* getAttachmentURL() {
   }
   const logViewMode = yield select(logViewModeSelector);
   if (logViewMode === DETAILED_LOG_VIEW) {
-    const itemIdsArray = yield select(testItemIdsArraySelector);
-    const underPathItemsIds = itemIdsArray.slice(1).join('.');
-    return URLS.logsUnderPath(activeProject, underPathItemsIds);
+    const activeRetry = yield select(activeRetrySelector);
```
AmsterGet: I think will be good to select `activeRetry` [here](https://github.com/reportportal/service-ui/pull/1799/files#diff-c334c7297dda5b7cff0603a10390c49eR28) to avoid one unnecessary select action

PR #1816, opened 2019-07-23
EPMRPP-39629: Fix cumulative wizard edit form

vitalii-chornyi:


app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/controls/filtersControl/filtersControl.jsx
```diff
@@ -134,6 +134,9 @@ export class FiltersControl extends Component {
   }
 
   getFormAppearanceComponent = (activeFilter) => {
+    if (!activeFilter) {
+      this.clearFormAppearance();
```
DiscoElevator: I don't think it's a good place for a side-effect here. This is just a getter function

vitalii-chornyi:
> Filter controls are inactive. I think it unexpected (see the screenshot below).
> image
> I think we can fix this issue as in 4th version (see the screenshot below).
> image

@AmsterGet Why do we propose "to choose another filter (edit icon)" if there are no filters on the project?

AmsterGet:
> > Filter controls are inactive. I think it unexpected (see the screenshot below).
> > image
> > I think we can fix this issue as in 4th version (see the screenshot below).
> > image
> 
> @AmsterGet Why do we propose "to choose another filter (edit icon)" if there are no filters on the project?

@vitalii-chornyi As in 4th version user can create the new one from wizard, so this message can be used. After clicking the edit icon user will see the empty filter list with `Add filter` button (see the screenshot below).
image

app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/controls/filtersControl/filtersControl.jsx
vitalii-chornyi: Fixed

vitalii-chornyi:
> > > Filter controls are inactive. I think it unexpected (see the screenshot below).
> > > image
> > > I think we can fix this issue as in 4th version (see the screenshot below).
> > > image
> > 
> > 
> > @AmsterGet Why do we propose "to choose another filter (edit icon)" if there are no filters on the project?
> 
> @vitalii-chornyi As in 4th version user can create the new one from wizard, so this message can be used. After clicking the edit icon user will see the empty filter list with `Add filter` button (see the screenshot below).
> image

Fixed

PR #1830, opened 2019-07-25
EPMRPP-38466 - Added attribute key field with auto-complete

Andrew-Omelchenko:


app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/productStatusControls.jsx
```diff
@@ -195,6 +204,8 @@ export class ProductStatusControls extends Component {
         </FieldProvider>
         <FieldProvider
           name="contentParameters.widgetOptions.customColumns"
+          format={this.formatCustomColumns}
+          parse={this.parseCustomColumns}
           validate={validators.customColumns}
         >
           <CustomColumnsControl />
```
AmsterGet: I think it would be good to provide a search URL here to get it from props in the `CustomColumnItem `.

Andrew-Omelchenko: Fixed.

PR #1842, opened 2019-07-29
EPMRPP-39674 - React: Incorrect response of delete(x) icon for "issues" in the Log section

Andrew-Omelchenko:


app/src/pages/inside/logsPage/logItemInfo/logItemInfo.jsx
```diff
@@ -206,6 +213,24 @@ export class LogItemInfo extends Component {
     });
   };
 
+  handleUnlinkTicket = (logItem) => (ticketId) => {
```
AmsterGet: You don't need to create a function that returns another function.
You already have access to `logItem` from `props`.

Andrew-Omelchenko: Fixed.

PR #1851, opened 2019-07-30
EPMRPP-39206. React. Discard button is enabled by default for cloned …

tritseratops:
…filter. Added dirtyFilterIdsSelector to controllers/filter/selectors/js. Added isFilterDirty function to launchFilterToolbar, updated isDiscardDisabled to use that function isFilterDirty.

### DIFF
--- a/app/src/controllers/filter/selectors.js
+++ b/app/src/controllers/filter/selectors.js
@@ -40,3 +40,15 @@ export const unsavedFilterIdsSelector = createSelector(
       })
       .map((item) => item.id),
 );
+
+export const dirtyFilterIdsSelector = createSelector(
+  launchFiltersSelector,
+  userFiltersSelector,
+  (filters, savedFilters) =>
+    filters
+      .filter((filter) => {
+        const savedFilter = savedFilters.find((item) => item.id === filter.id);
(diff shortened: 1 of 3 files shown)


app/src/controllers/filter/selectors.js
```diff
@@ -40,3 +40,21 @@ export const unsavedFilterIdsSelector = createSelector(
...
+
+export const dirtyFilterIdsSelector = createSelector(
+  launchFiltersSelector,
+  userFiltersSelector,
+  (filters, savedFilters) =>
+    filters
+      .map((filter) => ({
+        ...filter,
+        conditions: filter.conditions.filter(
+          (condition) => !(condition.filteringField === 'name' && condition.value === ''),
```
DiscoElevator: What is this condition for?

tritseratops: If you mean condition in line 52 - it is avoid filters with empty fields. This is a good question. it might happen that this condition will disable Discard button if we remove all data from name condition. Let me check this option. I will return back to you.

tritseratops: I updated commit. That condition was redundant, I deleted it. I copied it from other selector. Still curious what was it doing.

PR #1869, opened 2019-08-02
EPMRPP-39875. Dashboard. Print page.

ElenaRomanchuk:


app/src/pages/inside/dashboardItemPage/dashboardItemPage.jsx
```diff
@@ -343,6 +349,19 @@ export class DashboardItemPage extends Component {
                     {formatMessage(messages.delete)}
                   </GhostButton>
                 )}
+                <GhostLink
```
AmsterGet: According to the story these buttons should be hidden and corresponding actions should be moved to the `actions` using `GhostMenuButton` (see the screenshot below).
image

AmsterGet: Task description has been changed. Leave it as it is.

PR #1870, opened 2019-08-02
EPMRPP-39458 fixed maxlength on add filter input fields

twinfreaks:


app/src/components/inputs/inputConditional/inputConditional.jsx
```diff
@@ -124,7 +127,7 @@ export class InputConditional extends Component {
           value={value.value}
           placeholder={placeholder}
           disabled={disabled}
-          maxLength={maxLength}
+          maxLength={inputprops.maxlength ? inputprops.maxlength : maxLength}
```
DiscoElevator: Why not just use `maxLength` prop? What's the difference?

app/src/pages/admin/allUsersPage/usersToolbar/usersEntities.jsx
```diff
@@ -100,6 +100,11 @@ export class UsersEntities extends Component {
         title: intl.formatMessage(messages.name),
         active: true,
         removable: false,
+        customProps: {
```
DiscoElevator: Why not `maxlength: 256,`?

app/src/pages/admin/allUsersPage/usersToolbar/usersEntities.jsx
```diff
@@ -181,6 +186,9 @@ export class UsersEntities extends Component {
         removable: false,
         customProps: {
           placeholder: null,
+          inputprops: {
```
DiscoElevator: Why not `maxlength: 256,`?

app/src/pages/admin/allUsersPage/usersToolbar/usersEntities.jsx
twinfreaks: I did it similar to other fields (line 219)

app/src/pages/admin/allUsersPage/usersToolbar/usersEntities.jsx
twinfreaks: I did it similar to other fields (line 219)

app/src/components/inputs/inputConditional/inputConditional.jsx
twinfreaks: maxLength is a generic prop that was set to these input fields. inputprops.maxlength is a custom property for these kind of field

DiscoElevator: Still don't see any difference - I see 2 identical props with different names.

app/src/pages/admin/allUsersPage/usersToolbar/usersEntities.jsx
DiscoElevator: That input has different interface because of the original input is incapsulated by the underlying library.

PR #1871, opened 2019-08-02
EPMRPP-39883 - React: User is not redirected to login page in case of server sends the 401 code on request

Andrew-Omelchenko:


app/src/controllers/user/sagas.js
```diff
@@ -105,6 +107,21 @@ function* fetchUserWorker() {
     user = yield call(fetch, URLS.user());
     yield put(fetchUserSuccessAction(user));
   } catch (err) {
+    const error = err.message;
+    if (error === ERROR_UNAUTHORIZED) {
+      yield put(
+        showNotification({
```
DiscoElevator: 1. Message should not be shown when user starts on login page without redirecting on token expiration.
2. Redirecting on token expiration has been already implemented (see `initAuthInterceptor` in `common/utils/fetch`). Have you investigated the reason why it doesn't work in this case?

Andrew-Omelchenko: Fixed.

app/src/common/utils/fetch.js
```diff
@@ -48,7 +48,7 @@ export const initAuthInterceptor = (store) => {
     (error) => {
       if (error.response && error.response.status === 401) {
         const isAuthorized = isAuthorizedSelector(store.getState());
-        isAuthorized && store.dispatch(logoutAction());
+        !isAuthorized && store.dispatch(logoutAction());
```
AmsterGet: This fix was created to avoid unnecessary dispatch of  `logoutAction` for unauthorized user. It means that if the user is not already in the system, this action shouldn't be dispatched.

Andrew-Omelchenko: Fixed.

DiscoElevator:
Closed in favor of #1979

PR #1876, opened 2019-08-05
EPMRPP-39639 Fix Cumulative trend chart controls on widget wizard

AmsterGet:


app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/cumulativeTrendControls.jsx
```diff
@@ -124,6 +132,10 @@ export class CumulativeTrendControls extends Component {
         },
       },
     });
+
+    this.state = {
```
DiscoElevator: I think it's time to use [FieldArray](https://redux-form.com/7.3.0/docs/api/fieldarray.md/) instead of custom state.
There will be a new component which renders attribute array (inputs, remove buttons) and "add" button.

AmsterGet: Fixed

AmsterGet:
Fixed

PR #1879, opened 2019-08-06
EPMRPP-40034 - REACT. Launches. Incorrect response in sorting of launches, when user moves between pages

Andrew-Omelchenko:


app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -170,9 +182,18 @@ export class LaunchFiltersContainer extends Component {
         ...sortObject,
       }));
     }
-    this.props.fetchLaunchesWithParamsAction({
-      [SORTING_KEY]: sortingString,
-    });
+    this.props.updatePageProperties(
```
DiscoElevator: Sorting on launches is located inside of the active filter or the component's inner state in case of no filter. No need to put it into the url, since there's no sync between these places. Please investigate options of getting sorting params from the active filter or the state.

DiscoElevator:
Closed in favor of #1906

PR #1881, opened 2019-08-06
EPMRPP-37271. Full screen mode only according to the internal size of the tab.

ElenaRomanchuk:
image


app/src/common/img/cancel-inline.svg
```diff
@@ -0,0 +1,5 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 24 24" version="1.1" width="24px" height="24px">
```
DiscoElevator: there's already a cross icon: `app/src/common/img/cross-icon-inline.svg`

app/src/components/containers/fullscreen/fullscreen.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+  componentDidMount() {
+    document.addEventListener('keyup', this.exitFullScreen);
+  }
+
+  componentWillUnmount() {
+    document.removeEventListener('keyup', this.exitFullScreen);
+  }
+
+  detectExitKey = (event) => {
```
DiscoElevator: Can't see usage. Is there any?

ElenaRomanchuk:
I had a mess in class methods. Thanks. Fixed!

PR #1883, opened 2019-08-06
EPMRPP-40031 - React. Launches. All launches. Possibility to unclick a Filter tab

Andrew-Omelchenko:


app/src/pages/inside/common/launchFiltersToolbar/filterList/filterList.jsx
```diff
@@ -24,7 +25,7 @@ export const FilterList = ({
           active={filter.id === activeFilterId}
           share={filter.share}
           unsaved={unsavedFilterIds.indexOf(filter.id) > -1}
-          onClick={() => onSelectFilter(filter.id)}
+          onClick={() => onSelectFilter(activeFilterId === filter.id ? ALL : filter.id)}
```
DiscoElevator: What about the `Latest` mode? Should it be changed to `All` in this case?

Andrew-Omelchenko: From the `Expected Flow`
2. I click on a quick filter. Quick filter is gonna grey
3. I click on a quick filter one more time. The quick filter should be unclicked.
Do `Latest launches` belong to quick filters?

DiscoElevator: > 4. I click on a quick filter one more time. The quick filter should be unclicked. All launches are should be highlighted by grey

This dropdown has 2 options: All, Latest and the value is stored separately. It is unclear what should be done in case of the "Latest" mode. Please clarify with BA.

Andrew-Omelchenko: Ok.

Andrew-Omelchenko: Fixed. Switches back to` All` or `Latest` mode.

PR #1894, opened 2019-08-09
EPMRPP-39642 - React. Cumulative trend chart. Add breadcrumbs in case user input wrong attribute key for Level 2

KatsiarynaTatarynovich:


app/src/components/widgets/charts/cumulativeTrendChart/cumulativeTrendChart.jsx
```diff
@@ -215,7 +215,7 @@ export class CumulativeTrendChart extends PureComponent {
 
     return this.state && this.state.chartData ? (
       <div className={cx('cumulative-trend-chart', { 'preview-view': isPreview })}>
-        {isDetailsView ? (
+        {isDetailsView && chartData.labels.length ? (
```
AmsterGet: Please create an additional variable f.e. `isChartDataAvailable = !!chartData.labels.length` and use it where you need.

PR #1912, opened 2019-08-14
EPMRPP-39581: React: Attributes on Unique bugs table widget are displayed as links

KatsiarynaTatarynovich:


app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.scss
```diff
@@ -25,7 +25,6 @@
   font-size: 12px;
   color: $COLOR--gray-60;
   text-decoration: none;
-  cursor: pointer;
```
DiscoElevator: It will break cursor for clickable attributes on launches page

PR #1919, opened 2019-08-16
EPMRPP-40149: React: Js error on attempt to open attachment with .html extension on Edge browser

KatsiarynaTatarynovich:


app/src/controllers/log/attachments/sagas.js
```diff
@@ -109,9 +109,14 @@ function* openAttachment({ payload: { id, contentType } }) {
...
-    const url = URL.createObjectURL(data);
-    window.open(url);
-    URL.revokeObjectURL(url);
+
+    if (window.navigator && window.navigator.msSaveOrOpenBlob) {
+      window.navigator.msSaveOrOpenBlob(data);
+    } else {
+      const url = URL.createObjectURL(data);
+      window.open(url);
+      URL.revokeObjectURL(url);
```
DiscoElevator: Please move this outside of the condition. URL should be revoked in any case to prevent memory leaks

PR #1927, opened 2019-08-19
EPMRPP-40306: React: Click on attribute (with structure key:value) invokes adding two filter tabs instead of one

KatsiarynaTatarynovich:


app/src/components/filterEntities/containers/filterEntitiesContainer.jsx
```diff
@@ -83,12 +83,19 @@ export class FilterEntitiesContainer extends Component {
...
 
-  handleAdd = (entity) => {
+  handleAdd = (entities) => {
     this.setState(
-      (prevState) => ({
-        values: { ...prevState.values, [entity.id]: entity.value },
-        visibleFilters: [...prevState.visibleFilters, entity.id],
-      }),
+      (prevState) => {
+        const entitiesObj = entities
```
DiscoElevator: `const entitiesObj = entities.reduce((acc, entity) => ({...acc, [entity.id]: entity.value}), {});`

DiscoElevator: Also, this change breaks adding filters by clicking a grid header (see `FilterEntitiesAddHandlerContainer`). I suggest supporting both cases (array and single entity)

app/src/components/filterEntities/containers/filterEntitiesContainer.jsx
```diff
@@ -83,12 +83,22 @@ export class FilterEntitiesContainer extends Component {
...
 
-  handleAdd = (entity) => {
+  handleAdd = (entities) => {
     this.setState(
-      (prevState) => ({
-        values: { ...prevState.values, [entity.id]: entity.value },
-        visibleFilters: [...prevState.visibleFilters, entity.id],
-      }),
+      (prevState) => {
+        const entitiesObj = (Array.isArray(entities) &&
```
DiscoElevator: I think it's better to use `if..else` to make it more readable

app/src/components/filterEntities/containers/filterEntitiesContainer.jsx
```diff
@@ -83,12 +83,22 @@ export class FilterEntitiesContainer extends Component {
...
+          entities.reduce((acc, entity) => ({ ...acc, [entity.id]: entity.value }), {})) || {
+          [entities.id]: entities.value,
+        };
+        const entitiesId = (Array.isArray(entities) && entities.map((entity) => entity.id)) || [
```
DiscoElevator: Same here

PR #1931, opened 2019-08-20
EPMRPP-33468 Widgets refactoring

AmsterGet:
Also has been fixed:
- [EPMRPP-41485](https://jiraeu.epam.com/browse/EPMRPP-41485)
- [EPMRPP-40480](https://jiraeu.epam.com/browse/EPMRPP-40480)


app/src/controllers/project/sagas.js
```diff
@@ -255,12 +255,19 @@ function* watchUpdatePAState() {
...
-  if (!isAdminAccess) {
+  let project = {};
+  if (isAdminAccess) {
+    try {
+      project = yield call(fetch, URLS.project(projectId));
+    } catch (error) {
+      yield put(showDefaultErrorNotification(error));
+    }
+  } else {
+    project = yield call(fetch, URLS.project(projectId));
```
DiscoElevator: What's the difference with line 261?

app/src/pages/inside/dashboardItemPage/widgetsGrid/widget/widget.jsx
```diff
@@ -89,14 +88,15 @@ export class SimpleWidget extends Component {
...
-    const uncheckedLegendItems = [...this.state.uncheckedLegendItems];
+    const uncheckedItemIndex = this.uncheckedLegendItems.indexOf(itemId);
+    const uncheckedLegendItems = [...this.uncheckedLegendItems];
     if (uncheckedItemIndex !== -1) {
       uncheckedLegendItems.splice(uncheckedItemIndex, 1);
     } else {
       uncheckedLegendItems.push(itemId);
     }
-    this.setState({ uncheckedLegendItems }, callback);
+    this.uncheckedLegendItems = uncheckedLegendItems;
```
DiscoElevator: What's the point of moving dynamic data from state to a class instance?

app/src/components/widgets/singleLevelWidgets/charts/common/passingRateChart/config/utils.js
```diff
@@ -0,0 +1,59 @@
+import { CHART_MODES, MODES_VALUES } from 'common/constants/chartModes';
+
+export const getPercentage = (value, totalItems) => (value / totalItems * 100).toFixed(2);
+
+export const getChartVieModeOptions = (viewMode, isPreview, totalItems) =>
```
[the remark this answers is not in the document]
AmsterGet: Fixed

app/src/pages/inside/dashboardItemPage/widgetsGrid/widget/widget.jsx
AmsterGet: This improvement is made for performance reasons.
We don't need to re-render chart components after updating the legend (it is controlled by the chart itself) see for example [this](https://github.com/reportportal/service-ui/pull/1931/files#diff-823184e033f3a12a980cda5cd468dfe9R148-R151) line

app/src/controllers/project/sagas.js
AmsterGet: I used to think that would be great to catch error on get project only for admin access, but for now I think that it isn't bad idea to handle error on every request.
Fixed.

PR #1941, opened 2019-08-22
EPMRPP-40034/EPMRPP-40085: Keep sorting when changing pagination parameters on Lau…

DiscoElevator:
…nches page


DiscoElevator:
Also fixes EPMRPP-40085

PR #1946, opened 2019-08-23
EPMRPP-40112. Disable widget resizing in print preview.

ElenaRomanchuk:


app/src/pages/inside/dashboardItemPage/widgetsGrid/widgetsGrid.jsx
```diff
@@ -253,7 +253,7 @@ export class WidgetsGrid extends Component {
       <div
         className={cx('widgets-grid', {
           mobile: this.state.isMobile,
-          'full-screen': this.props.isFullscreen,
+          'full-screen': this.props.isFullscreen || this.props.isPrintMode,
```
AmsterGet: I think would be better to rename class `full-screen` to f.e. `non-resizable`, because now it is responsible not only for fullscreen mode.

ElenaRomanchuk: Done.

PR #1948, opened 2019-08-23
Fix link/unlink and post issues for group operations with to investigate items.

ElenaRomanchuk:


DiscoElevator:
Fixes reportportal/reportportal#705

PR #1950, opened 2019-08-26
EPMRPP-40301 Adjust services info to a new backend

AmsterGet:


app/src/common/urls.js
```diff
@@ -215,7 +215,6 @@ export const URLS = {
       ...createFilterQuery(filterEntities),
     })}`,
 
-  apiInfo: () => `${urlBase}info`,
   compositeInfo: () => `${compositeBase}info`,
```
ElenaRomanchuk: I suggest to rename `compositeInfo` for example to `appInfo`, it will look more intuitive.

AmsterGet: Fixed

PR #1956, opened 2019-08-27
EPMRPP-40800. Permissions for dashboards.

ElenaRomanchuk:


ElenaRomanchuk:
I've just comitted EPMRPP-40803 task to this PR with permissions for adding widgets in case of empty dashboard.

PR #1975, opened 2019-09-02
EPMRPP-39587. Re-test React. Dashboard. Widget. Overall statistics. I…

tritseratops:
…ncorrect response of the components (Donut view). Fixed. Added function to identify if small view should be used. Updated small view title font to fit in center of donut. Updated resize chart function with forceUpdate and small logic fix. Force update usage is discouraged, but alternative is to move width and height to state and update state on resize. I did not do it , because I in that case change of width and height will affect performance in other places in addition to resize.

### DIFF
--- a/app/package-lock.json
+++ b/app/package-lock.json
@@ -5,34 +5,34 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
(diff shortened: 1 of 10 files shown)


tritseratops:
Sure, let me change it.

tritseratops:
Unfortunately I can't delete file from this pull request without adding additional commit, so I created new pull request.
https://github.com/reportportal/service-ui/pull/2039

PR #1980, opened 2019-09-05
EPMRPP-40088. Fix display of linked ticket with long ID

ElenaRomanchuk:
image


ElenaRomanchuk:
> Shouldn't this be displayed as in version 4? (see the screenshot below)
> image

I discussed this with BA and she decided to truncate the value using the ellipsis.

PR #1982, opened 2019-09-05
EPMRPP-39299 Fix login page elements everflow

AmsterGet:
image

### DIFF
--- a/Dockerfile-full
+++ b/Dockerfile-full
@@ -12,7 +12,7 @@ WORKDIR /usr/src/app
 COPY ./src/main/resources/public/package.json /usr/src/app/
 RUN npm install
 COPY ./src/main/resources/public/ /usr/src/app/
-RUN npm run test && npm run build
+RUN npm run build && npm run test
 
 FROM alpine:latest
 RUN apk --no-cache add ca-certificates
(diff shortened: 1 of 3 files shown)


AmsterGet:
The issue with fail builds was associated with compiled files usage for tests purposes.

Dockerfile-full
```diff
@@ -12,7 +12,7 @@ WORKDIR /usr/src/app
 COPY ./src/main/resources/public/package.json /usr/src/app/
 RUN npm install
 COPY ./src/main/resources/public/ /usr/src/app/
-RUN npm run test && npm run build
+RUN npm run build && npm run test
```
DiscoElevator: Doesn't look like a relevant change.

Makefile
```diff
@@ -50,8 +50,8 @@ build-server: checkstyle test
 # Builds the project
 build-statics:
 	npm --prefix $(UI_BUILD_ROOT) install
-	npm --prefix $(UI_BUILD_ROOT) run test
```
DiscoElevator: Doesn't look like a relevant change.

Dockerfile-full
AmsterGet: Please see the comment https://github.com/reportportal/service-ui/pull/1982#issuecomment-530579950

PR #1992, opened 2019-09-09
EPMRPP-39162: disable autocomplete in Link Issue modal for Rally BTS

DiscoElevator:


app/src/pages/inside/stepPage/modals/linkIssueModal/linkIssueFields/linkIssueFields.jsx
```diff
@@ -79,7 +83,7 @@ export class LinkIssueFields extends Component {
               name={`${issue}.issueLink`}
               fieldWrapperClassName={cx('field-wrapper')}
               label={this.props.intl.formatMessage(messages.issueLinkLabel)}
-              parse={this.parseValue}
+              onChange={this.handleIssueLinkChange}
```
AmsterGet: May be better: `onChange={withAutocomplete ? this.handleIssueLinkChange: null}`?

DiscoElevator: There's almost no difference in this case, but with handler renaming it will look more clear, I think. Fixed.

PR #2000, opened 2019-09-11
EPMRPP-40114: Fix disappearing status dropdown on log page

DiscoElevator:


app/src/controllers/log/selectors.js
```diff
@@ -255,10 +256,16 @@ export const logViewModeSelector = (state) => {
...
-  const filteredItems = items.filter((item) => 'hasContent' in item);
-  return filteredItems.length === 0;
-});
+export const isLogPageWithOutNestedSteps = createSelector(
+  logItemsSelector,
+  pagePropertiesSelector,
+  (items, pageQuery) => {
+    const query = extractNamespacedQuery(pageQuery, NAMESPACE);
+    const logStatus = query[LOG_STATUS_FILTER_KEY];
+    const filteredItems = items.filter((item) => 'hasContent' in item);
```
ElenaRomanchuk: `some` or `every` method would be better for this purpose.

PR #2006, opened 2019-09-13
EPMRPP-42015. Rename UUID to access token.

ElenaRomanchuk:
image


AmsterGet:
What about component, function names and other places where `uuid` still exists?

ElenaRomanchuk:
I renamed all modules and elements, which was related to uuid user, for except api-related blocks.

PR #2007, opened 2019-09-16
EPMRPP-41169: Disable attribute submit in case of duplicate.

DiscoElevator:
Also remove redux-form from attribute editor.

### DIFF
--- a/app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
+++ b/app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
@@ -85,21 +85,21 @@ export class AttributeEditor extends Component {
   parseValue = (value) => (value ? value.value : undefined);
 
   handleKeyChange = (key) => {
-    this.setState({
+    this.setState((oldState) => ({
       key: this.parseValue(key),
-      errors: this.getValidationErrors(this.parseValue(key), this.state.value),
-    });
+      errors: this.getValidationErrors(this.parseValue(key), oldState.value),
+    }));
   };
 
   handleValueChange = (value) => {
-    this.setState({
+    this.setState((oldState) => ({
       value: this.parseValue(value),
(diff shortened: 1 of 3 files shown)


app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
```diff
@@ -94,52 +84,86 @@ export class AttributeEditor extends Component {
   formatValue = (value) => (value ? { value, label: value } : null);
   parseValue = (value) => (value ? value.value : undefined);
 
+  handleKeyChange = (key) => {
+    this.setState({
+      key: this.parseValue(key),
+      errors: this.getValidationErrors(this.parseValue(key), this.state.value),
```
ElenaRomanchuk: You use state variables for calculating the new state. Due to the fact that React may batch setState calls, it would be better to use a function instead of an object in setState call.

app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
```diff
@@ -94,52 +84,86 @@ export class AttributeEditor extends Component {
...
+    });
+  };
+
+  handleValueChange = (value) => {
+    this.setState({
+      value: this.parseValue(value),
+      errors: this.getValidationErrors(this.state.key, this.parseValue(value)),
```
ElenaRomanchuk: The same

app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
```diff
@@ -94,52 +84,86 @@ export class AttributeEditor extends Component {
...
+    });
+  };
+
+  isAttributeUnique = () =>
+    !this.props.attributes.find(
```
ElenaRomanchuk: `!this.props.attributes.some`

PR #2010, opened 2019-09-16
EPMRPP-41153. Fix stopped launch on item log view.

ElenaRomanchuk:


ElenaRomanchuk:
> What about `STEP view` as described in the story? (see the screenshot below)
> 
> image
It was fixed earlier.

PR #2013, opened 2019-09-17
EPMRPP-41996: Fix displaying of filters for removed subtypes

DiscoElevator:


app/src/pages/inside/common/filterEntitiesGroups/launchLevelEntities.jsx
```diff
@@ -439,13 +440,37 @@ export class LaunchLevelEntities extends Component {
     });
     return defectTypeEntities;
   };
+
+  collectLostEntities = (entities) => {
+    const lostKeys = Object.keys(this.props.filterValues).filter(
+      (key) => !entities.find((entity) => entity.id === key),
```
ElenaRomanchuk: I think we should to investigate which set of array methods would be better here: `find` or `some`, `filter` and `map` or `reduce`.

DiscoElevator: Decided not to change, as it looks more readable

PR #2016, opened 2019-09-18
EPMRPP-40022-Upload-plugin-initialization-parameters

AmsterGet:
image

Uploading a single file for the `Upload plugin modal` was also done as part of this task.


app/src/pages/admin/pluginsPage/pluginsToolbar/actionPanel/uploadCustomBlock/uploadCustomBlock.jsx
```diff
@@ -0,0 +1,50 @@
...
+
+const cx = classNames.bind(styles);
+const messages = defineMessages({
+  initializationParamsLabel: {
+    id: 'UploadCustomBlock.initializationParamsLabel',
+    defaultMessage: 'Initialization parameters',
+  },
+});
+
+@reduxForm({
```
DiscoElevator: I think it's an overkill here. You can pass onChange handler from ImportModal to the custom block and store necessary state there (something like `customState`). Not the best solution in general, but that's all we can do in this case without refactoring.

AmsterGet: Fixed

PR #2019, opened 2019-09-19
EPMRPP-39182: Add filters to debug page

DiscoElevator:


DiscoElevator:
Does not work with pagination. Needs rework

PR #2027, opened 2019-09-23
Fix login url so it won't show on the url line

Avielyo10:


Avielyo10:
This should be a fix for this issue https://github.com/reportportal/service-authorization/issues/86

AmsterGet:
For the `4`th version it's also already fixed [here](https://github.com/reportportal/service-ui/pull/1814)

Avielyo10:
As you can see [here](https://github.com/reportportal/service-authorization/issues/86#issuecomment-531932745) that's not fixed.

AmsterGet:
Already fixed in the [`v4`](https://github.com/reportportal/service-ui/tree/v4) branch, but it is still not released.
A release for `v4` will be available when QA team finish the regression tests for` v5-beta`.

Avielyo10:
@AmsterGet Thanks that's great!
Do you have any time estimation?

englandprevails:
There is no public release with this security flaw fixed.  Why is this issue closed?

PR #2043, opened 2019-09-26
EPMRPP-40716. Message for empty search results in dashboards.

ElenaRomanchuk:
image


app/src/pages/inside/dashboardPage/dashboardList/EmptyDashboards/emptyDashboards.jsx
```diff
@@ -32,15 +40,31 @@ export class EmptyDashboards extends Component {
...
+    filter: PropTypes.string,
   };
 
   static defaultProps = {
     action: () => {},
     userDashboards: false,
+    filter: '',
+  };
+
+  renderNoFilteredItems = () => {
```
AmsterGet: Can you reuse `pages/inside/common/noResultsForFilter` component here?

ElenaRomanchuk: Ok. Sure. I never noticed it before.

PR #2045, opened 2019-09-26
EPMRPP-42212: Fix rendering of items with status "cancelled"

DiscoElevator:


DiscoElevator:
> There is no corresponding block for canceled item on history line.
> image

Decided to leave it as is. New ticket will be created if necessary.

PR #2056, opened 2019-09-27
EPMRPP-41164: React: In case no any BTS plugin installed, the message in tool-tip should notify user about it

KatsiarynaTatarynovich:


app/src/pages/inside/common/suiteTestToolbar/actionPanel/actionPanel.jsx
```diff
@@ -71,11 +71,11 @@ const messages = defineMessages({
   },
   noBugTrackingSystemToLinkIssue: {
     id: 'ActionPanel.noBugTrackingSystemToLinkIssue',
-    defaultMessage: 'Configure bug tracking system to link issue',
+    defaultMessage: 'BTS integration is not configured. Connect with Admin',
```
DiscoElevator: It repeats 4 times. I think it's better be moved to `common/constants/localization.js`

AmsterGet:
Closed if favor of https://github.com/reportportal/service-ui/pull/2070

PR #2061, opened 2019-09-27
EPMRPP-41493 Launches duration chart. Add interrupted status

AmsterGet:


DiscoElevator:
http://localhost:3000/#mtumas_personal/dashboard/780
There are few interrupted launches with duration > 0, but on the widget I see no bars:
изображение
изображение

Is it correct behavior?

AmsterGet:
Fixed. The issue was related to the average launches duration calculation.

PR #2074, opened 2019-10-01
EPMRPP-39093. Fix filtering options disappearing in case of invalid value is entered

ElenaRomanchuk:


app/src/pages/inside/common/filterEntitiesGroups/launchLevelEntities.jsx
```diff
@@ -336,7 +338,7 @@ export class LaunchLevelEntities extends Component {
         }),
         validationFunc: commonValidators.launchNumericEntity,
         title: intl.formatMessage(messages.SkippedTitle),
-        active: STATS_SKIPPED in filterValues,
+        active: visibleFilters.includes(STATS_SKIPPED),
```
DiscoElevator: Please also add `visibleFilters` check to defect type statistics(`getDynamicEntities`)

PR #2090, opened 2019-10-07
EPMRPP-39462: Localization fixes

DiscoElevator:
Includes fixes for EPMRPP-36495, EPMRPP-36493, EPMRPP-36574, EPMRPP-36494


app/src/pages/common/settingsPage/generalTab/generalTab.scss
```diff
@@ -27,21 +27,21 @@
...
-    margin-left: 345px;
+    margin-left: 320px;
   }
   @media (max-width: $SCREEN_XS_MAX) {
     display: none;
   }
 }
 .submit-button {
-  width: 70px;
+  width: 95px;
```
AmsterGet: I think it will be better to avoid dependency between translated text and the component's static size.
It should be adjusted by content.

app/src/pages/common/settingsPage/demoDataTab/demoDataTab.scss
```diff
@@ -80,7 +80,7 @@
 
 .generate-button {
   display: block;
-  width: 155px;
```
AmsterGet: Same here

PR #2093, opened 2019-10-07
EPMRPP-41637 - React: Login and Password fields should be highlighted…

mykhailo-kalashnikov-epam:
… in case of bad credentials provided

### DIFF
--- a/app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.jsx
+++ b/app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.jsx
@@ -23,13 +23,13 @@ import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
 import Parser from 'html-react-parser';
 import { connect } from 'react-redux';
-import { reduxForm } from 'redux-form';
+import { reduxForm, stopSubmit } from 'redux-form';
 import { FormattedMessage, injectIntl, intlShape, defineMessages } from 'react-intl';
 import Link from 'redux-first-router-link';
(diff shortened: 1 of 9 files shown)


app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.jsx
```diff
@@ -101,6 +109,13 @@ export class LoginForm extends React.Component {
     this.state = this.calculateLoginLimitState();
   }
 
+  componentWillReceiveProps(nextProps) {
```
AmsterGet: React no longer recommend using this method see [this](https://ru.reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops) link.
What do you think about `getDerivedStateFromProps` or `componentDidUpdate` lifecycle methods?

app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.jsx
```diff
@@ -105,6 +113,10 @@ export class LoginForm extends React.Component {
     if (prevProps.lastFailedLoginTime !== this.props.lastFailedLoginTime) {
       this.blockLoginForm();
     }
+    const { badCredentials } = this.props;
+    if (badCredentials) {
```
DiscoElevator: I think it needs a better condition check because now it fires when I type values in different fields, so I see validation errors even when I haven't submitted new values.
Maybe `if (badCredentials  !== prevProps.badCredentials && badCredentials) {`

PR #2125, opened 2019-10-18
EPMRPP-42616: Fix filter duplication after launches page refresh

DiscoElevator:


app/src/controllers/filter/reducer.test.js
```diff
@@ -62,4 +62,19 @@ describe('filter reducer', () => {
...
+      const oldState = true;
+      expect(
+        launchesFiltersReadyReducer(oldState, {
+          type: 'unknownAction',
+        }),
+      ).toBe(oldState);
+    });
+
+    test('should return true on FETCH_USER_FILTERS_SUCCESS action', () => {
+      expect(launchesFiltersReadyReducer(false, FETCH_USER_FILTERS_SUCCESS).toBe(true));
```
AmsterGet: `expect(launchesFiltersReadyReducer(false, FETCH_USER_FILTERS_SUCCESS)).toBe(true);`

PR #2128, opened 2019-10-18
EPMRPP-42510. Google analytics coverage. Logs page.

ElenaRomanchuk:


app/src/pages/inside/logsPage/sauceLabsSection/videoSection/videoPlayer/videoPlayer.jsx
```diff
@@ -23,6 +25,7 @@ export class VideoPlayer extends React.Component {
 
   componentDidMount() {
     this.player = videojs(this.videoNode.current, this.props);
+    this.player.on('play', () => this.props.onPlay());
```
AmsterGet: Why not just `this.player.on('play', this.props.onPlay);`?

ElenaRomanchuk: You are right, closure is unnecessary here.

PR #2131, opened 2019-10-21
EPMRPP-42903 Use Launches Comparison chart for Compare launches modal

AmsterGet:


app/src/components/widgets/singleLevelWidgets/charts/overallStatistics/index.js
```diff
@@ -0,0 +1 @@
+export { OverallStatistics } from './overallStatistics';
```
DiscoElevator: License is missing

AmsterGet: Fixed

PR #2152, opened 2019-10-29
EPMRPP-43363, EPMRPP-43364: Make PM be able to edit and delete shared widgets and dashboards

DiscoElevator:


DiscoElevator:
Also, a part of EPMRPP-43371 was done by accident (merge "edit" and "delete" item rows)

PR #2159, opened 2019-10-30
EPMRPP-39640. Custom tooltip for launch and test item name.

ElenaRomanchuk:


app/src/pages/inside/common/itemInfo/itemInfo.jsx
```diff
@@ -109,6 +111,17 @@ export class ItemInfo extends Component {
     }
   };
 
+  renderItemName = () => {
+    const { value } = this.props;
+    const Tooltip = () => <TextTooltip tooltipContent={value.name} />;
+    const ItemNameComponent = () => <span>{`${formatItemName(value.name)} `}</span>;
+    const ItemNameWithTooltip = withTooltip({
```
DiscoElevator: It's better to make it a dedicated component:
```
const ItemNameTooltip = withTooltip({
  TooltipComponent: TextTooltip,
  data: { dynamicWidth: true },
})(({ children }) => children);
```

And use it below:
```
<ItemNameTooltip tooltipContent={value.name}>
  <span>{formatItemName(value.name)}</span>
</ItemNameTooltip>
```

PR #2160, opened 2019-10-30
EPMRPP-43395 Hide extra popups on logout

AmsterGet:


app/src/pages/inside/launchesPage/modals/launchMergeModal/launchMergeModal.jsx
```diff
@@ -214,11 +217,15 @@ export class LaunchMergeModal extends Component {
...
-      closeModal();
-      this.props.hideScreenLockAction();
-    });
+    })
+      .then(() => {
+        this.props.data.fetchFunc();
+        closeModal();
+      })
+      .catch(this.props.showDefaultErrorNotification)
+      .finally(() => {
```
DiscoElevator: Doesn't work in IE: [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally)

AmsterGet: Fixed

PR #2166, opened 2019-10-31
EPMRPP-43417: React. Step level. Error on updating Passed item with ' Ignore in Auto Analysis' option

KatsiarynaTatarynovich:


app/src/controllers/step/actionValidators.js
```diff
@@ -15,7 +15,7 @@
  */
 
 export const validateIgnoreInAA = (item) => {
-  if (item.issue && !item.issue.issueType) {
+  if ((item.issue && !item.issue.issueType) || !item.issue) {
```
AmsterGet: It can be simplified: `!item.issue || !item.issue.issueType`

app/src/controllers/step/actionValidators.js
```diff
@@ -25,7 +25,7 @@ export const validateIgnoreInAA = (item) => {
 };
 
 export const validateIncludeInAA = (item) => {
-  if (item.issue && !item.issue.issueType) {
+  if ((item.issue && !item.issue.issueType) || !item.issue) {
```
AmsterGet: Same here

app/src/controllers/step/actionValidators.js
KatsiarynaTatarynovich: Fixed

PR #2174, opened 2019-11-05
EPMRPP-42479. Allow member and customer to finish not own launch.

ElenaRomanchuk:


app/src/common/constants/permissions.js
```diff
@@ -91,7 +91,7 @@ export const PERMISSIONS_MAP = {
     [ACTIONS.EDIT_OWN_ACCOUNT]: ALL,
     [ACTIONS.DELETE_LAUNCH]: OWNER,
     [ACTIONS.EDIT_LAUNCH]: OWNER,
-    [ACTIONS.FORCE_FINISH_LAUNCH]: OWNER,
```
AmsterGet: Force finish action should be available for not own launches only with `rerun` label

app/src/common/constants/permissions.js
```diff
@@ -128,7 +128,7 @@ export const PERMISSIONS_MAP = {
...
```
AmsterGet: Same here

PR #2178, opened 2019-11-05
EPMRPP-43276 Refine filters. Save filter conditions after update

AmsterGet:


app/src/components/filterEntities/containers/filterEntitiesContainer.jsx
```diff
@@ -51,12 +65,17 @@ export class FilterEntitiesContainer extends Component {
...
+      const newEntities = saveEntitiesConditions(
+        props.entities,
+        state.prevEntities,
+        state.visibleFilters,
+      );
       return {
         errors: {},
-        values: props.entities,
-        prevEntities: props.entities,
-        visibleFilters: Object.keys(props.entities),
```
ElenaRomanchuk: It breaks set up values for existing filter on navigating via direct link.

AmsterGet:
Not relevant any more

PR #2186, opened 2019-11-07
Fix login username and password double encoded

waynesun09:
With data as object and add contentType, the password was double
encoded and login will fail with special charcater like '@'.

To fix the username and password value, decode it before pass to
ajax.

Signed-off-by: Wayne Sun <gsun@redhat.com>

### DIFF
--- a/src/main/resources/public/js/src/core/core-service.js
+++ b/src/main/resources/public/js/src/core/core-service.js
@@ -39,6 +39,8 @@ define([
         return call('DELETE', urls.userLogout());
     };
     var userLogin = function (data) {
+        data['username'] = decodeURIComponent(data.username);
+        data['password'] = decodeURIComponent(data.password);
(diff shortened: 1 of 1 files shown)


AmsterGet:
We have restrictions for the login field (Latin, numeric characters, hyphen, underscore, period), so the `@` character is not allowed.

Сan it be delayed?
At the moment, all our forces are focused on version 5.

waynesun09:
@AmsterGet For the v4, I tested with '@' and it works, this happened after the fix
https://github.com/reportportal/service-ui/commit/fd0cf0542e0ad736c77859fc13f7a13e6b48460e

So current official release is not affected.
For authorization with LDAP, strong password is preferred and special characters should not be excluded. 

As it not affect the release version, I think it could be delayed until next v4 release.

AmsterGet:
Thanks for the quick response.

Added in the scope of version 4.4.

AmsterGet:
Closed as v4 is out of support.

PR #2192, opened 2019-11-12
EPMRPP-43903: React. Overall statistics. PUT request is sent after first clicking on the clickable area

KatsiarynaTatarynovich:


app/src/controllers/testItem/sagas.js
```diff
@@ -153,7 +153,7 @@ function* fetchTestItems({ payload = {} }) {
     const filter = yield call(fetch, URLS.filter(project, filterId));
 
     if (filter) {
-      yield put(updateFilterAction(filter));
+      yield put(setFilterAction(filter));
```
AmsterGet: I think we can just use `showFilterOnLaunchesAction` here.

ElenaRomanchuk: Agree. Especially if we decided not to remove PUT-request on navigating from widget. For now we have an issue with access rights.

AmsterGet: There is no need to update an existing filter when a user just navigates to a page.

PR #2193, opened 2019-11-12
EPMRPP-43904: React. Error appeared after login to RP if user was logged out from RP on 'Settings' page

KatsiarynaTatarynovich:


app/src/pages/common/settingsPage/defectTypesTab/defectTypesGroup.jsx
```diff
@@ -99,15 +99,16 @@ export class DefectTypesGroup extends Component {
...
-        {group.map((subType, i) => (
-          <DefectSubType
-            key={subType.id}
-            data={subType}
-            parentType={group[0]}
-            group={i === 0 ? group : null}
-            isPossibleUpdateSettings={this.isPossibleUpdateSettings}
-          />
-        ))}
+        {group &&
```
ElenaRomanchuk: `group` is marked as required in `propTypes`. And there is no valid render for empty `group` here. So, I think, this condition is redundant.

PR #2195, opened 2019-11-13
v5.1-clickable-area-fix

KatsiarynaTatarynovich:


AmsterGet:
This changes should be reverted after the demo.

PR #2208, opened 2019-11-19
EPMRPP-44711 History. Change history items mapping logic on logs page

AmsterGet:
Waiting for https://github.com/reportportal/service-ui/pull/2210

### DIFF
--- a/app/src/controllers/log/selectors.js
+++ b/app/src/controllers/log/selectors.js
@@ -79,7 +79,7 @@ export const querySelector = createQueryParametersSelector({
 
 export const historyItemsSelector = createSelector(historyEntriesSelector, (entriesFromState) => {
   if (!entriesFromState.length) return [];
-  const historyItems = [...entriesFromState[0].resources].map(normalizeHistoryItem);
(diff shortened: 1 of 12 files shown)


app/src/controllers/log/selectors.js
```diff
@@ -77,37 +77,12 @@ export const querySelector = createQueryParametersSelector({
...
-      return {
-        ...normalizeHistoryItem(historyItem, filteredSameHistoryItems),
-      };
-    });
-
-    currentLaunchItem.launchNumber = currentLaunch.launchNumber;
-    historyItems.push(currentLaunchItem);
+export const historyItemsSelector = createSelector(historyEntriesSelector, (entriesFromState) => {
+  if (!entriesFromState.length) return [];
+  const historyItems = [...entriesFromState[0].resources].map(normalizeHistoryItem);
```
DiscoElevator: No need to create new array in this case:
` const historyItems = entriesFromState[0].resources.map(normalizeHistoryItem);`

AmsterGet: Yes, you're right! Fixed.

PR #2211, opened 2019-11-21
EPMRPP-38631:  Webpack update

DiscoElevator:
1. Storybook builds, but throws an exception due to circular dependencies. Since it was broken before this PR, it will be fixed in scope of an other task.
2. Due to updated prettier, some files need to be reformatted - it will be done in other PR.
3. Some ESLint rules were turned off - will be reviewed in other PR.
4. Commented `CircularDependencyPlugin` is left intentionally - it can be turned on locally to look for circular dependencies.


app/webpack/dev.config.js
```diff
@@ -0,0 +1,80 @@
...
+                resources: path.resolve(__dirname, '../src/common/css/variables/**/*.scss'),
+              },
+            },
+          ],
+        },
+      ],
+    },
+    plugins: [
+      new webpack.HotModuleReplacementPlugin(),
+      new CleanWebpackPlugin({
```
AmsterGet: Can we move this plugin definition to the `base.config.js` ?
As I see it's absolutely the same with [this](https://github.com/reportportal/service-ui/pull/2211/files#diff-7c9b12d67fa5d63ba0049392cbed6b9cR64-R66).

app/webpack/base.config.js
```diff
@@ -0,0 +1,88 @@
...
+        loader: 'svg-inline-loader',
+      },
+    ],
+  },
+  plugins: [
+    new WebpackNotifierPlugin({ skipFirstNotification: true }),
+    new HtmlWebpackPlugin({
+      template: path.resolve(__dirname, '../src/index.tpl.html'),
+      filename: 'index.html',
+    }),
```
AmsterGet: Favicon definition missed here.
`favicon: path.resolve(__dirname, '../src/common/img/favicon.ico'),`

PR #2215, opened 2019-11-22
EPMRPP-44650. Display full launch name on widget tooltip

ElenaRomanchuk:


ElenaRomanchuk:
Fixes reportportal/reportportal#740

PR #2220, opened 2019-11-25
EPMRPP-44683. Remove requests on get history and activity on log page…

ElenaRomanchuk:
… in debug mode.


DiscoElevator:
Fixes reportportal/reportportal#747

PR #2245, opened 2019-12-03
EPMRPP-43830: REACT.Widgets. Project Activity panel. Incorrect illustration of the changes

KatsiarynaTatarynovich:


app/src/components/widgets/singleLevelWidgets/tables/projectActivity/activities/notifications.jsx
```diff
@@ -35,18 +30,10 @@ const messages = defineMessages({
...
-    id: 'Notifications.onEmail',
-    defaultMessage: 'configured',
-  },
-  [SWITCH_OFF_NOTIFICATIONS]: {
-    id: 'Notifications.offEmail',
-    defaultMessage: 'switched off',
-  },
 });
 
 @injectIntl
```
ElenaRomanchuk: What do you think about rewriting this component as a pure function?

DiscoElevator: I like the idea, but it's not relevant for the scope of this PR and also it's not necessary

PR #2247, opened 2019-12-04
EPMRPP-44552: React. Widgets. 'Switches' column in 'Flaky test cases table'. Change the sequence of displayed switches

KatsiarynaTatarynovich:


app/src/components/widgets/singleLevelWidgets/tables/flakyTests/flakyTests.jsx
```diff
@@ -19,8 +19,22 @@ import PropTypes from 'prop-types';
...
 import * as cfg from './flakyTestsCfg';
 
+const prepareWidgetData = ({ flaky }) =>
+  flaky.map((item) => ({
+    flakyCount: item.flakyCount,
+    itemName: item.itemName,
+    total: item.total,
+    uniqueId: item.uniqueId,
+    startTime: item.startTime,
+    statuses: item.statuses.reverse(),
```
ElenaRomanchuk: It seems to me, it would be better to use spread operator here: `{...item, statuses: item.statuses.revese()}`

app/src/components/widgets/singleLevelWidgets/tables/flakyTests/flakyTests.jsx
```diff
@@ -19,8 +19,15 @@ import PropTypes from 'prop-types';
 import { TestsTableWidget } from '../components/testsTableWidget';
 import * as cfg from './flakyTestsCfg';
 
+const prepareWidgetData = ({ flaky }) =>
+  flaky.map((item) => ({ ...item, statuses: item.statuses.reverse() }));
```
DiscoElevator: `reverse` mutates the array, so please create a new one to prevent bugs:
`[...item.statuses].reverse()`

PR #2248, opened 2019-12-04
[DNM] Fix empty log view: fix historyItemsSelector.

ElenaRomanchuk:


ElenaRomanchuk:
Will be fixed in backend-side.

PR #2311, opened 2020-01-28
EPMRPP-45322: Launches. Log view. Remove !!!MARKDOWN MODE!!! from logs and automatically switch on Markdown Mode

KatsiarynaTatarynovich:


app/src/pages/inside/logsPage/logsGrid/logMessageBlock/logMessageBlock.jsx
```diff
@@ -37,14 +38,16 @@ export class LogMessageBlock extends Component {
 
   render() {
     const { value, refFunction, customProps } = this.props;
+    const markdownModeRegExp = new RegExp(MARKDOWN_MODE, 'g');
```
DiscoElevator: I think it would be enough to seek for this substring only at the beginning of the string (as it was before).

PR #2320, opened 2020-02-19
initial commit for Ukrainian language support

olgmaks:


app/src/pages/inside/profilePage/localizationBlock/localizationBlock.jsx
```diff
@@ -121,6 +132,8 @@ export class LocalizationBlock extends Component {
   };
 
   onChangeLanguage = (lang) => {
+    console.log('change lang');
```
DiscoElevator: Looks like you've forgot to remove `console.log`

olgmaks: fixed

PR #2323, opened 2020-02-27
EPMRPP-48347 Clear user credential fields

AmsterGet:


app/src/controllers/plugins/utils.js
```diff
@@ -50,6 +50,13 @@ export const isPostIssueActionAvailable = (integrations) =>
...
   );
 
-// TODO: remove check for AUTHORIZATION_GROUP_TYPE when designs and backend implementation will exist
 export const filterAvailablePlugins = (plugins = []) =>
   plugins.filter((item) => item.enabled && item.groupType !== AUTHORIZATION_GROUP_TYPE);
+
+export const clearSecretFields = (data, secretFields = []) =>
+  secretFields.reduce((acc, secretField) => {
+    const params = data.integrationParameters;
+    delete params[secretField];
```
DiscoElevator: Mutating arguments is a bad practice.

app/src/controllers/plugins/utils.js
```diff
@@ -50,6 +50,13 @@ export const isPostIssueActionAvailable = (integrations) =>
       item.integrationParameters.defectFormFields.length,
   );
 
-// TODO: remove check for AUTHORIZATION_GROUP_TYPE when designs and backend implementation will exist
 export const filterAvailablePlugins = (plugins = []) =>
   plugins.filter((item) => item.enabled && item.groupType !== AUTHORIZATION_GROUP_TYPE);
+
+export const clearSecretFields = (data, secretFields = []) =>
```
DiscoElevator: Looks like we already have such util function: `common/utils/omit`

AmsterGet: Oh! Thanks! Duplicate deleted.

PR #2334, opened 2020-03-11
EPMRPP-48786 Hide availability switcher for auth plugins

AmsterGet:


app/src/common/constants/pluginNames.js
```diff
@@ -29,14 +29,14 @@ export const SAML = 'saml';
...
-export const INTEGRATION_NAMES_BY_GROUP_TYPES_MAP = {
+export const PLUGIN_NAMES_BY_GROUP_TYPES_MAP = {
   [BTS_GROUP_TYPE]: [JIRA, RALLY],
   [NOTIFICATION_GROUP_TYPE]: [EMAIL],
   [OTHER_GROUP_TYPE]: [SAUCE_LABS],
   [AUTHORIZATION_GROUP_TYPE]: [SAML, LDAP, AD],
 };
 
-export const GROUP_TYPES_BY_INTEGRATION_NAMES_MAP = {
+export const GROUP_TYPES_BY_PLUGIN_NAMES_MAP = {
```
ElenaRomanchuk: Is this map always reversed copy of `PLUGIN_NAMES_BY_GROUP_TYPES_MAP`? Why not to use just GROUP_TYPES_BY_PLUGIN_NAMES_MAP (and use filter method to get plugins by group type?)

AmsterGet: `PLUGIN_NAMES_BY_GROUP_TYPES_MAP` is frequantly used in `selectors`.
For the updated `isAuthorizationPlugin` util function approach with `GROUP_TYPES_BY_PLUGIN_NAMES_MAP` will work faster.
So I decided to leave two of them.

PR #2337, opened 2020-03-13
EPMRPP-48790: All launches view. A user cannot open a tab in a new window

KatsiarynaTatarynovich:


app/src/pages/inside/common/launchFiltersToolbar/filterList/filterItem/filterItem.jsx
```diff
@@ -31,19 +33,42 @@ const stopPropagation = (func) => (e) => {
   func(e);
 };
 
+const handleClick = (e) => {
+  e.preventDefault();
+};
+
+const getLaunchesPageLink = (filter, projectId, allLatest, active) => {
```
ElenaRomanchuk: What about to transform this function to selector? Also I think it would be better to connect it directly to FilterItemBase and don't pass redundant props through whole hierarchy.

KatsiarynaTatarynovich: @AmsterGet What do you think about transformation to the selector?

AmsterGet: @KatsiarynaTatarynovich please see the comment https://github.com/reportportal/service-ui/pull/2337#discussion_r393594676

app/src/controllers/filter/selectors.js
```diff
@@ -68,3 +74,19 @@ export const dirtyFilterIdsSelector = createSelector(
       })
       .map((item) => item.id),
 );
+
+export const getLaunchesPageLinkSelector = createSelector(
```
AmsterGet: May be better `getLaunchFilterLinkSelector`

ElenaRomanchuk: I suggest to rename it to `createLaunchesFilterLinkSelector`. May be it would be better to move it to launches controller?

PR #2338, opened 2020-03-13
EPMRPP-48788 Log level. Fix logs with attachments parameter behavior

AmsterGet:


AmsterGet:
We've been decided to remove localStorage syncronization in version 5.1 for logs with attachments, hide passed logs and hide empty steps.
It can be solved as a part of another task in the scope of version 5.2 (see the comments for this issue in JIRA).

app/src/controllers/log/storageUtils.js
```diff
@@ -17,36 +17,23 @@
 import { getStorageItem, updateStorageItem } from 'common/utils';
 import { MARKDOWN } from 'common/constants/logViewModes';
 import {
-  DEFAULT_WITH_ATTACHMENTS,
```
ElenaRomanchuk: Remove it from constants too. It isn't used somewhere else.

AmsterGet: Thank you. Removed

PR #2339, opened 2020-03-13
EPMRPP-49009: Add description and custom block to NoItemMessage

DiscoElevator:


DiscoElevator:
Closed in favor of #2341

PR #2347, opened 2020-03-20
EPMRPP-48789 Log level. Update history entry locally

AmsterGet:
Log level.
When user navigate from item to item via buttons on log view. And change a defect type of item, the arrow is disabled.


app/src/controllers/log/actionCreators.js
```diff
@@ -50,3 +51,8 @@ export const setPageLoadingAction = (isLoading) => ({
   type: SET_LOG_PAGE_LOADING,
   payload: isLoading,
 });
+
+export const updateHistoryEntryLocally = (testCaseHash, id, data) => ({
```
ElenaRomanchuk: The name of action doesn't clear represents its purpose. May be you need just action for setting issue to history entity.

app/src/pages/inside/logsPage/logItemInfo/logItemInfo.jsx
```diff
@@ -329,14 +332,26 @@ export class LogItemInfo extends Component {
     });
   };
 
+  onDefectEdited = (data) => {
+    const { logItem, fetchFunc } = this.props;
+
+    if (data) {
+      // on log level user can change defect only for one item
```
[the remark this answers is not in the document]
AmsterGet: Fixed

app/src/controllers/log/actionCreators.js
AmsterGet: Fixed

PR #2350, opened 2020-03-30
EPMRPP-49416: Add center align to ItemList

DiscoElevator:


app/src/components/inputs/inputRadio/inputRadio.jsx
```diff
@@ -31,27 +31,31 @@ export const InputRadio = ({
...
-      checked={value === ownValue}
-      name={name}
-    />
-    <span className={cx('toggler', { checked: value === ownValue, 'at-top': circleAtTop })} />
-    {children && <span className={cx('children-container')}>{children}</span>}
-  </label>
-);
+  checked,
+}) => {
+  const isChecked = checked || value === ownValue;
```
ElenaRomanchuk: I suggest don't overload component with additional `checked` props. `value` and `ownValue` are enough.

PR #2377, opened 2020-05-11
EPMRPP-45443: React. Group operations for items with similar failure reason. Launch name is missing in the hover state of item on Edit defect type modal

KatsiarynaTatarynovich:


KatsiarynaTatarynovich:
Will be done on the backend side.

PR #2384, opened 2020-05-14
EPMRPP-51655: Change values on general tab to days converted to seconds

DiscoElevator:


DiscoElevator:
Merge only after WS is done and merged to develop

PR #2390, opened 2020-05-20
EPMRPP-50447: Most Flaky test cases. Add tool tip with description how '% of switches' is calculated

KatsiarynaTatarynovich:


app/src/components/widgets/singleLevelWidgets/tables/components/testsTableWidget/testsTable/testsTableRow/testsTableRow.jsx
```diff
@@ -76,7 +107,23 @@ export class TestsTableRow extends Component {
           <span>{name}</span>
         </NavLink>
         {Matrix && count && (
-          <div className={cx('col', 'col-count')}>
+          <div
+            className={cx('col', 'col-count')}
+            title={
+              widgetType
+                ? formatMessage(titleMessages[widgetType], {
```
ElenaRomanchuk, line 114: It's not clear that widgetType will affect tooltip on matrix. Also component could be broken in case of no corresponding messages for specified widget type. What do you think about lifting up tooltip logic and pass callback `getMaxtrixTooltip` to the common components?

app/src/components/containers/localizationContainer/localizationContainer.jsx
```diff
@@ -46,7 +46,7 @@ export class LocalizationContainer extends React.Component {
     this.messages = {
       ru: localeRU,
       be: localeBE,
-      uk: localeUA,
+      uk: localeUK,
```
ElenaRomanchuk, line 49: `ua` instead of UK. UK is United Kingdom, not Ukraine.

PR #2391, opened 2020-05-21
EPMRPP-49116: Show details to user about tariff options

KatsiarynaTatarynovich:


app/src/components/main/modal/modalLayout/modalField/modalField.jsx
```diff
@@ -16,6 +16,8 @@
 
 import PropTypes from 'prop-types';
 import classname from 'classnames/bind';
+import Parser from 'html-react-parser';
+import FilterIcon from 'common/img/info-inline.svg';
```
ElenaRomanchuk, line 20: Maybe `InfoIcon`?

app/src/components/main/modal/modalLayout/modalField/modalField.jsx
```diff
@@ -28,6 +30,8 @@ export const ModalField = ({
   labelWidth,
   alignLeft,
   noMinHeight,
+  showInfoIcon,
```
ElenaRomanchuk, line 33: I think the `showInfoIcon` property is redundant because the condition of show tooltip is title existence. So I suggest to rename `iconTitle` to smth like `labelTip` and use it in the condition.

PR #2406, opened 2020-06-16
EPMRPP-50434 Fix edit launch modal for empty owner

AmsterGet:


app/src/pages/inside/common/modals/editItemModal/editItemModal.jsx
```diff
@@ -235,14 +235,28 @@ export class EditItemModal extends Component {
...
+      userId,
+    } = this.props;
+
+    if (item.owner) {
+      return userId === item.owner;
+    } else if (parentLaunch && parentLaunch.owner) {
+      return userId === parentLaunch.owner;
+    }
+
+    return true;
```
KatsiarynaTatarynovich, line 250: Why exactly this meaning (`true`)? Do conditions act as negation?

AmsterGet: It means that launch (item) haven't any owner and can be edited by anyone.

PR #2417, opened 2020-06-24
[2] EPMRPP-52640: [UI] Breadcrumbs component, EPMRPP-52645: [UI] Check correct work with BackEnd side

KatsiarynaTatarynovich:


app/src/components/main/grid/gridHeader/gridHeader.jsx
```diff
@@ -41,6 +41,9 @@ export const GridHeader = ({
         key={column.id || i}
         title={column.title}
         align={column.align}
+        width={column.width}
```
AmsterGet, line 44: `width` sounds as absolute width value. I think that name f.e. `size` will be better.

PR #2442, opened 2020-08-03
EPMRPP-48983: React. Admin part. Icon for Billing plugin on is not added

KatsiarynaTatarynovich:


app/src/components/integrations/containers/integrationInfoContainer/integrationInfoContainer.jsx
```diff
@@ -77,7 +77,7 @@ export class IntegrationInfoContainer extends Component {
     return (
       <Fragment>
         <InfoSection
-          image={PLUGIN_IMAGES_MAP[name]}
+          image={PLUGIN_IMAGES_MAP[name] || PLUGIN_DEFAULT_IMAGE}
```
AmsterGet, line 80: We can create special util function smth like `getPluginImage` or `getPluginIcon` to hide logic and data structure associated with plugin icons.
What do you think?

AmsterGet: Decided to leave it as is.

PR #2464, opened 2020-10-01
EPMRPP-46496 Delete items modal. Remove warning for launch owner

AmsterGet:


app/src/controllers/testItem/utils.js
```diff
@@ -121,3 +121,13 @@ export const groupItemsByParent = (items) =>
       [item.parent]: [...group, item],
     };
   }, {});
+
+export const isItemOwner = (userId, item, launch) => {
+  if (item.owner) {
```
KatsiarynaTatarynovich, line 126: Should we convince that `item` exists?

KatsiarynaTatarynovich (changes requested):

app/src/controllers/testItem/utils.js
AmsterGet: It is required to use this utility with the ʻitem` property.
It is called ʻisItemOwner` to check the user's ownership of the corresponding item.
Therefore, it is the developer's responsibility to provide the required arguments to use this feature.

AmsterGet: And it is just copied from https://github.com/reportportal/service-ui/pull/2406/files#diff-dd4a7cd981a131a9b1d016eefd80f869R238

KatsiarynaTatarynovich (approved):

PR #2466, opened 2020-10-01
EPMRPP-52021: [UI] Update designs for Project plans view

KatsiarynaTatarynovich:


app/src/components/buttons/ghostButton/ghostButton.scss
```diff
@@ -90,6 +90,20 @@
       border-color: transparent;
       border-radius: 2px;
     }
+
+    &.transparent-border-hover {
```
AmsterGet, line 94: Why not use the same color as [here](https://github.com/reportportal/service-ui/pull/2466/files#diff-d77f56b743ba16c57f49e68360668a74R76-R87)?

AmsterGet (approved):

PR #2490, opened 2020-10-29
EPMRPP-56833: UI. Implement filter entities

KatsiarynaTatarynovich:


app/src/pages/inside/common/filterEntitiesGroups/stepLevelEntities.jsx
```diff
@@ -675,50 +668,11 @@ export class StepLevelEntities extends Component {
...
-            },
-          },
-        ],
-        ...rest,
-      );
-    } else {
-      this.props.onFilterAdd(entity, ...rest);
-    }
-  };
+  handleAdd = (entity, ...rest) => this.props.onFilterAdd(entity, ...rest);
```
AmsterGet, line 671: It looks like these methods are no longer needed and should be replaced with the value from props in the component's render method.

AmsterGet (changes requested):

AmsterGet:
Can be merged after https://github.com/reportportal/service-ui/pull/2491

AmsterGet (approved):

PR #2501, opened 2020-11-05
EPMRPP-57124 Most time consuming test cases widget. Fix widget view. Fix axis captions

AmsterGet:


app/src/components/widgets/common/utils/utils.js
```diff
@@ -143,3 +144,14 @@ export const getDefaultTestItemLinkParams = (projectId, filterId, testItemIds) =
   },
   type: TEST_ITEM_PAGE,
 });
+
+export const getTimeType = (max) => {
+  if (max > 0) {
```
KatsiarynaTatarynovich, line 149: Can we use the moment here? As a result, we will not need time markers

KatsiarynaTatarynovich (changes requested):

KatsiarynaTatarynovich (approved):

PR #2502, opened 2020-11-05
EPMRPP-57146: React. Cumulative trend chart and Component health check shouldn't be force updated after updating of widget name/description/sharing option

KatsiarynaTatarynovich:


app/src/pages/inside/dashboardItemPage/modals/editWidgetModal/editWidgetModal.jsx
```diff
@@ -132,6 +133,12 @@ export class EditWidgetModal extends Component {
     } = this.props;
 
     const data = prepareWidgetDataForSubmit(this.preprocessOutputData(widgetSettings));
+    const isForceUpdated =
```
AmsterGet, line 136: Maybe `isForceUpdateNeeded`, because we need to determine if an update is needed before taking action.

AmsterGet (approved):

PR #2509, opened 2020-11-17
EPMRPP-57664: Cumulative trend chart. Re-write with new logic. Change the order of colors

KatsiarynaTatarynovich:


app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/chartjsConfig.js
```diff
@@ -23,8 +23,8 @@ const Color = require('color');
 const getDefects = (fields) => fields.filter((item) => /defects/.test(item));
 
 const getExecutions = () => [
-  'statistics$executions$passed',
   'statistics$executions$failed',
+  'statistics$executions$passed',
   'statistics$executions$skipped',
```
AmsterGet, line 28: I think that will be good to separate passed items from failed and skipped and not mix them (see the screenshot below), so the order will be failed, skipped, passed.

image

app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/chartjsConfig.js
```diff
@@ -23,8 +23,8 @@ const Color = require('color');
...
 ];
```
AmsterGet, line 30: Also please clarify the order of the legend items with BA (see the screenshot below), because it is also controlled by the `executions` order.

image

KatsiarynaTatarynovich: BA is ok with this, add a description in the subtask

AmsterGet (approved):

PR #2510, opened 2020-11-17
EPMRPP-57663: Cumulative trend chart. Re-write with new logic. Change widget view - remove the tumbler

KatsiarynaTatarynovich:


KatsiarynaTatarynovich:
Do not need these logic's updates

AmsterGet (approved):

PR #2516, opened 2020-11-18
EPMRPP-57746: Cumulative trend chart. Re-write with new logic. Change a way how defect types % is calculated

KatsiarynaTatarynovich:


app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/chartjsConfig.js
```diff
@@ -160,7 +161,14 @@ const getChartOptions = (widget, options) => {
           if (!value) {
             return '';
           }
-          const totalValue = totalDataset.data[tooltipItem.index];
+          const isDefectType = /defects/.test(dataset.label);
+          const totalDefectsValue = defects.reduce(
```
AmsterGet, line 165: I think it superfluous to calculate statistics on defects for any label every time.
Maybe it's better to transfer this calculation to the condition block?
F.e.

```
          let totalValue;
          const isDefectType = /defects/.test(dataset.label);

          if (isDefectType) {
            const defects = data.datasets.filter((item) => /defects/.test(item.label));

            totalValue = defects.reduce(
              (total, defect) => total + defect.data[tooltipItem.index],
              0,
            );
          } else {
            totalValue = totalDataset.data[tooltipItem.index];
          }
```

AmsterGet (changes requested):

AmsterGet (approved):

PR #2524, opened 2020-11-23
EPMRPP-56810 Remove app header

AmsterGet:


app/src/layouts/appLayout/appSidebar/appSidebar.jsx
```diff
@@ -63,38 +79,69 @@ export class AppSidebar extends Component {
...
+      {
+        onClick: () => this.onClickButton(SIDEBAR_EVENTS.CLICK_FILTERS_BTN),
+        link: { type: PROJECT_FILTERS_PAGE, payload: { projectId: activeProject } },
+        icon: FiltersIcon,
+        message: <FormattedMessage id={'Sidebar.filtersBtn'} defaultMessage={'Filters'} />,
+      },
+    ];
+
+    if (projectRole !== CUSTOMER) {
+      topItems.push({
```
KatsiarynaTatarynovich, line 110: We should pop this item in case of projectRole === CUSTOMER. Because now, if I change the role from PM to CUSTOMER (f.e.) I will be able to see debug button

KatsiarynaTatarynovich (changes requested):

app/src/layouts/appLayout/appSidebar/appSidebar.jsx
AmsterGet: Decided to leave it as is.

KatsiarynaTatarynovich (approved):

PR #2539, opened 2020-12-10
EPMRPP-57880: Cumulative trend chart. Improve clickable area. Implement clickable area for Cumulative trend chart

KatsiarynaTatarynovich:


app/src/controllers/testItem/utils.js
```diff
@@ -132,3 +132,14 @@ export const isItemOwner = (userId, item, launch) => {
 
   return true;
 };
+
+export const getProviderTypeInfo = (queryProviderType, providerType, id) => {
```
ElenaRomanchuk, line 136: It seems this function breaks Single-responsibility and Interface segregation principles of SOLID best practice, and doesn't implicitly do the thing named in its signature. I think it shouldn't contain checking of existence `queryProviderType`. To avoid overriding of query params you can spread it after calculated values: `{...calculatedProviderType, ...query}`. Maybe you don't need this function at all.

ElenaRomanchuk (approved):

PR #2540, opened 2020-12-13
EPMRPP-57707 Add extension components to launches page

AmsterGet:


KatsiarynaTatarynovich (approved):

app/src/controllers/plugins/uiExtensions/registerPlugin.jsx
```diff
@@ -12,7 +12,11 @@ const createPluginRegistrationFunction = (store) => (plugin) => {
   const wrappedExtensions = extensions.map((extension, i) => ({
     name: `${plugin.name}__${i}`,
     ...extension,
-    component: <extension.component {...createImportProps(plugin.name)} />,
+    component: extension.factory ? (
```
ElenaRomanchuk, line 15: What does it mean `factory` property? Is it a flag? If so, I recommend to rename it to smth like `isFactory`. If it's a real factory, why do you use `component` as a factory? And why do you need factory here? Why it's impossible to use existed realization?

AmsterGet: Agree with you, in this case it would be better to use the `factory` property directly to create the component.
After that, `factory` will mean the function that will create React components.
Speaking of the existing approach - we have no way to use the `component` property as a valid React component (with props, children, etc.), because `<extension.component {... createImportProps (plugin.name)} />` creates a React element that is an object and cannot be used as a function.

ElenaRomanchuk: I think, `extension.component` should be predictable and have the same type and the same usage in every case. I think we can do this without a factory by just inline wrapper component like:
`component: (props) => <extension.component {...createImportProps(plugin.name)} {...props}/>` 
And fix usage of existing plugin components according to this approach.

ElenaRomanchuk (changes requested):

ElenaRomanchuk (approved):

PR #2547, opened 2020-12-22
EPMRPP-58586: React. Incorrect behavior when updating type defect

KatsiarynaTatarynovich:


app/src/common/urls.js
```diff
@@ -156,7 +156,8 @@ export const URLS = {
     `${urlBase}project/${activeProject}/notification`,
   suite: (activeProject, suiteId) => `${urlBase}${activeProject}/item/${suiteId}`,
 
-  testItems: (activeProject, ids) => `${urlBase}${activeProject}/item/v2${getQueryParams({ ids })}`,
+  testItems: (activeProject, version = '', ids) =>
```
AmsterGet, line 159: It makes sense for a parameter to leave its default value when it is the last one in the function parameters.
So it is better to move the `version` parameter to the end of the parameter sequence.

AmsterGet (changes requested):

ElenaRomanchuk (approved):

PR #2570, opened 2021-02-02
EPMRPP-58973: Add new end-points

KatsiarynaTatarynovich:


app/src/common/urls.js
```diff
@@ -89,6 +89,17 @@ export const URLS = {
     `${urlBase}${activeProject}/launch/attribute/values?${
       key ? `filter.eq.attributeKey=${key}&` : ''
     }filter.cnt.attributeValue=${searchTerm}`,
+  itemAttributeKeysByLaunchName: (activeProject, launchName = '') => (searchTerm = '') =>
+    `${urlBase}${activeProject}/item/step/attribute/keys?${
```
AmsterGet, line 93: Can the `getQueryParams` util be used here?

ElenaRomanchuk: Agree. `getQueryParams` also solves the problem with encryption symbols in URL

ElenaRomanchuk (changes requested):

app/src/common/urls.js
KatsiarynaTatarynovich: @AmsterGet @ElenaRomanchuk the method getQueryParams is not suitable for us, since it turns into a string with the saved key if the value is empty. In this case, the key should not be part of the parameter at all.

ElenaRomanchuk: @KatsiarynaTatarynovich You should encrypt `launchName` in URL. So, I advise to use `getQueryParams` but exclude redundant property by the condition.

ElenaRomanchuk (changes requested):

ElenaRomanchuk (approved):

app/src/common/urls.js
```diff
@@ -93,6 +93,17 @@ export const URLS = {
...
       key ? `filter.eq.attributeKey=${key}&` : ''
     }filter.cnt.attributeValue=${searchTerm}`,
+  itemAttributeKeysByLaunchName: (activeProject, launchName) => (searchTerm = '') =>
+    `${urlBase}${activeProject}/item/step/attribute/keys${getQueryParams({
+      'filter.eq.name': launchName || undefined,
+      'filter.cnt.attributeKey': searchTerm,
+    })}`,
+  itemAttributeValuesByLaunchName: (activeProject, launchName, key) => (searchTerm = '') =>
+    `${urlBase}${activeProject}/item/step/attribute/values${getQueryParams({
+      'filter.eq.name': launchName || undefined,
```
AmsterGet, line 103: Can these properties differ from `undefined` or `some_string_value`?

KatsiarynaTatarynovich: No, it can't

PR #2571, opened 2021-02-04
EPMRPP-58976 Return attachment modal for images

AmsterGet:


app/src/pages/inside/logsPage/modals/attachments/attachmentImageModal.jsx
```diff
@@ -0,0 +1,144 @@
...
+      this.nativeImageSize = this.getImageSize();
+    }
+  };
+
+  generateRotationCommand = (amount) => `rotate(${amount}deg)`;
+
+  rotateImageHandler = () => {
+    const rotationA = this.state.rotationAmount;
+    this.props.tracking.trackEvent(LOG_PAGE_EVENTS.ROTATE_ICON_ATTACHMENT_MODAL);
+    this.setState({
```
ElenaRomanchuk, line 93: Use callback to increment state value to avoid batching error.

app/src/pages/inside/logsPage/modals/attachments/attachmentImageModal.jsx
```diff
@@ -0,0 +1,144 @@
...
+        imageStyle.width =
+          (rotationAmount / 90) % 2 === 1
+            ? `${this.nativeImageSize.height}px`
+            : `${this.nativeImageSize.width}px`;
+      }
+    }
+    return imageStyle;
+  };
+
+  initNativeImageSize = () => {
```
ElenaRomanchuk, line 82: It looks like you need memoization here. This allows you to avoid creating new redundant class property `nativeImageSize` and also won't trigger an unnecessary reflow process. And then you can exclude calculation from the component class.

ElenaRomanchuk (changes requested):

app/src/pages/inside/logsPage/modals/attachments/attachmentImageModal.jsx
AmsterGet: Fixed

app/src/pages/inside/logsPage/modals/attachments/attachmentImageModal.jsx
AmsterGet: Added to Tech Debt as a low priority fix.

ElenaRomanchuk (approved):

PR #2578, opened 2021-02-16
EPMRPP-58985 - Change background for pages

chivekrodis:


app/src/components/preloaders/spinningPreloader/spinningPreloader.scss
```diff
@@ -19,6 +19,7 @@
   position: relative;
   width: 100%;
   height: 100%;
+  background-color: $COLOR--white-two;
```
AmsterGet, line 22: I think will be better to leave this component without background as it can be used in different places with different backgrounds and may inherit it from parents.

AmsterGet (changes requested):
Also please add white background to the log view (available from test item pages - see the screenshot below).
image

The GhostButton must have an option to be with transparent background (as in designs - see the screenshot below)
image

AmsterGet (approved):

PR #1282, opened 2019-03-20
EPMRPP-37347. Admin. Projects. Single panel.

ElenaRomanchuk:
EPMRPP-37347. Admin. Projects. Single panel.

PR #2580, opened 2021-02-16
EPMRPP-59007 - Create filter on launches page without saving and open launch in new tabs leads to infinite loading

chivekrodis:
[#1282](https://github.com/reportportal/reportportal/issues/1282)

### DIFF
--- a/app/src/controllers/testItem/sagas.js
+++ b/app/src/controllers/testItem/sagas.js
@@ -35,6 +35,7 @@ import {
   PROJECT_LOG_PAGE,
   filterIdSelector,
   pageSelector,
+  testItemIdsSelector,
 } from 'controllers/pages';
 import { PAGE_KEY } from 'controllers/pagination';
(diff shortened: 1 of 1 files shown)


app/src/controllers/testItem/sagas.js
```diff
@@ -190,10 +192,22 @@ function* fetchTestItems({ payload = {} }) {
...
+      }
+    } catch {
+      const testItemIds = yield select(testItemIdsSelector);
+      const link = {
+        type: TEST_ITEM_PAGE,
+        payload: {
+          filterId: ALL,
+          projectId: project,
+          testItemIds,
+        },
```
AmsterGet, line 208: For correct redirection to the required page, please provide the `query` property to receive all necessary parameters from URL:
```
meta: {
   query: pageQuery,
},
```

AmsterGet (changes requested):

AmsterGet (approved):

PR #2591, opened 2021-02-22
EPMRPP-45170 - React. Show all method type on item test page after redirect from failed trend case chart

chivekrodis:


AmsterGet (changes requested):
From the task description:
image
Please also provide fix for these widgets.

AmsterGet (approved):

PR #2594, opened 2021-02-25
EPMRPP-59834 Add clickable numbers for rules

katrin-kot:


app/src/controllers/testItem/selectors.js
```diff
@@ -297,19 +297,26 @@ export const statisticsLinkSelector = createSelector(
...
       isLatest,
     };
-
     if (ownProps.statuses) {
       params['filter.in.status'] = ownProps.statuses.join(',');
     }
     if (ownProps.types === null) {
       delete params['filter.in.type'];
     }
+    if (ownProps.key === null) {
```
AmsterGet, line 314: No need to do this check, because if the `ownProps.key` or `ownProps.value` will be undefined the parameter will be ommited from the link.
Also please rename the properties to `attributeKey` and `attributeValue` to be more clear.

AmsterGet (changes requested):

app/src/controllers/plugins/uiExtensions/createImportProps.js
```diff
@@ -231,11 +256,20 @@ export const createImportProps = (pluginName) => ({
...
+    getItemNameConfig,
+    getDefectTypeLocators,
+  },
   validators: {
     attributesArray,
     isNotEmptyArray,
     requiredField,
-    helpers: { composeValidators, bindMessageToValidator },
   },
+  helpers: { composeValidators, bindMessageToValidator },
```
AmsterGet, line 274: Why are you moving this `helpers` object up, are you fixing this also from the plugin side to avoid errors? These helpers are part of the validation engine, so it makes sense to store them along with the validation functions.

AmsterGet (changes requested):

PR #2597, opened 2021-03-01
EPMRPP-59881 Add parameter with `type: 'table'`

katrin-kot:

PR #2599, opened 2021-03-02
EPMRPP-59882 Add launch name to the item tooltip

katrin-kot:


app/src/pages/inside/logsPage/historyLine/historyLineItem/historyLineItemTooltip/historyLineItemTooltip.jsx
```diff
@@ -62,6 +66,12 @@ export const HistoryLineItemTooltip = injectIntl(
           {statusLocalization[status] ? formatMessage(statusLocalization[status]) : status}
         </span>
       </span>
+      <div className={cx('launch-name-block')}>
```
AmsterGet, line 69: This content in the tooltip should only appear when `includeAllLaunches` is `true`, otherwise all elements will have the same parent launch and don't need to be highlighted.
Can be done after [this](https://github.com/reportportal/service-ui/pull/2597#pullrequestreview-601540918)

AmsterGet (changes requested):

app/src/pages/inside/logsPage/historyLine/historyLineItem/historyLineItemTooltip/historyLineItemTooltip.scss
```diff
@@ -57,8 +57,21 @@
   }
 }
 
+.launch-name-value {
+  width: 279px;
+  font-size: 13px;
+  font-weight: 600;
+  font-family: $FONT-SEMIBOLD;
+  word-wrap: break-word;
+  display: -webkit-box;
```
ElenaRomanchuk, line 66: `display: -webkit-box;` doesn't work in IE. I suggest using `@supports` at-rule in Sass to introduce fallback for IE and any other browsers.

ElenaRomanchuk (changes requested):

AmsterGet (approved):

PR #2605, opened 2021-03-04
EPMRPP-60308 - fix item height

chivekrodis:


AmsterGet (changes requested):
Don't forget when the component you are changing is being used in other places.
Patterns:
**Before**:
image

**After**:
image

PR #2612, opened 2021-03-11
EPMRPP-60011 Disable Unique Errors tab

katrin-kot:


app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
```diff
@@ -110,7 +114,7 @@ export class ViewTabs extends Component {
       {
         id: UNIQUE_ERRORS_VIEW,
         title: formatMessage(messages[UNIQUE_ERRORS_VIEW]),
-        link: listViewLink,
+        link: {},
```
chivekrodis, line 117: do we really need a redirect to root? we can stay on the same page
you can use CSS ```pointer-events: none``` to prevent any event on the disabled button

app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
```diff
@@ -143,7 +150,13 @@ export class ViewTabs extends Component {
           <Link
             key={page.id}
             to={page.link}
-            className={cx('view-tab-link', { active: viewMode === page.id })}
+            title={
+              page.id === UNIQUE_ERRORS_VIEW && formatMessage(messages.uniqueErrorsViewTooltip)
+            }
+            className={cx('view-tab-link', {
+              active: viewMode === page.id && page.id !== UNIQUE_ERRORS_VIEW,
+              disabled: page.id === UNIQUE_ERRORS_VIEW,
```
chivekrodis, line 158:
```
page.id === UNIQUE_ERRORS_VIEW
```
construction repeat couple times maybe better extract to single variable

chivekrodis (changes requested):

app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
katrin-kot: No redirection, we stay on the same page

chivekrodis: URL doesn't change after click on this tab?

katrin-kot: URL changes,  'pointer-events: none' doesn't match,  selectLocationState(state) refreshes the page

app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
```diff
@@ -134,21 +138,31 @@ export class ViewTabs extends Component {
...
-            {page.title}
-          </Link>
-        ))}
+        {pages.map((page) => {
+          const isUniqueErrorsViewTab = page.id === UNIQUE_ERRORS_VIEW;
+          return (
+            <Link
+              key={page.id}
+              to={page.link}
+              title={isUniqueErrorsViewTab && formatMessage(messages.uniqueErrorsViewTooltip)}
```
AmsterGet, line 155: This needs to be done at the configuration level - it would be a bit odd to add such conditions to other pages if needed in the future.
The best solution here is to provide additional properties to `pages` array (`disabled` and `hint`) for `UNIQUE_ERRORS_VIEW` item.

AmsterGet (changes requested):

app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
```diff
@@ -110,27 +110,35 @@ export class ViewTabs extends Component {
...
+        hint: false,
       },
       {
         id: UNIQUE_ERRORS_VIEW,
         title: formatMessage(messages[UNIQUE_ERRORS_VIEW]),
         link: {},
         icon: ListIcon,
         available: isStepLevel && !isTestItemsList,
+        disabled: true,
+        hint: true,
```
AmsterGet, line 123: You can specity necessary message just here

app/src/pages/inside/common/infoPanel/viewTabs/viewTabs.jsx
```diff
@@ -106,49 +110,66 @@ export class ViewTabs extends Component {
...
         icon: ListIcon,
         available: true,
+        disabled: false,
+        hint: false,
       },
       {
         id: UNIQUE_ERRORS_VIEW,
         title: formatMessage(messages[UNIQUE_ERRORS_VIEW]),
-        link: listViewLink,
+        link: {},
```
AmsterGet, line 119: Empty link action will cause error (see the screenshot below)
image
To avoid this make sense to leave previous link (`listViewLink`) here and specify `shouldDispatch={!page.disabled}` as `Link` prop (https://github.com/faceyspacey/redux-first-router-link#additional-props) or simply do not render `Link` component at all for disabled items (render just content).

AmsterGet (changes requested):

AmsterGet (approved):

PR #2618, opened 2021-03-15
EPMRPP-53936 - fix integration pass field

chivekrodis:


katrin-kot (approved):

app/src/components/integrations/containers/integrationSettingsContainer/integrationSettingsContainer.jsx
```diff
@@ -57,6 +59,7 @@ export class IntegrationSettingsContainer extends Component {
       enabled: true,
       integrationParameters: formData,
     };
+    const integrationParameters = omit(data.integrationParameters, OMIT_KEYS);
```
AmsterGet, line 62: Such `omit` already exists in the `src/controllers/plugins/sagas.js` `function* updateIntegration`.
To reuse this normalized `integrationParameters` make sense to provide integration data to the callback from the saga (line 127) and use it [here](https://github.com/reportportal/service-ui/pull/2618/files#diff-1b7acf2538fb238032aed222558b2f44ad5cb896602ad6bf286658e8cfb4908eR73).

AmsterGet (changes requested):

app/src/components/integrations/containers/integrationSettingsContainer/integrationSettingsContainer.jsx
```diff
@@ -70,9 +67,9 @@ export class IntegrationSettingsContainer extends Component {
       isGlobal,
       id,
       pluginName,
-      () => {
+      (normalizeData) => {
```
AmsterGet, line 70: `normalizeData` sounds like an action, may be name it just `updatedParameters` to use property shorthand?

AmsterGet (approved):

PR #2626, opened 2021-03-18
EPMRPP-44322 Removed percentage ratio for  0 in Passing rate per launch and Passing rate summary widgets

katrin-kot:


chivekrodis (approved):

app/src/components/widgets/singleLevelWidgets/charts/common/passingRateChart/passingRateChart.jsx
```diff
@@ -92,6 +92,9 @@ export class PassingRateChart extends Component {
 
     // eslint-disable-next-line func-names
     d3.selectAll(nodeElement.querySelectorAll('.bar .c3-chart-texts .c3-text')).each(function(d) {
+      if (d.value === 0) {
+        this.textContent = '';
+      }
```
AmsterGet, line 97: This solves the problem, but it's not entirely clear what is going on and why we need this line when resizing helpers.
So I suggest to fix this issue inside of `src/components/widgets/singleLevelWidgets/charts/common/passingRateChart/config/getConfig.js` when we exactly format labels:
```
labels: {
      show: !isPreview,
      format: (value) => (isPreview || value === 0 ? '' : `${getPercentage(value, totalItems)}%`),
},
```

AmsterGet (changes requested):

AmsterGet (approved):

PR #2632, opened 2021-03-23
EPMRPP-44728 Fixed bug with 'X' button if editing mode is opened for custom defect type

katrin-kot:


chivekrodis (approved):

app/src/pages/common/settingsPage/defectTypesTab/defectSubType.jsx
```diff
@@ -208,7 +208,7 @@ export class DefectSubType extends Component {
           <DefectSubTypeForm
             form={data.locator}
             initialValues={data}
-            onDelete={this.showDeleteConfirmationDialog}
+            onDelete={this.stopEditing}
```
AmsterGet, line 211: Now it isn't delete action (`onCancel` prop name will be better here). Could you please also rename corresponding stuff inside of `DefectSubTypeForm` to `cancel` instead of `delete`.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2641, opened 2021-03-29
EPMRPP-58068 Allow custom protocols in markdown

AmsterGet:
Waiting approval from Security team.
Alternative solution: specify allowed link protocols by regexp.


AmsterGet:
Closed due to no response from security. The final solution will be investigated separately.

PR #2645, opened 2021-03-31
 EPMRPP-60628 Adopted modal content for bulk operation

katrin-kot:


app/src/pages/inside/stepPage/modals/editDefectModals/makeDecisionModal/makeDecisionModal.jsx
```diff
@@ -67,7 +67,9 @@ const MakeDecision = ({ data }) => {
   const [issueAction, setIssueAction] = useState({});
 
   useEffect(() => {
-    setModalHasChanges(!isEqual(itemData.issue, state.issue) || !isEmptyObject(issueAction));
+    setModalHasChanges(
+      state.issue ? !isEqual(itemData.issue, state.issue) || !isEmptyObject(issueAction) : false,
```
AmsterGet, line 71: The user should have possibility to interact with issue action even when there is no issue in the state.
Also notice how the item data gets to the next action.

AmsterGet (changes requested):

PR #2651, opened 2021-04-02
EPMRPP-60941 Fixed unlink action in bulk operation

katrin-kot:


app/src/controllers/groupOperations/utils.js
```diff
@@ -18,7 +18,7 @@ import { EXECUTE_GROUP_OPERATION } from './constants';
 
 export const validateItems = (items = [], validator, state) =>
   items.reduce((acc, item) => {
-    const error = validator(item, items, state);
+    const error = validator && validator(item, items, state);
```
chivekrodis, line 21: Should we put items without external issues to unlink modal? We can omit them  in prepareDataToSend

chivekrodis (changes requested):

app/src/controllers/step/actionCreators.js
```diff
@@ -52,12 +51,7 @@ export const includeInAutoAnalysisAction = defineGroupOperation(
   validateIncludeInAA,
 );
 
-export const unlinkIssueAction = defineGroupOperation(
-  NAMESPACE,
-  'unlink-issue',
-  unlinkIssue,
-  validateUnlinkIssue,
```
AmsterGet, line 59: Your changes break the validation process for an independent `unlink` operation.
The existing validation should be left as is.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2653, opened 2021-04-06
EPMRPP-61003  Fixed visibility of Comment icon

katrin-kot:


app/src/pages/inside/stepPage/modals/editDefectModals/makeDecisionModal/optionsStepForm/optionsStepForm.jsx
```diff
@@ -37,14 +37,16 @@ export const OptionsStepForm = ({ info }) => {
 
   const renderCommentBlock = () => {
```
AmsterGet, line 38: It may be best not to call this function at all if there is no issue comment.

AmsterGet (approved):

PR #2666, opened 2021-04-15
EPMRPP-60569 - add All loaded TI from the history line option

chivekrodis:


app/src/pages/inside/stepPage/modals/editDefectModals/makeDecisionModal/optionsStepForm/optionsBlock/optionsBlock.jsx
```diff
@@ -89,8 +92,43 @@ export const OptionsBlock = ({
...
+      ...historyItems
+        .reduce((items, item) => {
+          const currentDefectType =
+            (item.issue && TIDefectsGroup.find((type) => type.locator === item.issue.issueType)) ||
+            {};
+          return currentDefectType.typeRef === TO_INVESTIGATE.toUpperCase() &&
+            item.issue &&
+            item.id !== currentTestItem.id
+            ? [...items, { ...item, itemId: item.id }]
+            : items;
```
KatsiarynaTatarynovich, line 121: Since `item.issue` appears in two conditions, it should be taken out in a separate check, this will simplify the perception of the entire code.
You could also take out `item.id !== currentTestItem.id`.
Something like
`if (!item.issue && item.id !== currentTestItem.id) {`
           `return items;`
        `  }`
          `const currentDefectType = TIDefectsGroup.find((type) => type.locator === item.issue.issueType) || {};`
         ` return currentDefectType.typeRef === TO_INVESTIGATE.toUpperCase()`
           ` ? [...items, { ...item, itemId: item.id }]`
           ` : items;`

KatsiarynaTatarynovich (changes requested):

PR #2669, opened 2021-04-16
EPMRPP-61294 Disabled the view of "Machine..." tab

katrin-kot:


chivekrodis (approved):

app/src/pages/inside/common/accordion/accordion.jsx
```diff
@@ -31,7 +31,10 @@ export const Accordion = ({ tabs, headerClassNames, contentClassNames, toggleTab
             key={tab.id}
             tab={tab}
             onClick={() => toggleTab(tab.id)}
-            headerClassNames={tab.shouldShow ? headerClassNames : 'hidden'}
+            headerClassNames={cx(headerClassNames, {
```
KatsiarynaTatarynovich, line 34: Are you sure that `headerClassNames` and `contentClassNames` pass as props? I do not see their transfers in the code.

KatsiarynaTatarynovich (changes requested):

app/src/pages/inside/common/accordion/accordion.jsx
katrin-kot: This is an optional props, but if it is, then it is passed like this.

KatsiarynaTatarynovich: But I don't see that we are passing these parameters at all. Do we really need them?

katrin-kot: Yes, they are not needed yet, I will delete them

app/src/pages/inside/common/accordion/accordion.jsx
```diff
@@ -31,25 +31,21 @@ export const Accordion = ({ tabs, headerClassNames, contentClassNames, toggleTab
             key={tab.id}
             tab={tab}
             onClick={() => toggleTab(tab.id)}
-            headerClassNames={cx(headerClassNames, {
+            headerClassNames={cx({
```
KatsiarynaTatarynovich, line 34: Since we no longer need `headerClassNames` and `contentClassNames`, there is no need to pass them as props to the child component. We should work with `classNames` directly in `AccordionTab` component using property from `tab` props.

KatsiarynaTatarynovich (changes requested):

PR #2670, opened 2021-04-16
EPMRPP-60912 - Unification of all preloaders

chivekrodis:


app/src/pages/inside/common/durationBlock/durationBlock.jsx
```diff
@@ -170,7 +170,7 @@ export class DurationBlock extends Component {
     return (
       <Fragment>
         <span className={cx('in-progress')}>
-          <img src={InProgressGif} alt="In progress" />
+          <DottedPreloader color="black" />
```
KatsiarynaTatarynovich, line 173: Have you discussed the view with the designer?

KatsiarynaTatarynovich (changes requested):

app/src/layouts/common/sidebar/userBlock/userBlock.jsx
```diff
@@ -35,21 +35,24 @@ import styles from './userBlock.scss';
...
 
-const UserAvatar = ({ photoTimeStamp }) => (
+const UserAvatar = ({ photoTimeStamp, preloaderColor }) => (
   <div className={cx('avatar-wrapper')}>
     <Image
       className={cx('avatar')}
       src={URLS.dataPhoto(photoTimeStamp, true)}
       alt="avatar"
       fallback={DefaultUserImage}
+      preloaderColor={preloaderColor}
```
KatsiarynaTatarynovich, line 45: Could you please explain why you decided to work with color through a parent if this method is not used anywhere else?

KatsiarynaTatarynovich (changes requested):

app/src/layouts/common/sidebar/userBlock/userBlock.jsx
chivekrodis: `UserAvatar` uses in sidebar with tooltip we need here a white color preloader. Also, `UserAvatar` uses in pop-up by clicking on user avatar and here we need a charcoal color. That reason why I have needed lifting this props.

chivekrodis: Duration block, preloader on admin panel for all users image, attachments with photos, issue tooltip are having a charcoal color.
UserAvatar has white color.

KatsiarynaTatarynovich (approved):

PR #2681, opened 2021-04-28
EPMRPP-61866 Changed icon for "All logs" tab

katrin-kot:


app/src/pages/inside/logsPage/logItemInfo/infoTabs/infoTabs.scss
```diff
@@ -47,6 +47,9 @@
     color: $COLOR--charcoal-grey;
     fill: $COLOR--charcoal-grey;
   }
+  &:nth-child(2):hover svg {
```
AmsterGet, line 50: It is unreliable to use selector by item order here.
It would be better to control this case by providing property f.e. `stroked` from tab config.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2713, opened 2021-06-02
EPMRPP-62755 Fixed the position of the cross sign

katrin-kot:


app/src/pages/inside/stepPage/modals/linkIssueModal/linkIssueFields/linkIssueFields.scss
```diff
@@ -54,8 +54,8 @@
 
 .remove-issue-fields-button {
   position: absolute;
-  right: -4px;
-  top: 25px;
+  right: 5px;
+  top: 12px;
```
chivekrodis, line 58: seems like better align cross sign in the middle the `link to issue` field
please, discuss with task reporter
image

katrin-kot: Fixed the position to the level of the upper border

AmsterGet: I thnk it was a bug that the crosses where placed not near the inputs.
The best choice here is (see the screenshot below)

image

It is an initial placement for this crosses discussed with designer, so please move them according to the picture.

app/src/pages/inside/stepPage/modals/linkIssueModal/linkIssueFields/linkIssueFields.scss
```diff
@@ -54,8 +54,8 @@
 
 .remove-issue-fields-button {
   position: absolute;
-  right: -4px;
-  top: 25px;
+  right: 40%;
```
AmsterGet, line 57: Not the crosses locates in diffrent places in show less/more view.

Show less:

image

Show more:

image

May be better to use static indentation for the crosses.
Just add `display: inline-block` to the `issue-fields-list-item` and `right: -50px` to the `remove-issue-fields-button`.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2725, opened 2021-06-08
EPMRPP-62669 - add tooltip in case analyzer is not running

chivekrodis:


app/src/components/inputs/inputRadio/inputRadio.jsx
```diff
@@ -92,9 +96,6 @@ InputRadio.defaultProps = {
   onChange: () => {},
   onFocus: () => {},
   onBlur: () => {},
-  inputClassNames: {
-    togglerClassName: '',
-    childrenClassName: '',
-  },
   title: '',
+  mode: '',
```
AmsterGet, line 100: May be specify a meaningful default value, f.e. `mode: default`?

app/src/pages/inside/stepPage/modals/makeDecisionModal/makeDecisionModal.jsx
```diff
@@ -50,6 +51,9 @@ const MakeDecision = ({ data }) => {
   const dispatch = useDispatch();
   const activeProject = useSelector(activeProjectSelector);
   const historyItems = useSelector(historyItemsSelector);
+  const isAnalyzerAvailable = useSelector(analyzerExtensionsSelector).find(
```
AmsterGet, line 54: Why not just check `analyzerExtensions` length?

chivekrodis: How can I be sure that an existing extension is a analyzer?

app/src/components/inputs/inputRadio/inputRadio.jsx
chivekrodis: If I get you right, we should wrap old styles with `mode-default` class?

app/src/components/inputs/inputRadio/inputRadio.jsx
```diff
@@ -75,11 +82,8 @@ InputRadio.propTypes = {
   onChange: PropTypes.func,
   onFocus: PropTypes.func,
   onBlur: PropTypes.func,
-  inputClassNames: PropTypes.shape({
-    togglerClassName: PropTypes.string,
-    childrenClassName: PropTypes.string,
-  }),
   title: PropTypes.string,
+  mode: PropTypes.string,
```
AmsterGet, line 86: You may specify: `mode: PropTypes.oneOf(['default', 'dark'])`.

app/src/components/inputs/inputRadio/inputRadio.jsx
AmsterGet: 'default' mode value may be used in prop definitions just to specify the way how the mode should look like, without any special styles. But separation some specific, not common styles with `mode-default` sounds good too.

AmsterGet (approved):

PR #2729, opened 2021-06-11
EPMRPP-62781: decrease the space beetween buttons

oranje322:


app/src/components/buttons/ghostButton/ghostButton.scss
```diff
@@ -230,6 +230,9 @@
       background: transparent;
       color: $COLOR--topaz-hover;
     }
+    &.transparent-less-width {
+      min-width: 90px;
+    }
```
chivekrodis, line 235: What the purpose to create a new class and manage it with prop? Seems like we can add property directly to the `transparent-background` class.

chivekrodis (changes requested):

app/src/pages/inside/stepPage/modals/postIssueModal/postIssueModal.jsx
```diff
@@ -489,6 +489,7 @@ export class PostIssueModal extends Component {
           disabled={false}
           transparentBorder
           transparentBackground
+          transparentLessWidth
```
AmsterGet, line 492: Now it can be removed.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2740, opened 2021-06-16
EPMRPP-63017: editing of defect type undefined

oranje322:
typeof issue.comment is 'string'

### DIFF
--- a/app/src/pages/inside/stepPage/stepGrid/defectType/defectType.jsx
+++ b/app/src/pages/inside/stepPage/stepGrid/defectType/defectType.jsx
@@ -94,7 +94,7 @@ export const DefectType = track()(
       </div>
       <div className={cx('comment')}>
         <ScrollWrapper autoHeight autoHeightMax={90}>
-          <MarkdownViewer value={issue.comment} />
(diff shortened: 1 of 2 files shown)


app/src/pages/inside/stepPage/stepGrid/defectType/defectType.jsx
```diff
@@ -94,7 +94,7 @@ export const DefectType = track()(
       </div>
       <div className={cx('comment')}>
         <ScrollWrapper autoHeight autoHeightMax={90}>
-          <MarkdownViewer value={issue.comment} />
+          <MarkdownViewer value={issue.comment === 'undefined' ? '' : issue.comment} />
```
chivekrodis, line 97: It doesn't solve issue. You only hide comment with `undefined` string on step view, but if you open modal window for single test item you can see `undefined` in comment editor.
I guess better place to resolve this issue is `prepareDataToSend` in `makeDecisionModal` component. There we are manage what kind of information we send to the server.

chivekrodis (approved):

AmsterGet (approved):

PR #2742, opened 2021-06-17
EPMRPP-62981: buttons should not be displayed in debug mode

oranje322:


app/src/pages/inside/stepPage/modals/makeDecisionModal/makeDecisionModal.jsx
```diff
@@ -295,6 +297,7 @@ const MakeDecision = ({ data }) => {
             isBulkOperation={isBulkOperation}
             collapseTabsExceptCurr={collapseTabsExceptCurr}
             isNarrowView={!collapsedRightSection}
+            debugMode={debugMode}
```
chivekrodis, line 300: What the purpose to detect mode on top level component and passing it through the props? Seems it can be done directly in necessary component.

chivekrodis (changes requested):

AmsterGet (approved):

PR #2794, opened 2021-08-12
EPMRPP-64576 - add a status change btn to the log page

chivekrodis:


app/src/pages/inside/common/statusDropdown/statusDropdown.jsx
```diff
@@ -46,34 +49,41 @@ const messages = defineMessages({
 @connect(
   (state) => ({
     currentProject: activeProjectSelector(state),
+    currentPage: pageSelector(state),
```
AmsterGet, line 52: Not need to get the page type here, you can specify the necessary prop from the place where this component used.

app/src/pages/inside/common/statusDropdown/statusDropdown.jsx
```diff
@@ -46,34 +49,41 @@ const messages = defineMessages({
...
   }),
   {
-    fetchFunc: fetchTestItemsAction,
+    fetchTestItems: fetchTestItemsAction,
```
AmsterGet, line 55: The same is about `fetchFunc`. It will be better to provide necessary function directly from parent component instead of choosing it here.

app/src/pages/inside/common/statusDropdown/statusDropdown.jsx
```diff
@@ -120,17 +132,25 @@ export class StatusDropdown extends Component {
     const { intl } = this.props;
 
     const STATUS_TYPES = status === IN_PROGRESS ? [] : [PASSED, FAILED, SKIPPED];
+    const isTestItemsPage = this.getIsExpectedCurrentPage(TEST_ITEM_PAGE);
 
     if (STATUS_TYPES.indexOf(status) < 0) STATUS_TYPES.push(status);
 
     return STATUS_TYPES.map((item) => ({
-      label: formatStatus(intl.formatMessage, item),
+      label: isTestItemsPage ? (
```
AmsterGet, line 140: Use more abstract and generic condition here, f.e. `withIndicator`.
Or, may be it will be better, can we use just one status label component and manage its content via the prop?

app/src/pages/inside/common/statusDropdown/statusDropdown.jsx
```diff
@@ -139,10 +159,12 @@ export class StatusDropdown extends Component {
           onChange={this.updateItem}
           customClasses={{
             dropdown: cx('dropdown'),
-            selectBlock: cx('select-block'),
+            selectBlock: cx('select-block', { 'select-block-log': isLogPage }),
```
AmsterGet, line 162: Please rename to more generic

AmsterGet (changes requested):

app/src/pages/inside/logsPage/logItemInfo/logItemInfo.jsx
```diff
@@ -46,12 +47,15 @@ const messages = defineMessages({
...
-  retries: retriesSelector(state),
-}))
+@connect(
+  (state) => ({
+    logItem: activeLogSelector(state),
+    historyItems: historyItemsSelector(state),
+    retryItemId: activeRetryIdSelector(state),
+    retries: retriesSelector(state),
+  }),
+  { reloadLogPage: fetchLogPageData },
```
AmsterGet, line 57: Why are you reload the whole page? May be better to fetch just the history items?
This should reduce the time the user waits for the action to complete.
As an better alternative, we can update the necessary item locally via redux action.

AmsterGet (changes requested):
I think this separator is redundant.

image

AmsterGet (approved):

PR #2796, opened 2021-08-16
EPMRPP-64580 - prevent testCaseId field overlap

chivekrodis:


app/src/pages/inside/logsPage/logItemInfo/logItemInfoTabs/logItemDetails/logItemDetails.scss
```diff
@@ -143,4 +143,5 @@
   line-height: 1.2;
   color: $COLOR--charcoal-grey;
   padding-bottom: 10px;
+  word-break: break-all;
```
AmsterGet, line 146: You may also use 
```javascript
    word-break: break-word;
  }
```
to beautify wrapped content in some browsers.

AmsterGet (approved):

PR #2815, opened 2021-09-15
EPMRPP-65563: fix empty stacktrace section

oranje322:


chivekrodis (changes requested):
Please recheck your implementation for test without errors

app/src/pages/inside/common/stackTrace/stackTrace.jsx
```diff
@@ -104,6 +105,13 @@ export class StackTrace extends Component {
     this.fetchItems();
   }
 
+  componentDidUpdate(prevProps) {
+    const { items, logItem } = this.props;
+    if (this.isItemsExist() === 0 && prevProps.items !== items && logItem.status === FAILED) {
+      this.fetchItems();
+    }
```
chivekrodis, line 112: I believe we can attach an error log to any test item or rewrite the status of a test item. So maybe `status===failed` is not the best or universal solution. Please discuss with the lead whether we will support such a case or not.

AmsterGet: @oranje322 Not only failed items have error logs, it's true.

app/src/pages/inside/common/stackTrace/stackTrace.jsx
```diff
@@ -104,6 +104,13 @@ export class StackTrace extends Component {
     this.fetchItems();
   }
 
+  componentDidUpdate(prevProps) {
+    const { items, logItem } = this.props;
+    if (this.isItemsExist() === 0 && prevProps.items !== items && prevProps.logItem !== logItem) {
```
chivekrodis, line 109: Does it necessary check `this.isItemsExist() === 0 && prevProps.items !== items`? Looks like you should check only `logItem` or maybe better `logItem.id`

chivekrodis (changes requested):

chivekrodis (approved):

AmsterGet (approved):

PR #2817, opened 2021-09-16
EPMRPP-65343: composite attribute go out of the border

oranje322:


app/src/components/filterEntities/entitiesGroup/entitiesGroup.jsx
```diff
@@ -105,7 +106,7 @@ export class EntitiesGroup extends Component {
             <div key={id} className={cx('entity-item', { vertical })}>
               <EntityComponent
                 entityId={id}
-                smallSize={entitySmallSize}
+                smallSize={entity.id === ENTITY_ATTRIBUTE ? false : entitySmallSize}
```
chivekrodis, line 109: It looks not flexible. You bind `smallSize` to false forever. But you can provide this property from upper component which collect all filter property.

chivekrodis (changes requested):

app/src/components/filterEntities/entitiesGroup/entitiesGroup.jsx
oranje322: Why am I forever binding `smallSize` to false? In this case, smallSize will only be false for the composite attribute. And that what i want. For others inputs, `smallSize ` inherit from `entitySmallSize  `. And in this case will be true.

chivekrodis: Ok. my bad. I mean bind `false` for `ENTITY_ATTRIBUTE` whenever it used. Maybe it is ok, I just highlight this case.
@AmsterGet pls, check this solution

AmsterGet: `EntitiesGroup` looks like a generic component for entities, and it does not need to know which entities are used in it.
Accordingly, I propose to bind the `false` value for `ENTITY_ATTRIBUTE` inside `EntityInputConditionalAttributes` as we do not support small view for this component.

AmsterGet (approved):

PR #2829, opened 2021-09-22
EPMRPP-65959: added saml new field

oranje322:


app/src/components/integrations/integrationProviders/samlIntegration/samlFormFields/samlFormFields.jsx
```diff
@@ -173,6 +177,16 @@ export class SamlFormFields extends Component {
             <Input mobileDisabled />
           </FieldErrorHint>
         </IntegrationFormField>
+        <IntegrationFormField
+          name="callbackUrl"
+          disabled={disabled}
+          label={formatMessage(messages.callbackUrl)}
+          lineAlign={lineAlign}
+        >
+          <FieldErrorHint>
```
AmsterGet, line 186: I think we don't need `FieldErrorHint`, since there is no validation here.

AmsterGet (changes requested):

AmsterGet (approved):

PR #2857, opened 2021-10-11
EPMRPP-53964: null shouldn't be displayed for launches tables

renkyoji:


AmsterGet:
WIll be fixed on BE side.

PR #2930, opened 2021-11-30
EPMRPP-68774 || Run Unique Error analysis from the Burger menu

Vadim73i:


app/src/pages/inside/uniqueErrorsPage/modals/uniqueErrorsAnalyzeModal.jsx
```diff
@@ -35,30 +46,72 @@ const cx = classNames.bind(styles);
...
+        this.props.showNotification({
+          message,
+          type: NOTIFICATION_TYPES.SUCCESS,
+        });
+        const analysing = launch.analysing;
+        const item = {
+          ...launch,
+          analysing: [...analysing, ANALYZER_TYPES.CLUSTER_ANALYSER],
+        };
+        this.props.updateLaunchLocallyAction(item);
```
AmsterGet, line 99: In case of running cluster analysing from unique errors page we don't have items on launches page.
So I think the different action should be used here in this case. This action can be replaced with a callback, it will be `this.props.updateLaunchLocallyAction` if we call this modal from launches page, or it can be smth else (may be parent launch update) if we use it from unique errors page. You can `dispatch` `fetchParentLaunchSuccessAction` with necessary payload.

AmsterGet (changes requested):

AmsterGet:
Also please remove the `[UI]` prefix from the commit message and PR name. It is just for simple Jira filtering and search, not the really part of task name.

image

chivekrodis (commented):
In case of analyzer is not running I meet the error and page is crush

image
image

chivekrodis (approved):

AmsterGet (approved):

PR #2941, opened 2021-12-03
EPMRPP-69011 || Add new custom dimensions for GA

oranje322:


chivekrodis (approved):

app/src/controllers/project/selectors.js
```diff
@@ -36,6 +37,8 @@ export const projectMembersSelector = (state) => projectInfoSelector(state).user
 
 export const projectCreationDateSelector = (state) => projectInfoSelector(state).creationDate || 0;
 
+export const projectIdSelector = (state) => projectInfoSelector(state).projectId;
```
AmsterGet, line 40: We already has `projectIdSelector` in `src/controllers/pages/selectors.js`. So please use another name here, f.e. `projectInfoIdSelector`.
`projectIdSelector` the right decision here, but now may be better to use another name to avoid collisions.
At all it is point to refactoring, because the existing `projectIdSelector` returns project name from URL.

AmsterGet (approved):

PR #2942, opened 2021-12-06
EPMRPP-69013 || Add new events for Launches

Vadim73i:


app/src/layouts/appLayout/appSidebar/appSidebar.jsx
```diff
@@ -100,7 +107,7 @@ export class AppSidebar extends Component {
         message: <FormattedMessage id={'Sidebar.dashboardsBtn'} defaultMessage={'Dashboards'} />,
       },
       {
-        onClick: () => this.onClickButton(SIDEBAR_EVENTS.CLICK_LAUNCHES_BTN),
+        onClick: () => this.onClickButton(SIDEBAR_EVENTS.clickLaunchesBtn(this.getFilterCount())),
```
chivekrodis, line 110: Why not just like `.clickLaunchesBtn(this.props.launchFilters.length)`?
Should we count unsaved filters?

app/src/components/main/analytics/events/launchesPageEvents.js
```diff
@@ -300,6 +300,26 @@ export const LAUNCHES_PAGE_EVENTS = {
...
+  refineBtnMore: (value) => ({
+    category: LAUNCHES_PAGE,
+    action: 'Select Parameter to Refine',
+    label: `${value}`,
+  }),
+  clickAttributes: (value) => ({
+    category: LAUNCHES_PAGE,
+    action: 'Click on icon Attributes',
+    label: `${value}`,
+  }),
```
chivekrodis, line 317: Please in case of using this func in many places like suite page, move it func into `testItemPages/actionEventsCreators` and use it like util.
Also according to the task not only launch and suite pages should be covered. Please add event on the test and step pages too.

chivekrodis (changes requested):

app/src/components/filterEntities/entitiesGroup/entitiesGroup.jsx
```diff
@@ -77,11 +77,14 @@ export class EntitiesGroup extends Component {
...
   toggleEntity = (entityId) => {
+    const { tracking, events, onRemove, onAdd } = this.props;
     const entity = this.getEntity(entityId);
     if (entity.active) {
-      this.props.onRemove(entityId);
+      onRemove(entityId);
     } else {
-      this.props.onAdd(entity);
+      onAdd(entity);
+      const title = entity ? entity.title : '';
```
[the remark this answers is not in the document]
Vadim73i: There is no need for additional nesting.
This makes the code more readable.

app/src/layouts/appLayout/appSidebar/appSidebar.jsx
```diff
@@ -100,7 +119,8 @@ export class AppSidebar extends Component {
         message: <FormattedMessage id={'Sidebar.dashboardsBtn'} defaultMessage={'Dashboards'} />,
       },
       {
-        onClick: () => this.onClickButton(SIDEBAR_EVENTS.CLICK_LAUNCHES_BTN),
+        onClick: () =>
+          this.onClickButton(SIDEBAR_EVENTS.clickLaunchesBtn(this.getFiltersStatistic())),
```
AmsterGet, line 123: I believe that tracking the number of user filters needs to be done elsewhere.
The launches page is responsible for displaying the launches and their filters, and it might be better to keep track of the number of filters when the launches page is mounted.

AmsterGet (changes requested):

app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -98,10 +109,47 @@ export class LaunchFiltersContainer extends Component {
...
 
+  state = {
+    isFilterTracked: false,
+  };
+
+  componentDidUpdate(prevProps) {
+    const { launchesFiltersReady, launchFilters } = this.props;
+
+    if (launchesFiltersReady) {
+      if (launchFilters !== prevProps.launchFilters || !this.state.isFilterTracked) {
```
chivekrodis, line 120: You shouldn't compare all launchFilters obj, because tracking will be invoke everytime filter condition is change, better check only filters length to avoid it.

chivekrodis (changes requested):

app/src/pages/inside/common/launchFiltersContainer/launchFiltersContainer.jsx
```diff
@@ -117,8 +117,16 @@ export class LaunchFiltersContainer extends Component {
...
-        this.trackFilters();
+      const launchFilterArrayStatistic = this.getFiltersCountStatistic(launchFilters);
+      const prevPropsLaunchFilterArrayStatistic = this.getFiltersCountStatistic(
+        prevProps.launchFilters,
+      );
+      if (
+        JSON.stringify(launchFilterArrayStatistic) !==
+          JSON.stringify(prevPropsLaunchFilterArrayStatistic) ||
+        !this.state.isFilterTracked
+      ) {
```
chivekrodis: Why not just?
```js
if (launchFilters.length !== prevProps.launchFilters.length || !this.state.isFilterTracked) {}
```
Previous solution looks much easier at the first glance

Vadim73i: Requirements changed. It is necessary to track saved and unsaved filters.

chivekrodis: I see. But in https://github.com/reportportal/service-ui/pull/2942/commits/14c3bc06b77b532cf1d298d2e2c925a094290c25 you track saved and unsaved filter like `0#0` what the difference? Also for now on every change add\remove\change conditions `this.getFiltersCountStatistic` is called

chivekrodis (approved):

AmsterGet (approved):

PR #2969, opened 2022-01-10
EPMRPP-73376 || Wrong styles for darkmode input dropdown

oranje322:


app/src/common/css/variables/newColors.scss
```diff
@@ -17,3 +17,10 @@
 /* stylelint-disable color-no-hex */
 
 $COLOR--darkmode-topaz-text: #3abcd0;
+$COLOR--darkmode-topaz-selected: #1dbdd6;
+$COLOR--darkmode-topaz-pressed: #9ee7f2;
```
chivekrodis, line 21: What is the reason for adding colors that are not currently used?

AmsterGet (approved):

app/src/common/css/variables/newColors.scss
AmsterGet: Decided to leave it as is, because these colors will be needed in the scope of another redesign task.

PR #2972, opened 2022-01-12
EPMRPP-73556 || Unique errors. Add event for 'Click on Icon Arrow to Expand Stack Trace Message on Modal "Test Item Details"'

Vadim73i:


app/src/pages/inside/stepPage/modals/testItemDetailsModal/testItemDetailsModal.jsx
```diff
@@ -295,9 +295,13 @@ export class TestItemDetailsModal extends Component {
     const {
       data: { item, eventsInfo },
     } = this.props;
+    const customEventsInfo = {
+      ...eventsInfo,
+      onOpenStackTraceEvent: () => eventsInfo.clickExpandEvent,
```
chivekrodis, line 300: looks like you can add this event directly from stepPage or uniqueErrorsPage with right naming.
Also please remove L153-L155 properties in this file, they looks unnecessary and `ContainerWithTabs` component soesn't use it

chivekrodis (changes requested):

chivekrodis (approved):

AmsterGet (approved):

PR #2978, opened 2022-01-18
EPMRPP-72590 || add ability filter clusters on the unique error page

chivekrodis:


app/src/controllers/uniqueErrors/sagas.js
```diff
@@ -55,18 +68,44 @@ function* fetchClusters(payload = {}) {
...
-      }),
-    ),
+  const supportedIntegration = integrations.find(
+    (integration) =>
+      integration &&
+      integration.integrationType &&
+      integration.integrationType.details &&
+      integration.integrationType.details.metadata &&
+      integration.integrationType.details.metadata.supportedFeatures &&
+      integration.integrationType.details.metadata.supportedFeatures.includes('newErrors'),
```
AmsterGet, line 81: As `newErrors` is a part of plugin, I think make sense to rename to smth like `uniqueErrors` or `uniqueErrorsClusters`.

app/src/controllers/uniqueErrors/sagas.js
```diff
@@ -55,18 +68,44 @@ function* fetchClusters(payload = {}) {
...
+    url = URLS.projectIntegrationByIdCommand(
+      project,
+      supportedIntegration.id,
+      COMMAND_GET_CLUSTERS,
+    );
+    requestParams.method = 'PUT';
+    const uniqueErrorsParams = yield select(pagePropertiesSelector, NAMESPACE);
+    requestParams.data = {
+      launchId,
+      mode: uniqueErrorsParams.mode || ALL,
```
AmsterGet, line 96: May be in core we don't need to know about this parameter and its possible values.
Can we just spread this params? Will the BE accept request params with empty `mode`?

app/src/controllers/testItem/utils.js
```diff
@@ -132,3 +133,8 @@ export const isItemOwner = (userId, item, launch) => {
 
   return true;
 };
+
+export const cleanUpQuery = (query) => {
+  const keys = Object.keys(query).filter((item) => !/^item+[0-9]/.test(item));
+  return omit(query, keys);
+};
```
AmsterGet, line 140: `cleanUpQuery` sounds too common, but you don't parametrized what to clean up.
Please add a corresponding parameter or rename it to be more explicit.

AmsterGet (changes requested):

app/src/controllers/uniqueErrors/sagas.js
```diff
@@ -93,7 +94,7 @@ function* fetchClusters(payload = {}) {
     const uniqueErrorsParams = yield select(pagePropertiesSelector, NAMESPACE);
     requestParams.data = {
       launchId,
-      mode: uniqueErrorsParams.mode || ALL,
+      mode: uniqueErrorsParams.mode,
       pageNumber: query[PAGE_KEY],
```
AmsterGet, line 98: Why not `...uniqueErrorsParams` according to the previous comment?

chivekrodis: `uniqueErrorsParams` can contain many parameters not only mode, should we pass all of them?

AmsterGet: Why not if they are part of unique errors? It is better than expose some properties that hidden in plugin.

AmsterGet (approved):

PR #2994, opened 2022-01-28
EPMRPP-74112 || fix wrong result for unique errors filter if all errors are new

chivekrodis:


app/src/controllers/uniqueErrors/sagas.js
```diff
@@ -73,7 +74,7 @@ function* getIntegration() {
   }
   const plugin = plugins.find((item) => item.name === supportedIntegration.name);
 
-  return plugin.enabled && supportedIntegration;
+  return yield plugin.enabled && supportedIntegration;
```
AmsterGet, line 77: You may just return value without `yield` here as the generator don't receive any arguments.
I suggest to execute this generator (line 105) via `call` saga function:
```javascript
const integration = yield call(getIntegration);
```

AmsterGet (changes requested):

AmsterGet (approved):

PR #3016, opened 2022-02-09
EPMRPP-72586 || Tooltips does not apply styles

oranje322:


chivekrodis (commented):
I guess we don't have a markup with this type of tooltip, but please make sure that result should look like on the screenshot below
image

AmsterGet (approved):

PR #3021, opened 2022-02-10
EPMRPP-74318 || jira cloud and azure dev ops plugin icons are missing

Vadim73i:


app/src/components/main/image/image.jsx
```diff
@@ -97,7 +97,7 @@ export class Image extends Component {
   };
 
   normalizeFileTypeForSvg = (file) => {
-    if (file.type === 'text/xml') {
+    if (file.type === 'text/xml' || file.type === 'text/xml;charset=utf-8') {
```
AmsterGet, line 100: Please make it more flexible to support different charsets, use `file.type.indexOf('text/xml')` or smth like that.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3053, opened 2022-03-31
EPMRPP-76241 || Plugins receiving before user login

tr1ble:


app/src/controllers/auth/sagas.js
```diff
@@ -111,7 +111,6 @@ function* loginSuccessHandler({ payload }) {
   const projectId = yield select(activeProjectSelector);
   yield put(fetchProjectAction(projectId));
   yield take(FETCH_PROJECT_SUCCESS);
-  yield put(fetchPluginsAction());
```
AmsterGet, line 114: Please don't remove the requests related to private plugins loading.

app/src/controllers/initialData/sagas.js
```diff
@@ -44,13 +44,13 @@ function* fetchInitialData() {
     const { payload: activeProject } = yield take(SET_ACTIVE_PROJECT);
     yield put(fetchProjectAction(activeProject));
     yield take(FETCH_PROJECT_SUCCESS);
-    yield put(fetchPluginsAction());
```
AmsterGet, line 47: Same here.

app/src/controllers/initialData/sagas.js
```diff
@@ -44,13 +44,13 @@ function* fetchInitialData() {
...
     yield put(fetchGlobalIntegrationsAction());
     yield put(authSuccessAction());
   } else {
     yield put(resetTokenAction());
```
AmsterGet, line 50: You need to fetch **public** plugins somewhere here when the https://jiraeu.epam.com/browse/EPMRPP-76240 will be completed by @chingiskhan-epam .

AmsterGet (changes requested):

app/src/controllers/plugins/uiExtensions/sagas.js
```diff
@@ -13,7 +13,9 @@ import {
 
 function* fetchExtensionsMetadata() {
   const plugins = yield select(pluginsSelector);
-  const uiExtensionPlugins = plugins.filter(
+  const publicPlugins = yield select(publicPluginsSelector);
+  const allPlugins = plugins.concat(publicPlugins);
+  const uiExtensionPlugins = allPlugins.filter(
```
AmsterGet, line 18: In this case `public` extensions will be fetched only after user login.
I suggest to call `fetchExtensionsMetadata` saga separately for public plugins as they are needed only for unauthorized user.
So the flow will look like this:
user enters app and there is no active user session -> fetch public plugins -> fetch public plugin extensions metadata.

app/src/controllers/plugins/uiExtensions/sagas.js
```diff
@@ -63,7 +65,9 @@ export function* fetchUiExtensions() {
   }
 
   const plugins = yield select(pluginsSelector);
-  const uiExtensionPlugins = plugins.filter(
+  const publicPlugins = yield select(publicPluginsSelector);
+  const allPlugins = plugins.concat(publicPlugins);
+  const uiExtensionPlugins = allPlugins.filter(
```
AmsterGet, line 70: This is also unnecessary, because there are a legacy extensions and we should avoid using them in new plugins.

AmsterGet (changes requested):

app/src/controllers/plugins/sagas.js
```diff
@@ -252,6 +262,13 @@ function* watchPluginChange() {
   );
 }
 
+function* watchPublicPluginChange() {
+  yield takeEvery(
+    [createFetchPredicate(PUBLIC_PLUGINS), FETCH_GLOBAL_INTEGRATIONS_SUCCESS],
```
AmsterGet, line 267: `FETCH_GLOBAL_INTEGRATIONS_SUCCESS` is not necessary here as it is deprecated mechanism.
Use ` yield takeEvery(createFetchPredicate(PUBLIC_PLUGINS), fetchExtensionsMetadata)`
Inside of `fetchExtensionsMetadata` you can select necessary plugins based on action type which will be presented as parameter.

app/src/controllers/initialData/sagas.js
```diff
@@ -49,6 +54,8 @@ function* fetchInitialData() {
...
+    yield put(resetPluginAction());
```
AmsterGet, line 57: Looks like the `resetPluginAction` is not necessary as you will define plugins to select based on action type.

AmsterGet (changes requested):

app/src/controllers/plugins/uiExtensions/sagas.js
```diff
@@ -2,18 +2,22 @@ import { select, call, all, put } from 'redux-saga/effects';
...
 } from './actions';
 
-function* fetchExtensionsMetadata() {
+export function* fetchExtensionsMetadata(...params) {
   const plugins = yield select(pluginsSelector);
-  const uiExtensionPlugins = plugins.filter(
+  const publicPlugins = yield select(publicPluginsSelector);
+  const currentPlugins =
+    params[0] && params[0].meta.namespace === PUBLIC_PLUGINS ? publicPlugins : plugins;
+  const uiExtensionPlugins = currentPlugins.filter(
```
AmsterGet, line 20: Please create a constant `isPublic` or smth like that to hold plugin type and use it to select necessary plugins.
Also use this value to choose right URL for metadata fetching in line 37 (pluginFile vs pluginPublicFile urls)

AmsterGet (changes requested):

app/src/controllers/plugins/uiExtensions/sagas.js
```diff
@@ -44,6 +54,7 @@ function* fetchExtensionsMetadata() {
     const metadataArray = results.map((metadata, index) => ({
       ...metadata,
       pluginName: uiExtensionPlugins[index].name,
+      pluginType: isPublic,
```
AmsterGet, line 57: Will it be `pluginType: true/false`?
Better `public: isPublic`.

app/src/controllers/plugins/uiExtensions/sagas.js
```diff
@@ -2,18 +2,22 @@ import { select, call, all, put } from 'redux-saga/effects';
...
   fetchExtensionsMetadataSuccessAction,
 } from './actions';
 
-function* fetchExtensionsMetadata() {
+export function* fetchExtensionsMetadata(...params) {
   const plugins = yield select(pluginsSelector);
-  const uiExtensionPlugins = plugins.filter(
+  const publicPlugins = yield select(publicPluginsSelector);
+  const isPublic = params[0] && params[0].meta.namespace === PUBLIC_PLUGINS;
+  const currentPlugins = isPublic ? publicPlugins : plugins;
```
AmsterGet, line 19: You select two plugin types before condition.
It can be optimised:
`currentPlugins = yield select(isPublic ? .. : ..);`

AmsterGet (changes requested):

AmsterGet (approved):

AmsterGet (approved):

PR #3054, opened 2022-04-04
EPMRPP-EPMRPP-76242 || Sign up extension point at login page

tr1ble:


AmsterGet (changes requested):

app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginBlock.jsx
```diff
@@ -48,29 +50,17 @@ export class LoginBlock extends PureComponent {
...
-    );
+    return extensions.find((extension) => extension.pluginName === SIGN_UP_PLUGIN_NAME);
   };
 
   render() {
     return (
       <PageBlockContainer header={messages.welcome} hint={messages.login}>
         <LoginForm />
-        {this.getLoginPageExtensions().signUp && this.getLoginPageExtensions().signUp.component}
+        {this.getSignUpExtension() && <ExtensionLoader extension={this.getSignUpExtension()} />}
```
AmsterGet, line 63: No need to hardcode this for sign up flow plugin as it is private feature.
You can just render the collection of extensions via `extensions.map`.

AmsterGet (changes requested):

app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginBlock.jsx
```diff
@@ -29,8 +34,30 @@ const messages = defineMessages({
...
+
+  render() {
+    const { extensions } = this.props;
+
+    return (
+      <PageBlockContainer header={messages.welcome} hint={messages.login}>
+        <LoginForm />
+        {extensions.map((extension) => (
+          <ExtensionLoader extension={extension} />
+        ))}
```
AmsterGet, line 59: According to the layout you need to render extensions before (under) `LoginForm` (see the designs).

AmsterGet: Also please note that connected components already pure (see [react-redux](https://react-redux.js.org/api/connect) docs).
As we use React 16+ we prefer to write new components as functional using React hooks, so you can use [use-selector](https://react-redux.js.org/api/hooks#useselector) to extract the extensions inside of functional component.

AmsterGet (approved):

PR #3065, opened 2022-04-13
EPMRPP-76490 || add project settings navigation

chivekrodis:


app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -14,39 +14,161 @@
...
+    }
+    Object.keys(extensionTabs).forEach((tab) => {
+      if (tabsConfig[tab]) {
+        tabsConfig[tab].component = extensionTabs[tab].component;
+
+        delete extensionTabs[tab];
+      }
+    });
+    return { ...tabsConfig, ...extensionTabs };
+  };
```
Vadim73i, line 148: Why not use an array with objects instead of an object with objects?
```
{
  id: DEMO_DATA,
  name: formatMessage(messages.demoData),
  link: createTabLink(DEMO_DATA),
  component: <DemoDataTab />,
  eventInfo: SETTINGS_PAGE_EVENTS.DEMO_DATA_TAB,
  mobileDisabled: true,
}
```
This will allow us to do something like
```
items.map(({ id, link, eventInfo, name }) => (
  <NavLink to={link} 
    key={id}
    onClick={createTrackingFunction(eventInfo)}
    className={cx('item')}
    activeClassName={cx('active')}
  >
    {name}
  </NavLink>
))
```
instead
```
Object.keys(items).map((item) => (
  <NavLink
    to={items[item].link}
    key={item}
    onClick={createTrackingFunction(items[item].eventInfo)}
    className={cx('item')}
    activeClassName={cx('active')}
  >
    {items[item].name}
  </NavLink>
))
```

chivekrodis: This is good advice. Unfortunately not suitable for this case. Because we manage the menu item and content with routing. It's better to be able to find an element directly with a single operation like `config[activeTab]` than to use an array and iterate over it `.find`

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -14,39 +14,161 @@
...
+
+  const content = useMemo(() => {
+    const itemsConfig = createItemsConfig();
+    if (!activeTab || !itemsConfig[activeTab]) {
+      const firstItemName = Object.keys(itemsConfig)[0];
+      dispatch(itemsConfig[firstItemName].link);
+      return null;
+    }
+    return itemsConfig[activeTab].component;
+  }, [activeTab, createItemsConfig, dispatch]);
```
Vadim73i, line 164: createItemsConfig in the dependencies. And createItemsConfig will be called every time activeTab  is changed.
Maybe make createItemsConfig useCallback?

Vadim73i (approved):

app/src/pages/inside/projectSettingsPageContainer/menu/menu.jsx
```diff
@@ -0,0 +1,57 @@
...
+import React from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames/bind';
+import { useTracking } from 'react-tracking';
+import { NavLink } from 'components/main/navLink';
+import styles from './menu.scss';
+
+const cx = classNames.bind(styles);
+
+export const Menu = ({ items }) => {
```
AmsterGet, line 26: May be `NavMenu` as it render links?

app/src/pages/inside/projectSettingsPageContainer/menu/menu.jsx
```diff
@@ -0,0 +1,57 @@
...
+  const { trackEvent } = useTracking();
+  const createTrackingFunction = (eventInfo) => (eventInfo ? () => trackEvent(eventInfo) : null);
+
+  return (
+    <ul className={cx('menu')}>
+      {items &&
+        Object.keys(items).map((item) => (
```
AmsterGet, line 33: It's not the true `item`, it's key. Rename to `itemKey` or just to `key`.

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -14,39 +14,164 @@
...
+    }
+    Object.keys(extensionTabs).forEach((tab) => {
+      if (tabsConfig[tab]) {
+        tabsConfig[tab].component = extensionTabs[tab].component;
+
+        delete extensionTabs[tab];
+      }
+    });
+    return { ...tabsConfig, ...extensionTabs };
+  }, [accountRole, createExtensionTabs, createTabLink, formatMessage, userRole]);
```
AmsterGet, line 151: Looks like `formatMessage` can be omitted here as its change should not lead to itemsConfig recalculation.

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -14,39 +14,164 @@
...
+
+  const navigation = useMemo(() => {
+    const itemsConfig = createItemsConfig();
+    const title = <FormattedMessage id="SettingsPage.title" defaultMessage="Project Settings" />;
+    return <Navigation items={itemsConfig} title={title} />;
+  }, [createItemsConfig]);
+
+  const content = useMemo(() => {
+    const itemsConfig = createItemsConfig();
```
AmsterGet, line 160: You call `createItemsConfig` two times without any parameters. Seems like this value can be memorized via `useMemo` instead of `useCallback`.

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -14,39 +14,164 @@
...
+    if (!activeTab || !itemsConfig[activeTab]) {
+      const firstItemName = Object.keys(itemsConfig)[0];
+      dispatch(itemsConfig[firstItemName].link);
+      return null;
+    }
+    return itemsConfig[activeTab].component;
+  }, [activeTab, createItemsConfig, dispatch]);
```
AmsterGet, line 167: Remove `dispatch` from dependencies, see comment about `formatMessage`.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3085, opened 2022-04-28
EPMRPP-76676 || Logs error fix

tr1ble:


app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -44,7 +49,9 @@ export const PluginIcon = ({ pluginData, className, ...rest }) => {
         return null;
       }
 
-      return URLS.projectIntegrationByIdCommand(projectId, integration.id, COMMAND_GET_FILE);
+      return isPublic
+        ? URLS.pluginPublicFile(integration.name, FILE_KEY)
+        : URLS.projectIntegrationByIdCommand(projectId, integration.id, COMMAND_GET_FILE);
     }
```
AmsterGet, line 55: We have several cases when the plugin will try to load its icon via API:
1) Plugin supports `commonCommand`
2) Plugin supports integration based command
3) Plugin public and have to load icon via `URLS.pluginPublicFile`
4) Plugin private and have to load icon via `URLS.pluginFile`

These 4 cases should be supported until the 1st and 2nd options will be removed as legacy.

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -25,12 +25,17 @@ import { filterIntegrationsByName, isPluginSupportsCommonCommand } from 'control
...
+const PUBLIC_ACCESS_TYPE = 'public';
+const FILE_KEY = 'plugin-icon.svg';
+
 export const PluginIcon = ({ pluginData, className, ...rest }) => {
   const { details, name } = pluginData;
   const isDynamicIconAvailable = details && details.binaryData && details.binaryData.icon;
   const projectId = useSelector(activeProjectSelector);
   const globalIntegrations = useSelector(globalIntegrationsSelector);
 
+  const isPublic = details && details.accessType === PUBLIC_ACCESS_TYPE;
```
AmsterGet, line 37: Please use `PUBLIC_PLUGIN_ACCESS_TYPE` from `controllers/plugins/constants.js`.

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -44,7 +49,9 @@ export const PluginIcon = ({ pluginData, className, ...rest }) => {
         return null;
       }
 
-      return URLS.projectIntegrationByIdCommand(projectId, integration.id, COMMAND_GET_FILE);
+      return isPublic
+        ? URLS.pluginPublicFile(integration.name, FILE_KEY)
```
AmsterGet, line 53: Please use file key from `details.binaryData.icon`.

AmsterGet (changes requested):

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
tr1ble: How to check if plugin supports integration based command ? Because now during checking looks like Sign Up supports it, but it causes error

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -24,13 +24,20 @@ import { globalIntegrationsSelector } from 'controllers/plugins/selectors';
...
 export const PluginIcon = ({ pluginData, className, ...rest }) => {
   const { details, name } = pluginData;
   const isDynamicIconAvailable = details && details.binaryData && details.binaryData.icon;
   const projectId = useSelector(activeProjectSelector);
   const globalIntegrations = useSelector(globalIntegrationsSelector);
 
+  const isPublic = details && details.accessType === PUBLIC_PLUGIN_ACCESS_TYPE;
+  const isCommand =
+    (isPluginSupportsCommonCommand(pluginData, COMMAND_GET_FILE) ||
+      filterIntegrationsByName(globalIntegrations, name)[0]) &&
```
AmsterGet, line 38: Seems like you can calculate it inside of `calculateIconSrc` function to avoid calling these functions twice.
I also suggest to update `calculateIconSrc` function to be able to extract from it not only url, but `requestParams` also.
It will help to avoid [this](https://github.com/reportportal/service-ui/pull/3085/files#diff-aeef3323931ea4f09685d937f8bca6535b7a8066eac3d3ebb4359afd06c8aeafR69) condition.

AmsterGet (changes requested):

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -24,39 +24,59 @@ import { globalIntegrationsSelector } from 'controllers/plugins/selectors';
...
   const { details, name } = pluginData;
   const isDynamicIconAvailable = details && details.binaryData && details.binaryData.icon;
   const projectId = useSelector(activeProjectSelector);
   const globalIntegrations = useSelector(globalIntegrationsSelector);
 
-  const calculateIconSrc = () => {
+  const isPublic = details && details.accessType === PUBLIC_PLUGIN_ACCESS_TYPE;
+
+  const calculateIconParams = () => {
+    const putRequestParams = { method: 'PUT', data: { fileKey: 'icon' } };
```
AmsterGet, line 38: Please rename to smth like `commandParams` as it is a specific to commands API.
And move it inside of if block.

app/src/components/integrations/elements/pluginIcon/pluginIcon.jsx
```diff
@@ -24,39 +24,59 @@ import { globalIntegrationsSelector } from 'controllers/plugins/selectors';
...
+    const getRequestParams = { method: 'GET' };
```
AmsterGet, line 39: 'GET' is default method for fetch API, so it can be omitted.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3095, opened 2022-05-11
EPMRPP-76498 || Checkbox component

oranje322:


app/src/componentLibrary/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,70 @@
...
+        'dark-view-checkbox': darkView,
+        disabled,
+      })}
+    />
+    {children && <span className={cx('children-container', { disabled })}>{children}</span>}
+  </label>
+);
+Checkbox.propTypes = {
+  children: PropTypes.node,
+  value: PropTypes.bool,
```
chivekrodis, line 53: Why you have props in validation and have no directly in component above? Have we can manage default value?

app/src/componentLibrary/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,70 @@
...
+  disabled: PropTypes.bool,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
+  className: PropTypes.string,
+  darkView: PropTypes.bool,
```
chivekrodis, line 59: As we implement component from scratch using boolean props which define styling looks not flexible. Maybe better use the same pattern like in the new Button component?

app/src/componentLibrary/checkbox/checkbox.scss
```diff
@@ -0,0 +1,136 @@
...
+}
+
+.input:checked:hover + .checkbox {
+  border-color: $COLOR--topaz-hover-2;
+  background-color: $COLOR--topaz-hover-2;
+}
+
+.input:focus + .checkbox {
+  border: 2px solid $COLOR--topaz-focused;
+}
```
chivekrodis, line 73: II have checked input - then I unchecked - leave cursor from the checkbox
And I see frame around checkbox
Does it expected behaviour?

chivekrodis (changes requested):

app/src/componentLibrary/checkbox/checkbox.jsx
```diff
@@ -0,0 +1,75 @@
...
+  variant,
+}) => (
+  // eslint-disable-next-line
+  <label
+    className={cx(`${variant}-container`, className, {
+      disabled,
+    })}
+    onFocus={onFocus}
+    onBlur={onBlur}
+    tabIndex="1"
```
chivekrodis, line 39: Why you overwrite standard behaviour for tabIndex if looks like you don't process it at all?

app/src/componentLibrary/checkbox/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+### Events:
+
+- **onFocus**
+- **onBlur**
+- **onChange**
+
+### Variants:
+
+Checkbox can be used in two variants: **default** for light background and **dark-view** for dark background.
```
chivekrodis, line 19: maybe better use light or topaz. And just dark without '-view'?

chivekrodis (approved):

app/src/componentLibrary/checkbox/checkbox.scss
```diff
@@ -0,0 +1,140 @@
...
+  margin-left: -24px;
+  box-sizing: border-box;
+  cursor: pointer;
+}
+
+.children-container {
+  cursor: pointer;
+}
+
+.input:checked + .light {
```
AmsterGet, line 50: It would be nice to organize styles in the same way as in [`dropdown`](https://github.com/reportportal/service-ui/pull/3090/files#diff-f89add43efa0f0968bcaf30c2c8056c71f91dd598291516fb656f38d1ddd46d2R138) implementation using SCSS mixins.
It will be easier to add new variants and reduce code duplication.

app/src/componentLibrary/checkbox/checkbox.scss
```diff
@@ -0,0 +1,130 @@
...
+
+.input:checked:active:not(:disabled) + .dark {
+  @include checked-state($COLOR--darkmode-topaz-focused, $COLOR--darkmode-topaz-focused);
+}
+
+.input:disabled + .dark {
+  border-color: $COLOR--darkmode-gray-100;
+}
+
+.input:checked:disabled + .dark {
```
AmsterGet, line 127: It would be better to organize the variant styles by one common parent selector, f.e.
```css
.dark {
   .input {
      &:checked {}
      
      &:disabled {}
      
      ...
   }
}
```

Also inspired by the dropdown component.
And will be easier to edit and add new variants with such implementation.

AmsterGet: @oranje322 I'll merge it, could you please update this in another PR?

PR #3125, opened 2022-06-07
EPMRPP-77650 || Empty state for notifications

tr1ble:


Vadim73i (approved):

chivekrodis (commented):
It should be a fully new component inside src/pages/inside/projectSettingsPageContainer/content
better option is check `patternAnalysis` component there
right now your solution is affect settings on an admin page

app/src/pages/inside/projectSettingsPageContainer/content/notifications/constants.js
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright 2019 EPAM Systems
```
chivekrodis: Please be more careful while copy/paste anything
update year everywhere to current one

chivekrodis (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/notifications/constants.js
tr1ble: Yes, sorry, I saw it right after the commit

AmsterGet (approved):

PR #3126, opened 2022-06-08
EPMRPP-77743 || extends createImportProps, adjust settings page container

chivekrodis:


app/src/controllers/plugins/uiExtensions/createImportProps.js
```diff
@@ -35,7 +35,7 @@ import { BigButton } from 'components/buttons/bigButton';
 import { NavigationTabs } from 'components/main/navigationTabs';
 import { NoCasesBlock } from 'components/main/noCasesBlock';
 import { ItemList } from 'components/main/itemList';
-import { ModalLayout, ModalField } from 'components/main/modal';
```
AmsterGet, line 38: The old `ModalLayout` component is also used in other plugins, so until a redesign for them, we need to leave this component exposed here.

AmsterGet (changes requested):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -156,9 +158,13 @@ export const ProjectSettingsPageContainer = () => {
   }, [accountRole, extensionsConfig, createTabLink, userRole]);
 
   const navigation = useMemo(() => {
-    const title = <FormattedMessage id="SettingsPage.title" defaultMessage="Project Settings" />;
-    return <Navigation items={config} title={title} />;
-  }, [config]);
+    if (rawPage) {
```
AmsterGet, line 161: May be call it prop as `subPage`? Then we can specify in the URL what kind of sub page we need to see.
So it will be f.e.:
`settings/pluginPage?subPage=pluginSubPageName`

AmsterGet (approved):

PR #3164, opened 2022-07-05
Fix attributes

chivekrodis:


app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -131,14 +142,19 @@ export const AttributeEditor = ({
...
+        error={state.errors.value}
+        touched={valueTouched}
+        setTouch={setTouchValue}
+      >
         <AttributeInput
           minLength={1}
           attributes={attributes}
           attributeComparator={byValueComparator}
-          getURI={getURIValue(state.key)}
+          getURI={getURIValue}
```
AmsterGet, line 157: Is it not needed anymore?
Previously attribute value search was done using `key` to find necessary values from pairs, f.e.:
```javascript
key: 'build',
value: '1.7.2' | '1.6.4' | '1.7.3' etc...
```

app/src/componentLibrary/attributeList/attributeList.jsx
```diff
@@ -38,6 +39,7 @@ export const AttributeList = ({
   customClass,
   showButton,
   editable,
+  attributesListClassname,
```
AmsterGet, line 42: As it is a `AttributeList` component, we expect that `className` or `customClass` will be applied for the component's top level element.
On the other hand custom classes for some inner stuff may have specific names, f.e. `attributeClassName` or smth like that.

app/src/componentLibrary/fieldText/fieldText.jsx
```diff
@@ -66,25 +66,27 @@ export const FieldText = ({
...
+            onChange={disabled ? null : onChange}
+            onFocus={disabled ? null : onFocus}
+            onBlur={disabled ? null : onBlur}
+            onKeyUp={disabled ? null : onKeyUp}
+            onKeyDown={disabled ? null : onKeyDown}
+          />
+          {placeholder && !value && (
+            <span className={cx(VARIANT, 'placeholder')}>
+              {placeholder}
+              {isRequired && <span className={cx(VARIANT, 'asterisk')} />}
```
AmsterGet, line 86: I don't any styles for `.asterisk` class.
Can it be omitted? Can we use just `::after` pseudo-element inside of `.placeholder`?

app/src/components/containers/attributeListContainer/attributeListContainer.jsx
```diff
@@ -22,7 +22,7 @@ import { EditableAttributeList } from 'componentLibrary/attributeList/editableAt
 export const AttributeListContainer = ({ value, keyURLCreator, valueURLCreator, ...rest }) => {
   const projectId = useSelector(activeProjectSelector);
   const getURIKey = keyURLCreator(projectId);
-  const getURIValue = (key) => keyURLCreator(projectId, key);
+  const getURIValue = valueURLCreator(projectId);
```
AmsterGet, line 25: Same question about the value search here.
Do we really need to remove that feature?

AmsterGet (changes requested):

AmsterGet (approved):

PR #3171, opened 2022-07-08
EPMRPP-77854 || implement system message component

RuslanLyubimov:


app/src/componentLibrary/systemMessage/systemMessage.jsx
```diff
@@ -0,0 +1,52 @@
...
+import styles from './systemMessage.scss';
+
+const cx = classNames.bind(styles);
+const MODE_INFORMATION = 'info';
+const MODE_WARNING = 'warning';
+const MODE_SYS_ERROR = 'error';
+
+export const SystemMessage = ({ header, caption, children, mode, contentByChildWidth }) => {
+  return (
+    <div className={cx('system-message', { contentByChildWidth })}>
```
chivekrodis, line 27: usually we are using kebab-case for classes instead of camelCase

app/src/componentLibrary/systemMessage/systemMessage.scss
```diff
@@ -0,0 +1,105 @@
...
+  );
+}
+
+.system-message {
+  min-height: 104px;
+  height: auto;
+  width: 100%;
+  background: $COLOR--bg-000;
+  border-radius: 8px;
+  box-shadow: 0px 1px 3px $COLOR--item-shadow;
```
chivekrodis, line 55: Please extract it to the corresponding scss file into variables. $COLOR--item-shadow will be removed from colors variables file as we are using new approach with extracting box-shadow into own vars
$BOX_SHADOW--item: 0px 1px 3px rgba(55, 67, 98, 0.1);

chivekrodis (changes requested):

chivekrodis (approved):

AmsterGet (approved):

PR #3174, opened 2022-07-11
EPMRPP-78178 || Returned design for admin page, EPMRPP-78128 

tr1ble:
EPMRPP-78178 || Returned old design for admin page, EPMRPP-78128 || Timer fix for iOS

### DIFF
--- a/app/src/pages/common/settingsPage/generalTab/generalTab.scss
+++ b/app/src/pages/common/settingsPage/generalTab/generalTab.scss
@@ -14,81 +14,45 @@
  * limitations under the License.
  */
 
-.field-container {
-  flex-direction: column;
-  align-items: flex-start;
-  margin-bottom: 24px;
+.general-tab {
+  position: relative;
+  width: 90%;
+  max-width: 1000px;
+  margin: 0 auto;
+  @media (max-width: $SCREEN_SM_MAX) {
+    width: 100%;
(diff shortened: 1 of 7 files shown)


chivekrodis (commented):
Your solution only works because the environment doesn't have the retention plugin installed
And we don't have an environment without this plugin installed

tr1ble:
> Your solution only works because the environment doesn't have the retention plugin installed And we don't have an environment without this plugin installed

I did the same as for Notifications or Pattern analysis, so I don't fully understand

tr1ble:
> > Your solution only works because the environment doesn't have the retention plugin installed And we don't have an environment without this plugin installed
> 
> I did the same as for Notifications or Pattern analysis, so I don't fully understand what I supposed to do then

chivekrodis:
> > Your solution only works because the environment doesn't have the retention plugin installed And we don't have an environment without this plugin installed
> 
> I did the same as for Notifications or Pattern analysis, so I don't fully understand what I supposed to do then

Please discuss the issue with the team leader as this is a known issue.
The problem arises because we have a retention plugin. You export the general settings to plugin props and it wraps them up to provide some restrictions. So when you have a plugin we can't manage two versions as the plugin exposes its own wrapped component everywhere as we need to control the constraints. However, it is possible without a plugin.

PR #3181, opened 2022-07-19
EPMRPP-78127 || Login page style fix

tr1ble:


app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.scss
```diff
@@ -53,10 +53,12 @@
 
   &.attempts-exceed {
     @media (max-width: $SCREEN_SM_MAX) {
-      margin-bottom: -64px;
+      position: absolute;
```
AmsterGet, line 56: Why we use `absolute` positioning here? What is the relative parent for this block?
I think we need to make a small refactoring here for login page to avoid usage of "absolutely" positioned elements where it is redundant.

AmsterGet (changes requested):

app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/loginForm.scss
AmsterGet: Decided that refactoring will done in the scope of the next task associated with the Login page shifting issues.

AmsterGet (approved):

PR #3184, opened 2022-07-21
EPMRPP-78289 || extends autocomplete

chivekrodis:


app/src/componentLibrary/autocompletes/singleAutocomplete/singleAutocomplete.jsx
```diff
@@ -116,15 +137,23 @@ export class SingleAutocomplete extends Component {
...
-                          onBlur();
+                        onBlur: (e) => {
+                          if (createWithoutConfirmation) {
+                            selectItem(inputValue);
+                          }
+                          onBlur(e);
+                          isOptionUnique &&
+                            isOptionUnique(
+                              inputValue ? !options.find((v) => v === inputValue) : null,
+                            );
```
AmsterGet, line 148: Why we need to do it here?
And may be better to rename it to `setIsOptionUnique` as it is an action, not just boolean.

chivekrodis: I have no possibility to compare options values and value somewhere outside

AmsterGet (approved):

PR #3197, opened 2022-07-25
EPMRPP-78578 || fix dropdown option height

chivekrodis:


app/src/componentLibrary/dropdown/dropdown.scss
```diff
@@ -19,7 +19,6 @@
   align-items: center;
   padding: 7px 12px;
   width: 100%;
-  height: 36px;
```
AmsterGet, line 22: Is it ok that now we have flexible height?

AmsterGet (approved):

PR #3203, opened 2022-07-29
EPMRPP-78541 || Remove polyfills for IE from 5.8 bundle

Bam6ycha:


app/src/common/polyfills.js
```diff
@@ -21,7 +21,6 @@ import 'array.findindex';
 import 'promise-polyfill/src/polyfill';
 import 'polyfill-array-includes';
```
chivekrodis, line 22: What about this imports and above?
Also we have it in the package.json file too

app/src/common/polyfills.js
```diff
@@ -21,7 +21,6 @@ import 'array.findindex';
...
 import areIntlLocalesSupported from 'intl-locales-supported';
-import objectValues from 'object.values';
```
chivekrodis, line 24: You must remove it from package.json file too

app/src/common/polyfills.js
```diff
@@ -33,92 +32,22 @@ if (window.NodeList && !NodeList.prototype.forEach) {
...
-
 if (!Object.fromEntries) {
-  Object.fromEntries = function fromEntries(entries){
+  Object.fromEntries = function fromEntries(entries) {
     var res = {};
-    for(var i = 0; i < entries.length; i++) res[entries[i][0]] = entries[i][1];
+    for (var i = 0; i < entries.length; i++) res[entries[i][0]] = entries[i][1];
     return res;
   };
 }
```
chivekrodis, line 41: Looks like it should be removed too

chivekrodis (changes requested):

chivekrodis (approved):

PR #3207, opened 2022-08-02
EPMRPP-77765 || extends create import props

chivekrodis:


app/src/pages/inside/projectSettingsPageContainer/content/elements/ruleList/ruleItem/ruleItem.jsx
```diff
@@ -60,6 +60,7 @@ export const RuleItem = ({ item, actions, onToggle, disabled, content }) => {
               ))}
             </span>
           )}
+          {customNode}
```
AmsterGet, line 63: Do we need additional `customNode` here?
Can we try to handle this case using `actions`?

app/src/pages/inside/projectSettingsPageContainer/content/elements/ruleList/ruleList.jsx
```diff
@@ -22,33 +22,45 @@ import styles from './ruleList.scss';
...
         <RuleItem
           key={item.id}
           item={item}
-          actions={actions}
+          actions={actionsConfig}
           onToggle={onToggle}
           disabled={disabled}
-          content={ruleItemContent && <Content item={item} />}
+          content={ruleItemContent ? <Content item={item} {...rest} /> : null}
+          customNode={customNode ? <CustomNode item={item} {...rest} /> : null}
```
AmsterGet, line 46: Can we avoid spreading a lot of props here?
I think it can be done on plugin side by wrapping components.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3212, opened 2022-08-04
EPMRPP-78275 || refactor attribute with control

chivekrodis:


app/src/componentLibrary/attributeList/editableAttributeList.jsx
```diff
@@ -24,19 +25,27 @@ const NEW_ATTRIBUTE = {
...
 }) => {
   const handleAddNew = () => {
     onChange([...attributes, NEW_ATTRIBUTE]);
+    onAddNewAttribute();
   };
+  useEffect(() => {
+    if (defaultOpen && !attributes.length) {
+      handleAddNew();
+    }
+  }, [disabled, attributes]);
```
AmsterGet, line 48: Should we call this effect in case of `defaultOpen` will change its value?

app/src/components/containers/attributeListContainer/attributeListContainer.jsx
```diff
@@ -14,31 +14,145 @@
...
+  };
+
+  const AttributeList = (props) => (
     <EditableAttributeList
-      attributes={value}
       getURIKey={getURIKey}
       getURIValue={getURIValue}
-      {...rest}
+      newAttrMessage={formatMessage(messages.addAttribute)}
+      attributesListClassname={attributesListClassname}
```
AmsterGet, line 96: I don't see this prop inside of `EditableAttributeList`.

app/src/components/containers/attributeListContainer/attributeListContainer.jsx
```diff
@@ -14,31 +14,145 @@
...
+          </Checkbox>
+          <span className={cx('description')}>{attributesCaption}</span>
+        </>
+      )}
+      {withoutProvider ? (
+        <AttributeList attributes={attributes} {...rest} />
+      ) : (
+        <FieldProvider name={name} {...rest}>
+          <AttributeList />
+        </FieldProvider>
```
AmsterGet, line 120: Why are you decided to put form field provider inside the attributes component?
In case of fields we are trying to separate their stuff from the form engine. If you just want to reuse it, create a top-level wrapper for attributes like `AttributeListFormField` or leave it as it was before to give the component's user control which form to use and how.

app/src/componentLibrary/attributeList/editableAttributeList.jsx
```diff
@@ -24,19 +25,27 @@ const NEW_ATTRIBUTE = {
 };
 
 export const EditableAttributeList = ({
-  attributes,
+  value: attributes,
```
AmsterGet, line 28: Why are you decided to rename it?
I think it will be more clear to keep the same naming for the same things here and in `AttributeList`.

AmsterGet (changes requested):
The common suggestion:
When the form engine is separated from fields it gives some benefits:
- Easy migration to the new form engine
- Low coupling
- Inversion of control (component's users will decide by themselves what they need to use)

AmsterGet (approved):

PR #3222, opened 2022-08-10
EPMRPP-78833 || Remove is_admin from event_parameter

Bam6ycha:
ep.is_admin was removed.
Fix errors witch you may see in browser`s console related with Universal GA.

### DIFF
--- a/app/src/components/main/analytics/AnalyticsWrapper.jsx
+++ b/app/src/components/main/analytics/AnalyticsWrapper.jsx
@@ -67,8 +67,7 @@ const getAppVersion = (buildVersion) =>
     });
     if (data.actionType && data.actionType === PAGE_VIEW) {
       ReactGA.pageview(data.page);
-    }
-    if (data.place) {
+    } else if (data.place) {
       const eventParameters = {
         instanceID: instanceId,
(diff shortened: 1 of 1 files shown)


app/src/components/main/analytics/AnalyticsWrapper.jsx
```diff
@@ -77,7 +76,6 @@ const getAppVersion = (buildVersion) =>
         pattern_analysis: normalizeDimensionValue(isPatternAnalyzerEnabled),
         project_id: projectId,
         timestamp: Date.now(),
-        is_admin: isAdmin,
```
chivekrodis, line 80: Also you should do it for the Google Analytics Universal

chivekrodis (changes requested):

app/src/components/main/analytics/AnalyticsWrapper.jsx
Bam6ycha: We did not send is_admin dimension for GA Universal.
In this case we just use isAdmin variable in order to determine whether we should send projectID or not.

chivekrodis (approved):

AmsterGet (approved):

PR #3254, opened 2022-09-06
EPMRPP-79419 || Fixes for automation attributes

tr1ble:


app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/modals/addEditDefectTypeModal/addEditDefectTypeModal.jsx
```diff
@@ -188,22 +188,15 @@ const AddEditDefectTypeModal = ({
...
         )}
         <div className={cx('input')}>
           <FieldProvider name={NAME_FIELD_KEY} type="text">
-            <FieldErrorHint provideHint={false}>
-              <FieldText
-                label={formatMessage(messages.name)}
-                defaultWidth={false}
-                dataAutomationId="longNameField"
-              />
+            <FieldErrorHint provideHint={false} dataAutomationId="longNameField">
```
chivekrodis, line 198: Why you don't use constants like NAME_FIELD_KEY?
each filed have a name and usually we use a constant for it
Also what the benefit of pass `dataAutomationId` to FieldErrorHint component and not directly to the FieldText?
In some places you passed `dataAutomationId` to the FieldElement, some to the FieldErrorHint please use one general way across the whole app

tr1ble: We need to pass this attribute to upper container. It can be borh FieldErrorHint or FieldElement so we pass it depending on the situation

tr1ble: Attribute should be set to the field container which not only contains the field itself, but also contains field description, error, title, etc. which is related to the field. Some tests will not only work with the field itself, but will also verify the name of the field, the hint under it or the validation error text. So, it will be simpler to find the field container and find all these elements inside of it, than set separate test attribute to each element of each field. It was discussed with Ilya

tr1ble: https://kb.epam.com/display/EPMRPP/Element+naming+for+UI+automation

Vadim73i (approved):

PR #3265, opened 2022-09-18
EPMRPP-79645-integrationSettings-bug-fix

RuslanLyubimov:


app/src/components/integrations/elements/bts/btsPropertiesForIssueForm/btsPropertiesForIssueForm.jsx
```diff
@@ -455,7 +455,15 @@ export class BtsPropertiesForIssueForm extends Component {
...
                   defaultOptionValueKey={this.defaultOptionValueKey}
-                />
+                  disabled={disabled}
+                >
+                  <div className={cx('show-hint-wrapper')}>
+                    <span className={cx('show-hint-text')}>
+                      {intl.formatMessage(messages.showFieldsHeader)}
+                    </span>
+                    <ShowWithTooltip />
+                  </div>
```
chivekrodis, line 465: Maybe better decide here should we pass children or not not directly in the component

app/src/components/integrations/elements/integrationSettings/integrationForm/integrationForm.jsx
```diff
@@ -131,59 +131,63 @@ export class IntegrationForm extends Component {
...
             ) : (
-              <div className={cx('control-buttons-block')}>
-                <div className={cx('button-container')}>
-                  <Button
-                    onClick={handleSubmit(this.submitIntegration)}
-                    disabled={shouldFieldsBeHidden}
-                  >
-                    {formatMessage(COMMON_LOCALE_KEYS.SUBMIT)}
+              <div className={cx('integration-form-fields')}>
+                {shouldFieldsBeHidden ? null : (
```
chivekrodis, line 143: Can we simplify this condition to !shouldFieldsBeHidden && <>?

chivekrodis (changes requested):

app/src/components/integrations/elements/integrationSettings/connectionSection/connectionSection.jsx
```diff
@@ -197,7 +197,11 @@ export class ConnectionSection extends Component {
         >
           <div className={cx('integration-info-block')}>
             <div className={cx('sub-header-block')}>
-              <div className={cx('general-info')}>
+              <div
+                className={cx('general-info', {
+                  blocked: blocked && availableProjectIntegrations.length > 0,
```
Bam6ycha, line 202: I think that in these case we can avoid of using availableProjectIntegrations.length > 0. 
But it is not mandatory to fix.

app/src/components/integrations/elements/integrationSettings/connectionSection/connectionSection.jsx
```diff
@@ -206,6 +210,7 @@ export class ConnectionSection extends Component {
               <div
                 className={cx('connection-block', {
                   'connection-block-failed': !connected,
+                  'connection-block-disabled': blocked && availableProjectIntegrations.length > 0,
                 })}
```
Bam6ycha, line 214: The same thing.
It is important to use length>0 only when we use conditional rendering.
It is not mandatory to fix.

RuslanLyubimov: without this, styling will be improper

app/src/components/integrations/elements/integrationSettings/connectionSection/connectionSection.jsx
RuslanLyubimov: without this, styling will be improper

Bam6ycha: Okay, may be you are right.

chivekrodis (approved):

AmsterGet (approved):

PR #3270, opened 2022-09-20
EPMRPP-79257 || not loaded Unique errors tab data

RuslanLyubimov:


AmsterGet:
Will be fixed in the scope of another task.

PR #3273, opened 2022-09-23
EPMRPP-72304 || Update permissions

Bam6ycha:
Update permissionMap, permissions. Now 'MEMBER' has the same permissions as 'PROJECT_MANAGER'. Delete all another roles.


app/src/common/utils/isEmptyObject.js
```diff
@@ -15,5 +15,8 @@
  */
 
 export function isEmptyObject(obj) {
+  if (typeof obj !== 'object' && obj != null) {
+    throw Error(`argument must be typeOf object`);
+  }
   return Object.keys(obj).length === 0;
```
chivekrodis, line 21: Can you explain this changes?
Also I see this PR https://github.com/reportportal/service-ui/pull/3269
it's in develop branch now but obviously will be in 5.9 too
should we write tests here now?

Bam6ycha: I write tests in this case, because pipeline is failed in case of test coverage less then 60%.
And I chanced this function, because it has incorrect logic  according to it name.
If we pass primitive, for example 1, it returns false, but it is not even an object.

chivekrodis: Maybe better cherry-pick commit to prevent conflicts in the future
Also function called as `isEmptyObject` why you try put in something else except object?
if you have problem with a data maybe better resolve problem with them
@AmsterGet pls check

app/src/components/integrations/containers/integrationInfoContainer/instancesSection/instancesSection.jsx
```diff
@@ -114,7 +114,7 @@ const messages = defineMessages({
 
 @connect(
   (state) => ({
-    projectId: projectIdSelector(state),
+    projectKey: urlProjectKeySelector(state),
```
AmsterGet, line 117: `projectId` is still present in prop types and is used inside the component.

app/src/pages/common/membersPage/modals/inviteUserModal/inviteUserModal.jsx
```diff
@@ -86,13 +86,13 @@ const inviteFormSelector = formValueSelector('inviteUserForm');
...
       ? inviteFormSelector(state, 'project')
-      : projectIdSelector(state),
+      : urlProjectKeySelector(state),
     selectedUser: inviteFormSelector(state, 'user'),
     isAdmin: isAdminSelector(state),
     projectKey: projectKeySelector(state),
     initialValues: {
       role: DEFAULT_PROJECT_ROLE,
-      project: projectIdSelector(state),
+      project: urlProjectKeySelector(state),
```
AmsterGet, line 95: Please leave a comment here to explain why we use two different project key selectors in this case.

AmsterGet (changes requested):

app/src/common/utils/isEmptyObject.js
AmsterGet: @Bam6ycha @chivekrodis 
Hope we will use this utility correctly :)
Anyway the Dima's solution more robust and it will indicate wrong behavior in case of unexpected type (f.e. if it is passed through long components chain, backend, etc).
I suggest to merge it as is and provide another PR with changes related to this utility to develop branch.
It's time to plan versions sync activities.

app/src/pages/common/membersPage/modals/inviteUserModal/inviteUserModal.jsx
Bam6ycha: You are right. It is unnecessary to use two selectors. I will remove projectKeySelector

app/src/components/integrations/containers/integrationInfoContainer/instancesSection/instancesSection.jsx
Bam6ycha: Ah, I see.
Neither one nor the second is used in the component. So I can just remove them.

chivekrodis (approved):

AmsterGet (approved):

PR #3285, opened 2022-10-03
EPMRPP-78997 || No defect filter fix

tr1ble:


app/src/pages/inside/common/filterEntitiesGroups/launchLevelEntities.jsx
```diff
@@ -332,24 +333,29 @@ export class LaunchLevelEntities extends Component {
...
-          placeholder: intl.formatMessage(messages.STATS_PLACEHOLDER),
-        },
-        removable: true,
-        active: visibleFilters.includes(option.value),
-      };
-    });
+    return getGroupedDefectTypesOptions(defectTypes, intl.formatMessage)
+      .filter(
+        (option) =>
+          option.groupId !== NO_DEFECT.toUpperCase() && option.groupRef !== NO_DEFECT.toUpperCase(),
```
AmsterGet, line 339: Why `&&`? It is an impossible case, because you'll get `undefined` for the first/second operand every time.
I guess it should be written via `||`.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3287, opened 2022-10-04
EPMRPP-80065 || Filter names fixes

tr1ble:


app/src/pages/inside/common/utils.js
```diff
@@ -157,8 +157,9 @@ export const getGroupedDefectTypesOptions = (
     });
     defectTypesOptions = defectTypesOptions.concat(
       defectTypeGroup.map((defectType) => {
-        const label = defectTypesLocalization[defectType.typeRef.toLowerCase()]
-          ? formatMessage(defectTypesLocalization[defectType.typeRef.toLowerCase()])
+        const defectTypeLabelId = `Defect_Type_${defectType.locator.toUpperCase()}`;
+        const label = defectTypesLocalization[defectTypeLabelId]
+          ? formatMessage(defectTypesLocalization[defectTypeLabelId])
```
Vadim73i, line 162: Why is this all, if you can just 
label: defectType.longName,

Vadim73i (changes requested):

app/src/pages/inside/common/utils.js
tr1ble: It's done for default defect types translation

Vadim73i (approved):

app/src/pages/inside/common/utils.js
```diff
@@ -169,6 +172,9 @@ export const getGroupedDefectTypesOptions = (
           color: defectType.color,
           locator: defectType.locator,
           meta: {
+            groupRefLabel: defectTypesLocalization[defectType.typeRef.toLowerCase()]
+              ? formatMessage(defectTypesLocalization[defectType.typeRef.toLowerCase()])
+              : defectType.longName,
             longName: defectType.longName,
             shortName: defectType.shortName,
```
AmsterGet, line 179: Seems like the `shortName` used together with `groupRefLabel` every time.
Can we join these two properties into one?

AmsterGet (changes requested):

app/src/pages/inside/common/utils.js
tr1ble: It used together here, but FE in QA we don't need `groupRefLabel`

AmsterGet (approved):

PR #3293, opened 2022-10-07
EPMRPP-79388 || Autotest attributes

tr1ble:


app/src/components/main/notification/notificationList/notificationListItem/notificationItem.jsx
```diff
@@ -178,7 +178,7 @@ export class NotificationItem extends PureComponent {
     }
     return (
       <div key={message} onClick={this.messageClick}>
-        <div className={cx('message-container', type)}>
+        <div className={cx('message-container', type)} data-automation-id="listItem">
```
AmsterGet, line 181: Why not for a top-level `div`?
Is it necessary to mark these component? May be it will be enough to use container attribute for automation purposes in this case. Please clarify with task creator.

AmsterGet: I don't see the difficulty in choosing this tag if the attribute is one level higher.
But okay, let's leave it as is.

AmsterGet (approved):

PR #3297, opened 2022-10-11
EPMRPP-80067 || Prevent page jumping

Bam6ycha:


app/src/components/main/grid/gridBody/nestedGridRow/nestedGridRow.jsx
```diff
@@ -108,10 +108,38 @@ export const NestedGridRow = ({
   const showPreviousButton = isLoadPreviousButtonVisible(page);
   const showMoreButton = isLoadMoreButtonVisible(page);
   const showLoadCurrentStepButton = isLoadCurrentStepButtonVisible(page, data, allNestedStepsData);
+  const [firstRowId, setFirstRow] = React.useState(null);
+  const [direction, setDirection] = React.useState(null);
```
RuslanLyubimov, line 112: Maybe, destructure useState and useRef from react library?

Bam6ycha: In this case, I think, that unnecessary to increase the size of imported module.
We have everything in React namespace.

app/src/components/main/grid/gridBody/nestedGridRow/nestedGridRow.jsx
```diff
@@ -108,10 +108,38 @@ export const NestedGridRow = ({
   const showPreviousButton = isLoadPreviousButtonVisible(page);
   const showMoreButton = isLoadMoreButtonVisible(page);
   const showLoadCurrentStepButton = isLoadCurrentStepButtonVisible(page, data, allNestedStepsData);
+  const [firstRowId, setFirstRow] = React.useState(null);
```
chivekrodis, line 111: Same here

app/src/components/main/grid/gridBody/nestedGridRow/nestedGridRow.jsx
```diff
@@ -108,10 +108,38 @@ export const NestedGridRow = ({
...
+  const [direction, setDirection] = React.useState(null);
+  const itemIntoViewRef = React.useRef(null);
+
+  const calculateFirstRowIdOnLoadPreviousClick = (contentItems) => {
+    const isStep = (item) => 'hasContent' in item;
+    setFirstRow((contentItems.find((item) => !isStep(item)) || {}).id);
```
chivekrodis, line 117: What potential situation when we're don't find log and fallback to {}?

chivekrodis (changes requested):

app/src/components/main/grid/gridBody/nestedGridRow/nestedGridRow.jsx
Bam6ycha: For now I can not, imagine the situation, when we did not find id.
But this situations always accurse for me, when I developed, because was situation when we have nested step, and no logs inside, but have button(load prev).  I know, that situation like this, probably never accurse. 

As for me, I think that is better to have fallback, because we cannot cover all cases when we can did not find id.
But if this situation accurse we protect our self from app crashing.

chivekrodis (approved):

PR #3305, opened 2022-10-17
EPMRPP-79768 || Update creating project

Bam6ycha:
On add project we need to pass organizationId and projectKey. AssignedProjects now contains organizationSlug field.

### DIFF
--- a/app/src/controllers/administrate/projects/sagas.js
+++ b/app/src/controllers/administrate/projects/sagas.js
@@ -31,7 +31,6 @@ import { fetch, getStorageItem, setStorageItem } from 'common/utils';
 import { PROJECT_PAGE } from 'controllers/pages';
 import { confirmSaga, hideModalAction } from 'controllers/modal';
 import { PROJECT_MANAGER } from 'common/constants/projectRoles';
-import { projectOrganizationSlugSelector } from 'controllers/project';
 import {
   NAMESPACE,
   FETCH_PROJECTS,
@@ -72,21 +71,25 @@ function* watchSetViewMode() {
   yield takeEvery(START_SET_VIEW_MODE, setViewMode);
 }
 
-function* addProject({ payload: projectName }) {
+function* addProject({ payload: { projectName, organizationSlug } }) {
   try {
-    const { projectKey } = yield call(fetch, URLS.addProject(), {
+    const projectKey = projectName.trim().replace(/\s+/g, '_');
+    yield call(fetch, URLS.addProject(), {
       method: 'post',
       data: {
         entryType: PROJECT_TYPE_INTERNAL,
-        projectName,
+        projectKey,
+        organizationId: organizationSlug,
       },
     });
+
     const projectInfo = {
       projectKey,
       projectRole: PROJECT_MANAGER,
       entryType: PROJECT_TYPE_INTERNAL,
+      organizationSlug,
+      projectName,
     };
-    const organizationSlug = yield select(projectOrganizationSlugSelector);
     yield put(assignToProjectSuccessAction(projectInfo));
     yield put(hideModalAction());
     yield put(
--- a/app/src/controllers/administrate/projects/sagas.js
+++ b/app/src/controllers/administrate/projects/sagas.js
@@ -73,6 +73,10 @@ function* watchSetViewMode() {
 
 function* addProject({ payload: { projectName, organizationSlug } }) {
   try {
+    /*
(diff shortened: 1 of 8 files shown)


app/src/controllers/administrate/projects/sagas.js
```diff
@@ -72,21 +71,25 @@ function* watchSetViewMode() {
   yield takeEvery(START_SET_VIEW_MODE, setViewMode);
 }
 
-function* addProject({ payload: projectName }) {
+function* addProject({ payload: { projectName, organizationSlug } }) {
   try {
-    const { projectKey } = yield call(fetch, URLS.addProject(), {
+    const projectKey = projectName.trim().replace(/\s+/g, '_');
```
chivekrodis, line 76: Looks like we're creating `projectKey` here manually does it will be processed by BE side and override or this `projectKey` will be used permanently in the app. In case of this `projectKey` stay with us permanently should we care about creating new project using only API?

Bam6ycha: It is confuse me to, but we have such acceptance criteria  
screenshot_07

chivekrodis (approved):

RuslanLyubimov (approved):

app/src/controllers/administrate/projects/sagas.js
AmsterGet: I guess project key will be prefilled on UI when creating a new project, but user will be able to change it any time.

app/src/controllers/user/reducer.js
```diff
@@ -41,24 +41,26 @@ export const settingsReducer = (state = SETTINGS_INITIAL_STATE, { type, payload
 export const userAssignedProjectReducer = (state = {}, { type, payload }) => {
   switch (type) {
     case ASSIGN_TO_RROJECT_SUCCESS: {
-      const { projectName, projectRole, entryType } = payload;
+      const { projectKey, projectRole, entryType, organizationSlug, projectName } = payload;
```
AmsterGet, line 44: Do you provide `organizationSlug` when dispatching `ASSIGN_TO_RROJECT_SUCCESS` action?
I cannot find it.

app/src/controllers/administrate/projects/sagas.js
```diff
@@ -72,21 +71,25 @@ function* watchSetViewMode() {
...
+    yield call(fetch, URLS.addProject(), {
       method: 'post',
       data: {
         entryType: PROJECT_TYPE_INTERNAL,
-        projectName,
+        projectKey,
+        organizationId: organizationSlug,
```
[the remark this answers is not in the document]
Bam6ycha: Back-end expect organizationId field.

app/src/controllers/user/reducer.js
Bam6ycha: screenshot_08
I provide it from addProjectSaga.
But I see, I forgot provide it from another place.

app/src/controllers/administrate/projects/sagas.js
Bam6ycha: What value we will use to prefill projectKey field?
Do I need to implement this logic right now?

AmsterGet: Project name.
No, I think it will be in the scope of project creation modal redesign.

app/src/controllers/administrate/projects/sagas.js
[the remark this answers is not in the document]
AmsterGet: I see, but it a little bit confusing, may be we need to change naming on BE side?
Are they use this organisationId (organisationSlug) as id or slug to find proper organisation?
I think we can resolve this discussion, but please leave a comment with TODO to remember this case.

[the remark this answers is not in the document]
Bam6ycha: I think, you are absolutely right. I do not understand, why we provide organizationId. Because we don't have this value.
And how it looks like it all. Is it the same as organizationSlug or not? Or we planning to have two of those. This is unclear for me.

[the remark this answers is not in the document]
AmsterGet: I guess we will have both.
It is a question to discuss with BE team, so also please leave a comment in Jira ticket. Hope we will consider this question on the backlog refinement :)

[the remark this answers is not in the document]
Bam6ycha: Done!)

PR #3307, opened 2022-10-17
EPMRPP-80234 || Automation attributes fix

tr1ble:


AmsterGet (commented):
I'm not sure that this info should be a part of common component.
May be better to specify a class names here?

tr1ble:
> I'm not sure that this info should be a part of common component. May be better to specify a class names here?

I think it's OK, we have already add this info to common components, such ass checkbox or even breadcrumps

PR #3322, opened 2022-11-04
EPMRPP-80176 || Update passingRateSummary widget

Bam6ycha:


app/src/components/widgets/singleLevelWidgets/charts/passingRateSummary/passingRateSummary.jsx
```diff
@@ -43,11 +44,25 @@ export class PassingRateSummary extends Component {
...
+    const linkCreationParametersForFailed = includeSkipped
+      ? [FAILED, INTERRUPTED, SKIPPED]
+      : [FAILED, INTERRUPTED];
+
     const link = getStatisticsLink({
-      statuses: data.id === STATS_PASSED ? [PASSED] : [FAILED, INTERRUPTED, SKIPPED],
+      statuses: data.id === STATS_PASSED ? [PASSED] : linkCreationParametersForFailed,
       launchesLimit: widget.contentParameters.itemsCount,
-      types: null,
+      types: TEST_ITEM_PAGE,
```
AmsterGet, line 65: What is the reason of providing `TEST_ITEM_PAGE` here?

AmsterGet (changes requested):

app/src/components/widgets/singleLevelWidgets/charts/passingRateSummary/passingRateSummary.jsx
Bam6ycha: As I understand correctly, we need to display only method type 'Test'.
`NOTE: Only method type 'Test' should be displayed.` It is from acceptance criteria.

AmsterGet: So you can just remove `types` property at all to see the correct result.

AmsterGet: Function will use `'filter.in.type': LEVEL_STEP,` parameter

Bam6ycha: Ah, yes I see.
It is good remark!

app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/passingRateSummaryControls.jsx
```diff
@@ -35,6 +40,18 @@ const messages = defineMessages({
...
+  PassingRateOptionTotal: {
+    id: 'PassingRatePerLaunchControls.PassingRateOptionTotal',
+    defaultMessage: 'Total test cases (Passed, Failed, Skipped)',
+  },
+  PassingRateOptionExcludingSkipped: {
+    id: 'PassingRatePerLaunchControls.PassingRateExcludingSkipped',
+    defaultMessage: 'Total test cases excluding Skipped',
+  },
+  PassingRateFormGroupControlLabel: {
+    id: 'PassingRatePerLaunchControls.PassingRateFormGroupControlLabel',
```
AmsterGet, line 52: It will lead to duplicating messages in translated files after regeneration.
Could you move the common messages for these two widgets to `passingRateChart`?

AmsterGet (changes requested):

app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/passingRateSummaryControls.jsx
Bam6ycha: May be it was good idea, to keep this messages in RadioGroupControl component?

app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/passingRateSummaryControls.jsx
```diff
@@ -21,11 +21,17 @@ import { injectIntl, defineMessages } from 'react-intl';
...
 import { getWidgetModeOptions } from './utils/getWidgetModeOptions';
 import { ITEMS_INPUT_WIDTH } from './constants';
-import { TogglerControl, FiltersControl, InputControl } from './controls';
+import { TogglerControl, FiltersControl, InputControl, RadioGroupControl } from './controls';
 
 const DEFAULT_ITEMS_COUNT = '50';
+
+const FORM_GROUP_CONTROL = 'PassingRateFormGroupControlLabel';
+const EXCLUDING_SKIPPED = 'PassingRateOptionExcludingSkipped';
+const TOTAL_TEST_CASES = 'PassingRateOptionTotal';
```
AmsterGet, line 33: I see the same constants in passingRatePerLaunchControls..
Please keep them together with message and use to define message key as you use them to extract message.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3335, opened 2022-11-18
EPMRPP-80671 || Improve attribute options in Notifications

Bam6ycha:
Now when we have less than one attribute the options AND/OR are disabled. It becomes active only when at least two attr are added.

### DIFF
--- a/app/src/pages/common/settingsPage/notificationsTab/modals/addEditNotificationCaseModal/notificationCaseFormFields/notificationCaseFormFields.jsx
+++ b/app/src/pages/common/settingsPage/notificationsTab/modals/addEditNotificationCaseModal/notificationCaseFormFields/notificationCaseFormFields.jsx
@@ -65,10 +65,12 @@ export class NotificationCaseFormFields extends Component {
       getTrackingData: PropTypes.func,
     }).isRequired,
     attributesValue: PropTypes.array,
+    resetSection: PropTypes.func,
   };
   static defaultProps = {
     activeProject: '',
     attributesValue: [],
+    resetSection: () => {},
   };
 
   getDropdownInputConfig = () => {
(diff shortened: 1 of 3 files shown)


app/src/pages/common/settingsPage/notificationsTab/modals/addEditNotificationCaseModal/notificationCaseFormFields/notificationCaseFormFields.jsx
```diff
@@ -107,16 +109,25 @@ export class NotificationCaseFormFields extends Component {
   getAttributesConditionOptions = () => {
     const {
       intl: { formatMessage },
+      attributesValue,
+      resetSection,
     } = this.props;
+    const hasOneAttrOrLess = attributesValue.filter((attribute) => 'key' in attribute).length <= 1;
+
+    if (attributesValue.length === 1) {
+      resetSection([ATTRIBUTES_OPERATOR_FIELD_KEY]);
```
chivekrodis, line 118: This won't work if you already have a rule with 2 attributes and OR operator you can edit rule remove one attribute and according to the task we're expect operator should change to AND
`resetSection` reset to initial value

package-lock.json
```diff
@@ -0,0 +1,3 @@
+{
+  "lockfileVersion": 1
+}
```
chivekrodis, line 3: Please remove this file

chivekrodis (changes requested):

app/src/pages/common/settingsPage/notificationsTab/modals/addEditNotificationCaseModal/notificationCaseFormFields/notificationCaseFormFields.jsx
Bam6ycha: You are absolutely right!

chivekrodis (approved):

AmsterGet (approved):

PR #3340, opened 2022-11-23
EPMRPP-79532 || fix text in name field overlaps borders

KrasovskyAlexander:
[EPMRPP-79532](https://jiraeu.epam.com/browse/EPMRPP-79532) 
Fixed text in name field overlaps borders
---
Before: 
Screenshot_143
---
After: 
Screenshot_142
---


Bam6ycha (changes requested):
According to the acceptance criteria, tooltip needs to be shown when text has overflow ellipsis.
But I can not find the place where it was added.

KrasovskyAlexander:
@Bam6ycha tooltip will implementation in https://jiraeu.epam.com/browse/EPMRPP-80177. In this bug only need to fix the text

app/src/pages/common/settingsPage/generalTab/generalTab.jsx
```diff
@@ -302,7 +302,9 @@ export class GeneralTab extends Component {
         <form onSubmit={this.props.handleSubmit(this.onFormSubmit)}>
           <div>
             <div className={cx('fake-input-label')}>Name</div>
-            <div className={cx('fake-input')}>{this.props.projectId}</div>
+            <div className={cx('fake-input')}>
+              <div className={cx('fake-input-text')}>{this.props.projectId}</div>
+            </div>
```
chivekrodis, line 307: What the reason wrap text with one more `div` if you can do the same thing without creating nesting?
Please move properties for truncing to `.fake-input` class, remove width and manage width with a paddings
Also I suggest to communicate with design team as it strange that we have different left\right paddings looks like mistake

chivekrodis (changes requested):

app/src/pages/common/settingsPage/generalTab/generalTab.jsx
KrasovskyAlexander: @chivekrodis you correct!

chivekrodis (approved):

Bam6ycha (approved):

PR #3344, opened 2022-11-25
EPMRPP-78657 || Notification of an inexistent launch

Bam6ycha:
Implement warning message and highlighting inexistent launch names in edit and duplicate modals. Add endpoint which returns all existing launch names.

### DIFF
--- a/app/src/common/urls.js
+++ b/app/src/common/urls.js
@@ -110,7 +110,7 @@ export const URLS = {
   choiceSuggestedItems: (activeProject) => `${urlBase}${activeProject}/item/suggest/choice`,
   launchNameSearch: (activeProject) => (searchTerm = '') =>
     `${urlBase}${activeProject}/launch/names?filter.cnt.name=${searchTerm}`,
-  launchesAll: (activeProject) => `${urlBase}${activeProject}/launch/names`,
+  launchesExistingNames: (activeProject) => `${urlBase}${activeProject}/launch/names`,
   launchOwnersSearch: (activeProject) => (searchTerm = '') =>
     `${urlBase}${activeProject}/launch/owners?filter.cnt.user=${searchTerm}`,
   launches: (activeProject) => `${urlBase}${activeProject}/launch`,
--- a/app/src/common/urls.js
+++ b/app/src/common/urls.js
@@ -110,6 +110,7 @@ export const URLS = {
   choiceSuggestedItems: (activeProject) => `${urlBase}${activeProject}/item/suggest/choice`,
   launchNameSearch: (activeProject) => (searchTerm = '') =>
     `${urlBase}${activeProject}/launch/names?filter.cnt.name=${searchTerm}`,
+  launchesAll: (activeProject) => `${urlBase}${activeProject}/launch/names`,
   launchOwnersSearch: (activeProject) => (searchTerm = '') =>
     `${urlBase}${activeProject}/launch/owners?filter.cnt.user=${searchTerm}`,
   launches: (activeProject) => `${urlBase}${activeProject}/launch`,
--- a/app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
+++ b/app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
@@ -51,7 +51,7 @@ export const MultipleAutocomplete = ({
   parseInputValueFn,
   handleUnStoredItemCb,
   dataAutomationId,
-  storedItemsMap: isExistingLaunchNames,
+  existingItemsMap,
   ...props
 }) => {
   let updatePosition;
@@ -96,7 +96,7 @@ export const MultipleAutocomplete = ({
         selectedItems={value}
         onStateChange={onStateChange}
         options={options}
-        storedItemsMap={isExistingLaunchNames}
+        existingItemsMap={existingItemsMap}
         handleUnStoredItemCb={handleUnStoredItemCb}
       >
         {({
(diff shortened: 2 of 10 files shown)


app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
```diff
@@ -251,6 +253,9 @@ MultipleAutocomplete.propTypes = {
   parseInputValueFn: PropTypes.func,
   handleUnStoredItemCb: PropTypes.func,
   dataAutomationId: PropTypes.string,
+  existingLaunchNames: PropTypes.shape({
```
tr1ble, line 256: Should we use shape for 1 field ? And `existingLaunchNames` name is little bit confusing, because it sounds like it should be array of names, not object with 1 bool field. May be `isExistingLaunchNames' or it should be `string`, not `bool` ?

tr1ble (changes requested):

app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
Bam6ycha: Good point. But I think that we can use shape even for one field in object. And that object can contains a lot of fields. Every field consists of key:boolean.
Why I use this? Because already existing logic require us to use objects like this.
But isExistingLauncNames sounds better!

chivekrodis (changes requested):

app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleDownshift.jsx
```diff
@@ -26,9 +26,10 @@ export const MultipleDownshift = ({
   onChange,
   handleUnStoredItemCb,
   options,
+  isExistingLaunchNames,
   ...props
 }) => {
-  const [storedItemsMap, setStoredItems] = useState({});
+  const [storedItemsMap, setStoredItems] = useState(isExistingLaunchNames);
```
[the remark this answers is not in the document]
Bam6ycha: I thought about that, when implement this task.
But  I decided to did not change anything to not brake anything. I will try to simplify it.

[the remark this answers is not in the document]
chivekrodis: You're right. Now this is not necessary and may be better added to the technical debt. The recipient field also uses this component, so it's better to leave it as is
But I suggest renaming the isExistingLaunchNames property to AsyncMultipleAutocomplete and MultipleDropdown to something more generic since it's a common component

tr1ble (approved):

app/src/common/urls.js
```diff
@@ -110,6 +110,7 @@ export const URLS = {
   choiceSuggestedItems: (activeProject) => `${urlBase}${activeProject}/item/suggest/choice`,
   launchNameSearch: (activeProject) => (searchTerm = '') =>
     `${urlBase}${activeProject}/launch/names?filter.cnt.name=${searchTerm}`,
+  launchesAll: (activeProject) => `${urlBase}${activeProject}/launch/names`,
```
chivekrodis, line 113: Endpoint return array of launches names please rename property to something more connected to it

chivekrodis (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/notifications/modals/addEditNotificationModal/addEditNotificationModal.jsx
```diff
@@ -189,6 +190,8 @@ const AddEditNotificationModal = ({
   const { trackEvent } = useTracking();
   const dispatch = useDispatch();
   const [isEditorShown, setShowEditor] = React.useState(data.notification.attributes.length > 0);
+  const isExistingLaunchNames = useSelector(isExistingLaunchNamesSelector);
```
AmsterGet, line 193: Can we select it inside of `LaunchNamesContainer` directly?
As I see these data doesn't used anywhere else in current component,

app/src/pages/inside/projectSettingsPageContainer/content/notifications/modals/addEditNotificationModal/launchNamesContainer/launchNamesContainer.jsx
```diff
@@ -48,11 +48,20 @@ const messages = defineMessages({
...
+  existingItemsMap,
+  value,
+  ...rest
+}) => {
   const { formatMessage } = useIntl();
   const activeProject = useSelector(projectIdSelector);
   const [showMessage, setShowMessage] = useState(false);
 
+  useEffect(() => {
+    setShowMessage(!value.every((item) => !!existingItemsMap[item]));
```
AmsterGet, line 62: Can be simplified:
```javascript
setShowMessage(value.some((item) => !existingItemsMap[item]));
```

AmsterGet (changes requested):

AmsterGet:
Also please resolve conflicts.

app/src/pages/inside/projectSettingsPageContainer/content/notifications/modals/addEditNotificationModal/launchNamesContainer/launchNamesContainer.jsx
```diff
@@ -48,18 +49,14 @@ const messages = defineMessages({
...
+export const LaunchNamesContainer = ({ highlightUnStoredItem, value, ...rest }) => {
   const { formatMessage } = useIntl();
   const activeProject = useSelector(projectIdSelector);
   const [showMessage, setShowMessage] = useState(false);
+  const existingItemsMap = useSelector(isExistingLaunchNamesSelector);
 
   useEffect(() => {
-    setShowMessage(!value.every((item) => !!existingItemsMap[item]));
+    setShowMessage(value.some((item) => !existingItemsMap[item]));
   }, []);
```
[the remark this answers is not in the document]
Bam6ycha: Technically not. We get this data, when we open notification page in project settings.
Project notification page did not show until we did not get existing launch names.

PR #3347, opened 2022-11-28
EPMRPP-80606 || Attributes length increase (Github issue #1868)

tr1ble:


app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.jsx
```diff
@@ -21,16 +21,33 @@ import styles from './attributesBlock.scss';
...
     {attributes.filter(notSystemAttributePredicate).map((attribute) => (
       <div
         key={formatAttribute(attribute)}
-        className={cx('attribute', { cursor: isAttributeClickable })}
+        className={cx('attribute', {
+          cursor: isAttributeClickable,
+          'no-hover-effects': !hasHoverEffects,
+        })}
         onClick={isAttributeClickable ? () => onClickAttribute(attribute) : null}
+        title={formatAttribute(attribute, true)}
```
Bam6ycha, line 40: In my opinion it is better to create two different functions, than pass true or false as argument.
It looks unclear and person who reading that code forced to open function definition.

app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.scss
```diff
@@ -37,24 +38,41 @@
...
 
+  div {
+    max-width: 132px;
+    text-overflow: ellipsis;
+    overflow: hidden;
+    white-space: nowrap;
+  }
+
+  .value {
+    flex-basis: auto !important;
```
Bam6ycha, line 67: What the reason using `!imporatnt`?
Can we avoid it?

Bam6ycha (changes requested):

app/src/pages/common/settingsPage/notificationsTab/notificationsTab.jsx
```diff
@@ -70,7 +70,21 @@ const ruleFieldsConfig = {
     title: messages.attributesLabelWithOperator,
     getTitleVariables: ({ attributesOperator }) => ({ attributesOperator }),
     dataFormatter: (data) =>
-      data.reduce((acc, item) => `${acc.length ? `${acc}, ` : ''}${formatAttribute(item)}`, ''),
+      data.length
+        ? data.map((item) => (
```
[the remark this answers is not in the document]
tr1ble: It's a goot point, thanks

app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.scss
tr1ble: We need to change of current elements in this component. We can use `style` prop of element and pass flexBasis here, I'm not sure what aaproach is better :)

app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.jsx
tr1ble: I'm not sure than it's correct to create another same function that differs only with ' : ' instead of ':'

app/src/pages/inside/logsPage/logItemInfo/logItemInfoTabs/logItemDetails/logItemDetails.jsx
```diff
@@ -96,7 +96,7 @@ export class LogItemDetails extends Component {
             )}
             {logItem.attributes && !!logItem.attributes.length && (
               <span className={cx('attribute')}>
-                <AttributesBlock attributes={logItem.attributes} />
+                <AttributesBlock attributes={logItem.attributes} hasHoverEffects={false} />
```
Vadim73i, line 99: Rename to `noHoverEffects` and refactor other places.

app/src/components/main/attributeList/attributeList.jsx
```diff
@@ -82,6 +82,7 @@ export const AttributeList = track()(
     backgroundDark,
     tracking,
     eventsInfo,
+    maxCellWidth,
```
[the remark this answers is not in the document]
Vadim73i: I see this is used in a plugin.
Can we do this with a class instead of using inline styles?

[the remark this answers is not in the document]
tr1ble: Unfortunately no, this is especially inconvenient for the plugin, but I've changed string prop to number

app/src/components/main/attributeList/editableAttribute/attribute/attribute.jsx
```diff
@@ -45,8 +46,23 @@ export const Attribute = ({
         {Parser(CrossIcon)}
       </div>
     )}
-    <div className={cx('label', { [`background-dark`]: backgroundDark })}>
-      {formatAttribute(attribute)}
+    <div
+      className={cx('label', { [`background-dark`]: backgroundDark })}
+      title={formatAttribute(attribute, true)}
+    >
+      {attribute.key ? (
```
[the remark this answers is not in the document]
tr1ble: Because we don't need any style for value if it's an only one

app/src/pages/inside/common/itemInfo/attributesBlock/attributesBlock.jsx
tr1ble: I would like to avoid refactoring of whole places that use `formatAttribute` so I can make another method that differs only with divider or left like it's done now

Vadim73i (approved):

PR #3351, opened 2022-11-30
EPMRPP-78734 || combine duration tooltip 

KrasovskyAlexander:
[EPMRPP-78734](https://jiraeu.epam.com/browse/EPMRPP-78734) 
Combine the tooltip on the duration icon of force finished launch with the tooltip with duration descriptions.
Functions that used only for status tooltip in DurationBlock moved to durationTooltip component, Function that used both component move to utils. Also small change to withTooltip component to be able to make tooltip with a custom style 

Before: 
image

---
After: 
Screenshot_148
Screenshot_150

---


app/src/pages/inside/common/durationBlock/durationBlock.scss
```diff
@@ -59,3 +59,10 @@
     line-height: 18px;
   }
 }
+
+.tooltip-duration {
+  padding: 24px 32px;
+  background: rgba(34, 34, 34, 0.91);
+  box-shadow: 0px 1px 3px rgba(55, 67, 98, 0.1);
```
Bam6ycha, line 66: I think it is better to use here a rgba function. It enables you to use hex colors.
Please, use colors from our colors library.

app/src/pages/inside/common/infoLine/parentInfo/duration/durationTooltip/durationTooltip.jsx
```diff
@@ -15,16 +15,97 @@
...
+        <div>{formatMessage(messages.finishedDuration, { durationTime })}</div>
+        <div>{formatMessage(messages.finishedTime, { endTime })}</div>
+      </>
+    );
+  };
+
+  return (
+    <div className={cx('duration-tooltip')}>
+      <div className={cx('duration-tooltip-status')}>{getStatusTitle()}</div>
+      <FormattedMessage
```
Bam6ycha, line 105: What is the reason for using FormatMessage component here?
Can we use formatMessage function instead?

app/src/pages/inside/common/infoLine/parentInfo/duration/durationTooltip/durationTooltip.jsx
```diff
@@ -15,16 +15,97 @@
...
 
-export const DurationTooltip = () => (
-  <div className={cx('duration-tooltip')}>
-    <FormattedMessage
-      id="DurationTooltip.message"
-      defaultMessage="Duration is interval between first child starts and last child ends. But if child run in parallel, end time is a time of longest child, in this case duration will not be equal to child duration sum."
-    />
-  </div>
-);
+export const DurationTooltip = injectIntl(({ intl, status, timing, type }) => {
```
Bam6ycha, line 31: We usually use useIntl hook for that purpose.

Bam6ycha (changes requested):

Bam6ycha:
And we usually try to create a single commit for the whole task.
And after resolving comments, we name commits EPMRPP-78734 || Code review fixes - ${n}.

app/src/pages/inside/common/durationBlock/durationBlock.scss
KrasovskyAlexander: fixed ✅

app/src/pages/inside/common/infoLine/parentInfo/duration/durationTooltip/durationTooltip.jsx
KrasovskyAlexander: fixed ✅

app/src/pages/inside/common/infoLine/parentInfo/duration/durationTooltip/durationTooltip.jsx
KrasovskyAlexander: fixed ✅

app/src/pages/inside/common/infoLine/parentInfo/duration/durationTooltip/durationTooltip.jsx
```diff
@@ -102,10 +105,18 @@ export const DurationTooltip = injectIntl(({ intl, status, timing, type }) => {
...
-        defaultMessage="Duration is interval between first child starts and last child ends. But if child run in parallel, end time is a time of longest child, in this case duration will not be equal to child duration sum."
-      />
+      {formatMessage(messages.tooltipDescribe)}
     </div>
   );
-});
+};
+
+DurationTooltip.propTypes = {
+  intl: PropTypes.object.isRequired,
```
Bam6ycha, line 122: Do we pass intl through props?

Bam6ycha (changes requested):

app/src/components/main/tooltips/tooltip/withTooltip.jsx
```diff
@@ -89,7 +89,7 @@ export const withTooltip = ({ TooltipComponent, data = {} }) => (WrappedComponen
                     onMouseLeave={data.hoverable ? this.hideTooltip : null}
                   >
                     <div
-                      className={cx('tooltip-content')}
+                      className={cx('tooltip-content', data.customClassName)}
```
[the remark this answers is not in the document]
KrasovskyAlexander: Yes, because we didn’t have the opportunity to set custom styles for the tooltip, for example, in my case, we needed our own background-color and paddings. And it turns out like this if we do not pass the custom style to _withTooltip_, but hang the styles on the root div of the tooltip itself.
Screenshot_153
as you can see the padding and it shows the background color that is set without being able to change it.
I think this small change will improve the reusability of the component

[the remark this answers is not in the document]
KrasovskyAlexander: this props styled

Bam6ycha (approved):

Vadim73i (approved):

AmsterGet (approved):

PR #3358, opened 2022-12-13
EPMRPP-80922 || Ellipsis is displayed in wrong place for project integration with long name

Vadim73i:


app/src/componentLibrary/breadcrumbs/breadcrumbs.jsx
```diff
@@ -28,9 +28,23 @@ const MEATBALL_WIDTH = 36;
 const ARROW_WIDTH = 20;
 const MAX_SHOWN_DESCRIPTORS = 5;
 
-export const Breadcrumbs = ({ descriptors, titleTailNumChars, dataAutomationId }) => {
+export const Breadcrumbs = ({ descriptors, dataAutomationId }) => {
   const shownDescriptors = [...descriptors];
 
+  const titleTailNumChars = ((breadcrumbsCount) => {
+    if (breadcrumbsCount >= MAX_SHOWN_DESCRIPTORS) {
+      return 12;
```
Bam6ycha, line 36: Maybe it is better to create human-readable variables and use them instead of magic numbers.
Or we can create object and use computed properties.
What do u think?

Bam6ycha (changes requested):

Bam6ycha (approved):

PR #3361, opened 2022-12-16
EPMRPP-79899 || add max-width to container

KrasovskyAlexander:
[EPMRPP-79899](https://jiraeu.epam.com/browse/EPMRPP-79899) 

Before: 
image
---
After: 
Screenshot_158
Screenshot_159
Screenshot_160

---

### DIFF
--- a/app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/defectTypes.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/defectTypes.jsx
@@ -145,7 +145,7 @@ export const DefectTypes = ({ setHeaderTitleNode }) => {
   }, [defectTypes, canAddNewDefectType, isEditable]);
 
   return (
-    <>
+    <div style={{ maxWidth: '1200px', margin: '0 auto' }}>
       <TabDescription>{formatMessage(messages.description)}</TabDescription>
       <Divider />
       {(isInformationMessage || !canAddNewDefectType) && (
@@ -218,7 +218,7 @@ export const DefectTypes = ({ setHeaderTitleNode }) => {
           );
         })}
       </div>
-    </>
+    </div>
   );
 };
 DefectTypes.propTypes = {
--- a/app/src/pages/inside/projectSettingsPageContainer/content/notifications/notifications.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/notifications/notifications.jsx
@@ -224,7 +224,7 @@ export const Notifications = ({ setHeaderTitleNode }) => {
   return (
     <>
       {notifications.length ? (
-        <>
+        <div style={{ maxWidth: '1200px', margin: '0 auto' }}>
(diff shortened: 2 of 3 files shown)


app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/defectTypes.jsx
```diff
@@ -145,7 +145,7 @@ export const DefectTypes = ({ setHeaderTitleNode }) => {
   }, [defectTypes, canAddNewDefectType, isEditable]);
 
   return (
-    <>
+    <div style={{ maxWidth: '1200px', margin: '0 auto' }}>
```
Bam6ycha, line 148: Be as kind as to explain the reason to use style instead of className?

Bam6ycha (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/defectTypes.jsx
KrasovskyAlexander: I would have to create three identical classNames and import them for just 2 properties, there would be a lot more code
(3 lines -> ~20 lines), I thought that this approach was more appropriate here

Bam6ycha: I think that we need to follow recommendations from React official documentation :  https://reactjs.org/docs/dom-elements.html#style.
And if we want to reuse some styles we can use mixins: https://sass-lang.com/documentation/at-rules/mixin .
Also, I saw inline styles in this PR https://github.com/reportportal/service-ui/pull/3363/files.
U can create a mixin and reuse it.

AmsterGet (changes requested):
Looks like a workaround and similar to https://github.com/reportportal/service-ui/pull/3363.
I suggest to implement an additional top-level layout component that will responsible for adjusting content to support same behavior across different setting pages. Also if content should be centered on all setting pages (I guess the behavior must be the same if it's not the design issue), it make sense to write necessary css for the `src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx` component that wraps all setting pages.

app/src/pages/inside/projectSettingsPageContainer/content/defectTypes/defectTypes.jsx
AmsterGet: Please see the comment below.
And regarding inline style. In this case we can avoid it by creating a reusable component.
Mixins as an option here, but the new component or `projectSettingsPageContainer` styling is more preferable as it is not just styles itself, it is a part of layout. That means it should repeat the nodes structure together with styling.

Generally we try to avoid using inline styles in most cases, because it complicates styles overriding and such styles does not injected from the js to the `*.css` during build process. Another concern of using inline styles here is that it is not reusable, in case of requirements will change, we will need to update all setting pages instead of updating styles just in one place.

PR #3363, opened 2022-12-20
EPMRPP-79223 || add max-width and change onClick target integration item

KrasovskyAlexander:
[EPMRPP-79223](https://jiraeu.epam.com/browse/EPMRPP-79223) 
AND
[EPMRPP-79899](https://jiraeu.epam.com/browse/EPMRPP-79899) 

Before: 
image

---
After: 
Screenshot_161

---

### DIFF
--- a/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
@@ -83,12 +83,10 @@ export const Integrations = () => {
       {query.subPage && !!Object.keys(plugin).length ? (
         <IntegrationInfo goBackHandler={goBackHandler} data={plugin} integrationId={query.id} />
       ) : (
-        <div>
-          <IntegrationsList
-            availableIntegrations={availableGroupedPlugins}
-            onItemClick={onItemClick}
-          />
-        </div>
+        <IntegrationsList
+          availableIntegrations={availableGroupedPlugins}
+          onItemClick={onItemClick}
+        />
       )}
     </>
   );
--- a/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
@@ -83,7 +83,7 @@ export const Integrations = () => {
       {query.subPage && !!Object.keys(plugin).length ? (
         <IntegrationInfo goBackHandler={goBackHandler} data={plugin} integrationId={query.id} />
       ) : (
-        <div className={cx('integrations')}>
+        <div>
           <IntegrationsList
             availableIntegrations={availableGroupedPlugins}
             onItemClick={onItemClick}
--- a/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
@@ -79,7 +79,7 @@ export const Integrations = () => {
   };
 
   return (
-    <div style={{ maxWidth: '1200px', margin: '0 auto' }}>
+    <>
       {query.subPage && !!Object.keys(plugin).length ? (
         <IntegrationInfo goBackHandler={goBackHandler} data={plugin} integrationId={query.id} />
       ) : (
@@ -90,6 +90,6 @@ export const Integrations = () => {
           />
         </div>
       )}
-    </div>
+    </>
   );
 };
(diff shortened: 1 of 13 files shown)


Bam6ycha:
I think that we need to change the state of an arrow when we hover over the whole Integration item to the hover state.
If you look more carefully now when we hover over the arrow it changes its color.
I think that we need to implement changing the color of the arrow when we hover over the whole integration item.

app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
```diff
@@ -79,7 +79,7 @@ export const Integrations = () => {
   };
 
   return (
-    <>
+    <div style={{ maxWidth: '1200px', margin: '0 auto' }}>
```
Bam6ycha, line 82: The same as in https://github.com/reportportal/service-ui/pull/3361.
Follow the recommendations that @AmsterGet provided there.

Bam6ycha (changes requested):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.scss
```diff
@@ -24,4 +24,6 @@ $Z-INDEX-PAGE-CONTAINER-HEADER: 2;
 
 .content {
   padding: 33px 32px 60px;
+  max-width: 1200px;
```
Bam6ycha, line 27: This is a great solution, but it breaks styles in the Demo Data tab, Analyzer tab, General tab. Take a closer look at these tabs and compare the styles to the design system.
Also, note that there is an extra margin in the integrations. Because of them, integrations are not centered.
2022-12-26_14h57_47

Bam6ycha (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
```diff
@@ -83,7 +83,7 @@ export const Integrations = () => {
       {query.subPage && !!Object.keys(plugin).length ? (
         <IntegrationInfo goBackHandler={goBackHandler} data={plugin} integrationId={query.id} />
       ) : (
-        <div className={cx('integrations')}>
+        <div>
```
Bam6ycha, line 86: I think that div is unnecessary now.

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -199,3 +200,11 @@ export const ProjectSettingsPageContainer = () => {
     </SettingsLayout>
   );
 };
+
+export const CenteredContainerForPage = ({ children }) => (
+  <div className={cx('centred-content')}>{children}</div>
+);
+
```
Bam6ycha, line 207: Please, remove extra empty line.

Bam6ycha (changes requested):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
KrasovskyAlexander: fixed ✅

app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrations.jsx
KrasovskyAlexander: fixed ✅

Bam6ycha (approved):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -199,3 +200,10 @@ export const ProjectSettingsPageContainer = () => {
     </SettingsLayout>
   );
 };
+
+export const CenteredContainerForPage = ({ children }) => (
```
AmsterGet, line 204: Today it is responsible for centered content, tomorrow the behavior can be different.
Please rename to `SettingsPageContent`, also rename corresponding `className`.
And also usually we don't hold more than one component in a single file in case them don't use each other.
You can move this component together with styles to the `content` folder.

AmsterGet (changes requested):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
KrasovskyAlexander: fixed ✅

AmsterGet (approved):

PR #3371, opened 2023-01-03
EPMRPP-77501 || sorted integrations after add new item

KrasovskyAlexander:
[EPMRPP-77501](https://jiraeu.epam.com/browse/EPMRPP-77501)
sorted integrations by creationDate after add new item

### DIFF
--- a/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrationsList/integrationInfo/integrationCollection/integrationCollection.jsx
+++ b/app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrationsList/integrationInfo/integrationCollection/integrationCollection.jsx
@@ -23,10 +23,12 @@ import styles from './integrationCollection.scss';
 const cx = classNames.bind(styles);
 
 export const IntegrationCollection = ({ items, disabled, openIntegration }) => {
+  const sortedItems = items.sort((a, b) => b.creationDate - a.creationDate);
+
   return (
     <div className={cx('instances-list-wrapper')}>
       <ul className={cx('instances-list')}>
-        {items.map((item) => (
+        {sortedItems.map((item) => (
           <IntegrationCollectionItem
             key={item.id}
             item={item}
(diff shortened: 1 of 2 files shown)


app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrationsList/integrationInfo/integrationCollection/integrationCollection.jsx
```diff
@@ -23,10 +23,12 @@ import styles from './integrationCollection.scss';
 const cx = classNames.bind(styles);
 
 export const IntegrationCollection = ({ items, disabled, openIntegration }) => {
+  const sortedItems = items.sort((a, b) => b.creationDate - a.creationDate);
```
Bam6ycha, line 26: Do we really need to sort that array of integrations on every re-renders?

Could it be better to create an effect? Which runs on componentdidmount, and sort that array inside that effect. I suggest you use the useLayoutEffect hook here.

Bam6ycha (changes requested):

AmsterGet (changes requested):
@KrasovskyAlexander @Bam6ycha 
As I see in the previous task (`EPMRPP-67609`), the sorting was done on the BE side, so it will be redundant to sort integrations again on UI.
I would suggest to just add the new integration to the start of the list at `src/controllers/plugins/reducer.js` `addIntegration` function.

app/src/pages/inside/projectSettingsPageContainer/content/integrations/integrationsList/integrationInfo/integrationCollection/integrationCollection.jsx
```diff
@@ -23,12 +23,10 @@ import styles from './integrationCollection.scss';
 const cx = classNames.bind(styles);
 
 export const IntegrationCollection = ({ items, disabled, openIntegration }) => {
-  const sortedItems = items.sort((a, b) => b.creationDate - a.creationDate);
-
   return (
```
Bam6ycha, line 26: Optional:
I think here we can delete `return` statement.

AmsterGet: Done.

Bam6ycha (approved):

PR #3372, opened 2023-01-04
EPMRPP-81276 || Improve dropdown accessibility

Bam6ycha:


app/src/componentLibrary/dropdown/dropdown.jsx
```diff
@@ -77,30 +134,52 @@ export const Dropdown = ({
...
+        {...getItemProps({
+          item: option,
+          index,
+          selected: option.value === selectedItem.value,
+          variant,
+          option,
+          highlightHovered: highlightedIndex === index,
+          render: renderOption,
+          onChange: option.disabled ? null : () => handleChange(option),
+          onMouseEnter: setHighlightedIndex,
```
AmsterGet, line 179: You can provide `() => setHighlightedIndex(index)` right here, so the index prop no longer required inside of `DropdownOption`.

app/src/componentLibrary/dropdown/utils.js
```diff
@@ -0,0 +1,41 @@
...
+const calculateCurrentItemIndex = (index, itemsCount) =>
+  ((index % itemsCount) + itemsCount) % itemsCount;
+
+export const calculatePrevIndex = (index, options) => {
+  if (!options[index].disabled) {
+    return index;
+  }
+  const itemsCount = options.length;
+
+  return calculatePrevIndex(calculateCurrentItemIndex(index - 1, itemsCount), options);
```
AmsterGet, line 31: Are you calling this function recursively to skip disabled options?
Could you please replace it with a cycle?

app/src/componentLibrary/dropdown/utils.js
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export const calculateNextIndex = (index, options) => {
+  if (!options[index].disabled) {
+    return index;
+  }
+  const itemsCount = options.length;
+
+  return calculateNextIndex(calculateCurrentItemIndex(index + 1, itemsCount), options);
```
AmsterGet, line 40: Same here.

AmsterGet (changes requested):

app/src/componentLibrary/dropdown/utils.js
AmsterGet: Well, let's leave it as is.
But one more note: 
these two functions looks too similar, can we unit them together and reuse it via more specific ones?
f.e.
```
export const findNearestAvailableIndex = (index, options, step = 1) => {
 if (!options[index].disabled) {
    return index;
  }
 const itemsCount = options.length;

 return findNextAvailableIndex(calculateCurrentItemIndex(index + step, itemsCount), options);
}

export const calculateNextIndex = (index, options) => findNearestAvailableIndex(index, options);

export const calculatePrevIndex = (index, options) => findNearestAvailableIndex(index, options, -1);
```

AmsterGet (approved):

PR #3386, opened 2023-01-19
EPMRPP-81198 || fixed validation attribute value when you create launch filter

KrasovskyAlexander:
[EPMRPP-81198](https://jiraeu.epam.com/browse/EPMRPP-81198)
in other places of the application we have the Editable capability and we have props attribute.edited, but in this place we create new filter and this.props.attribute.edited === undefined
Screenshot_181

Before: 
https://jiraeu.epam.com/secure/attachment/2131767/81198.mp4 

---
After: 
Screenshot_180

---

### DIFF
--- a/app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
+++ b/app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
@@ -143,6 +143,8 @@ export class InputConditionalAttributes extends Component {
   render() {
     const { value, keyURLCreator, valueURLCreator, projectId } = this.props;
     const inputConditions = this.getConditions();
+    const edited = true;
+
     return (
       <div className={cx('input-conditional-attributes', { opened: this.state.opened })}>
         <div className={cx('attributes-block')}>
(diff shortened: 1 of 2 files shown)


app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
```diff
@@ -158,6 +160,7 @@ export class InputConditionalAttributes extends Component {
             projectId={projectId}
             onConfirm={this.onChangeAttributes}
             nakedView
+            attribute={{ edited }}
```
Bam6ycha, line 163: I am not sure, that it is a good solution.
May it would be better to provide a default value in defaultProps?
It would be nice, to describe the whole structure of attribute in propTypes.
And after that provide default values for that prop.

Bam6ycha (changes requested):

app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
KrasovskyAlexander: fixed ✓

Bam6ycha (approved):

AmsterGet (approved):

PR #3388, opened 2023-01-20
EPMRPP-81141 || addition to fix for EPMRPP-81141: form buttons are hidden too

KrasovskyAlexander:
[EPMRPP-81141](https://jiraeu.epam.com/browse/EPMRPP-81141)
after talk with Ashat and Helen Bobrova, we came to the conclusion that the form buttons should be hidden too in scope of this bug

Before: 
Screenshot_182
Screenshot_185

---
After: 
Screenshot_184

Screenshot_183

---

### DIFF
--- a/app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
+++ b/app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
@@ -54,7 +54,7 @@ export const IntegrationSettings = (props) => {
   const activeProject = useSelector(activeProjectSelector);
   const accountRole = useSelector(userAccountRoleSelector);
   const userRole = useSelector(activeProjectRoleSelector);
-  const isAbleToEditByRole = canUpdateSettings(accountRole, userRole);
+  const isEditable = canUpdateSettings(accountRole, userRole);
   const query = useSelector(querySelector);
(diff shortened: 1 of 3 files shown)


app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
```diff
@@ -47,6 +52,9 @@ export const IntegrationSettings = (props) => {
   const projectIntegrations = useSelector(namedProjectIntegrationsSelector);
   const projectId = useSelector(projectIdSelector);
   const activeProject = useSelector(activeProjectSelector);
+  const accountRole = useSelector(userAccountRoleSelector);
+  const userRole = useSelector(activeProjectRoleSelector);
+  const isAbleToEditByRole = canUpdateSettings(accountRole, userRole);
```
AmsterGet, line 57: I guess the behavior can be changed and the form will be disabled based not only on user role.
So me be name it f.e. as `isEditable` to be more general here.

KrasovskyAlexander: fixed ✓

AmsterGet (approved):

PR #3395, opened 2023-01-27
EPMRPP-79540 || Improve Checkbox accessibility

Bam6ycha:


AmsterGet:
Please remove the commits related to `EPMRPP-81163` from this PR - they were removed from develop recently.

app/src/components/containers/AttributeListFormField/attributeListFormField.jsx
```diff
@@ -59,8 +59,8 @@ export const AttributeListFormField = ({
       ? attributesNote || formatMessage(messages.attributesNote)
       : formatMessage(messages.attributesNotActive);
 
-  const attributeControlHandler = (e) => {
-    setShowEditor(e.target.checked);
+  const attributeControlHandler = (isChecked) => {
+    setShowEditor(isChecked);
```
AmsterGet, line 63: I'm looking for some best practices here and mostly see that component libraries try to follow the native DOM API as much as possible (see f.e. [material UI checkbox](https://mui.com/material-ui/api/checkbox), another example is a React Synthetic events), so I suppose in this case it's better to pass an `event` object to the callback.
Not only to be consistent with browser APIs, but this also can be useful for some event related properties (e.g. `stopPropagation`, `target`, etc.)

AmsterGet (changes requested):
I don't see any changes in packages, why the lock file was updated? Please update it only in case of adding or removing new packages or some other packages related activities (vulnerabilities fixes, packages upgrade, etc.).

app/src/componentLibrary/checkbox/checkbox.jsx
```diff
@@ -70,7 +98,7 @@ Checkbox.defaultProps = {
   children: '',
   value: false,
   disabled: false,
-  onChange: () => {},
+  // onChange: () => {},
```
AmsterGet, line 101: Seems like you accidentally commented it.

app/src/componentLibrary/checkbox/checkbox.jsx
```diff
@@ -39,16 +42,13 @@ export const Checkbox = ({
       event.preventDefault();
       return;
     }
+
     if (keyCode === enterKeyCode) {
       event.preventDefault();
-      onChange(!event.target.checked);
+      inputRef.current.click();
```
AmsterGet, line 48: Why not to call change handler with keyboard event `onChange(event)`?

AmsterGet (changes requested):

app/src/componentLibrary/checkbox/checkbox.jsx
AmsterGet: Got it!

AmsterGet (approved):

PR #3401, opened 2023-02-06
EPMRPP-81230 || Long attributes are not truncated

Bam6ycha:


app/src/components/widgets/multiLevelWidgets/mostPopularPatterns/mostPopularPatterns.jsx
```diff
@@ -160,8 +160,14 @@ export class MostPopularPatterns extends Component {
     return (
       <div className={cx('popular-patterns')}>
         <div className={cx('attribute-selector')}>
-          <div className={cx('attribute-label')}>{attributeKey}</div>
+          <div className={cx('attribute-key')}>
+            <span className={cx('description')}>KEY:</span>
```
AmsterGet, line 164: I suppose that these words need to be translated as we translate key and value terms in other cases (see the screenshot below). You can add it to the `COMMON_LOCALE_KEYS` as an option.

image

AmsterGet (changes requested):

AmsterGet (approved):

PR #3402, opened 2023-02-07
EPMRPP-81515 || Make decision modal. Broken styles in "Apply for" dropdown

Vadim73i:


app/src/pages/inside/stepPage/modals/makeDecisionModal/elements/inputDropdownRadio/inputDropdownRadio.jsx
```diff
@@ -55,8 +55,6 @@ export const InputDropdownRadio = ({
             onChange={onChangeOption}
             options={options}
             inputGroupClassName={cx('radio-input-group')}
-            mode="dark"
-            size="small"
```
AmsterGet, line 59: I guess it is better to provide these options once for the group of radio buttons.
So it is difficult to imagine case when some options in the group will be dark and some will be light.
You can update the `InputRadioGroup` component to support this.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3421, opened 2023-03-08
EPMRPP-82431 || GA4. Launches and test item pages. Part 1

AmsterGet:


app/src/components/main/analytics/events/common/ga4Utils.js
```diff
@@ -14,4 +14,8 @@
  * limitations under the License.
  */
 
-export { SIDEBAR_EVENTS } from './sideBarEvents';
+export const getBasicClickEventParameters = (category) => ({
+  action: 'click',
+  category,
+  place: '',
```
Bam6ycha, line 20: Maybe it is better to pass `place` as a parameter to this function?
And as temporary solution make that parameter optional `export const getBasicClickEventParameters = (category,place='') => ({
  action: 'click',
  category,
  place,
});`

And when we finally remove GA3, only this function will be rewrite in such way `export const getBasicClickEventParameters = (category,place) => ({
  action: 'click',
  category,
  ...(place && {place}),
});`

PR #3424, opened 2023-03-13
EPMRPP-81532 || Fix displaying colon to separate attributes

AmsterGet:


app/src/common/utils/attributeUtils.js
```diff
@@ -16,7 +16,7 @@
 
 export const getAttributeValue = ({ key, value } = {}) => value || key;
 
-export const formatAttribute = ({ key, value } = {}, divider = ':') => {
+export const attributeFormatter = ({ key, value } = {}, divider = ':') => {
```
Bam6ycha, line 19: As for me default value for the divider looks strange here.
Do we really need it?

app/src/common/utils/attributeUtils.js
```diff
@@ -26,8 +26,12 @@ export const formatAttribute = ({ key, value } = {}, divider = ':') => {
   return getAttributeValue({ key, value }) || '';
 };
 
+export const formatAttribute = (attribute) => {
```
Bam6ycha, line 29: And the name for this function does not look consistent.
We have `formatAttributeWithSpacedDivider`. May be better to rename it to `formatAttributeWithDefaultDivider`?

Bam6ycha (changes requested):

app/src/common/utils/attributeUtils.js
AmsterGet: Technically this function can be called without `divider`.
But I suppose we should not use it as is, only through `formatAttribute...` functions.
I'll remove it.

app/src/common/utils/attributeUtils.js
AmsterGet: Usually when we use smth without specifiers it means that we expect the default behavior.
So here we have `formatAttribute` - the default way to format attributes (and the frequently used) and the `formatAttributeWithSpacedDivider` which provides non-default behavior for some specific use cases.
Another reason why I decided to leave the default naming as is - to not change the API for plugins where this utility used as well.

Bam6ycha: I got it!

Bam6ycha (approved):

PR #3425, opened 2023-03-13
EPMRPP-82123 || Remove redundant validation in Filters

Bam6ycha:
Also change validation message for attribute filed.

### DIFF
--- a/app/localization/translated/be.json
+++ b/app/localization/translated/be.json
@@ -120,6 +120,7 @@
   "Attachments.noAttachmentsMessage": "Няма ўкладанняў для адлюстравання",
   "AttributeEditor.attributeKeyLengthHint": "Ключ павінен мець памер ад 1 да 512 сімвалаў",
   "AttributeEditor.attributeValueLengthHint": "Значэнне павінна мець памер ад 1 да 512 сімвалаў",
+  "AttributeEditor.attributeFilterValueLengthHint": "Значэнне павінна мець памер не больш за 512 сімвалаў",
   "AttributeEditor.keyLabel": "Ключ",
   "AttributeEditor.uniqueAttributeKeyHint": "Ключ атрыбуту павінен быць унікальным",
   "AttributeEditor.valueLabel": "Значэнне",
--- a/app/src/common/utils/validation/validate.js
+++ b/app/src/common/utils/validation/validate.js
@@ -94,6 +94,8 @@ export const attributeKey = (value) =>
 export const uniqueAttributeKey = (attributes) => (value) =>
   attributes.filter((attribute) => attribute === value).length <= 1;
 export const attributeValue = composeValidators([isNotEmpty, maxLength(512)]);
+export const attributesFilterValue = (value) =>
+  !value || composeValidators([isNotOnlySpaces, maxLength(512)])(value);
 export const attributesArray = (value) =>
   isEmpty(value) ||
   !value.length ||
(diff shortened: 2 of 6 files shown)


AmsterGet (changes requested):
The validation for empty `value` in attribute must be omitted for filters **only**.

Now it's broke the validation in other places where the `value` is mandatory, see the screenshot below.

image

app/src/components/main/attributeList/editableAttribute/attributeEditor/attributeEditor.jsx
```diff
@@ -52,6 +52,9 @@ const attributeValueValidator = composeBoundValidators([
   commonValidators.requiredField,
   bindMessageToValidator(validate.attributeValue, 'attributeValueLengthHint'),
 ]);
+const attributeFilterValueValidator = composeBoundValidators([
+  bindMessageToValidator(validate.attributesFilterValue, 'attributeFilterValueLengthHint'),
+]);
```
AmsterGet, line 57: The `composeBoundValidators` not needed if you use only one validator.
```javascript
const attributeFilterValueValidator = bindMessageToValidator(validate.attributesFilterValue, 'attributeFilterValueLengthHint');
```

AmsterGet: As `attributeFilterValueValidator` stored here in the commom `AttributeEditor` component and you use it as validator for non required value, so may be rename it f.e. to `nonRequiredAttibuteValueValidator`?
This will help to clearly reuse non required validation for attribute key in other places (not just in filters).

app/src/common/utils/validation/validate.js
```diff
@@ -94,6 +94,8 @@ export const attributeKey = (value) =>
 export const uniqueAttributeKey = (attributes) => (value) =>
   attributes.filter((attribute) => attribute === value).length <= 1;
 export const attributeValue = composeValidators([isNotEmpty, maxLength(512)]);
+export const attributesFilterValue = (value) =>
+  !value || composeValidators([isNotOnlySpaces, maxLength(512)])(value);
```
AmsterGet, line 98: Do we need to check `!value` here?
I see the `isNotOnlySpaces` and it has check for this.

image

app/localization/translated/be.json
```diff
@@ -120,6 +120,7 @@
   "Attachments.noAttachmentsMessage": "Няма ўкладанняў для адлюстравання",
   "AttributeEditor.attributeKeyLengthHint": "Ключ павінен мець памер ад 1 да 512 сімвалаў",
   "AttributeEditor.attributeValueLengthHint": "Значэнне павінна мець памер ад 1 да 512 сімвалаў",
+  "AttributeEditor.attributeFilterValueLengthHint": "Значэнне павінна мець памер не больш за 512 сімвалаў",
```
AmsterGet, line 123: Also I have a question regarding the hints.
They are look pretty similar. Can we leave just one message for all cases?
Point to discuss with BA.

AmsterGet (changes requested):

app/localization/translated/be.json
Bam6ycha: I've already discussed it with BA.
And the reason for using such a hint is the following.
If we have the hint 'Value should have the size from 1 to 512 symbols' it automatically makes that field required.
But in this case Value is not required and we only need to restrict users to enter a value longer than 512 symbols.

app/src/common/utils/validation/validate.js
Bam6ycha: Yes, we definitely need it here.
The reason for it is the following: when we clear the value of attr, it sets to `null`, and that breaks maxValue validator.
MaxValue validator does not check the value for `null` it only checks for `undefined`.
We can change the code of maxValue validator.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3429, opened 2023-03-16
EPMRPP-82439 || GA4 for launches & test item pages Part 3

evgenijkorob:


app/src/components/main/analytics/events/common/testItemPages/actionEventsCreators.js
```diff
@@ -146,16 +146,14 @@ export const getRefineFiltersPanelEvents = (category) => ({
   }),
 });
 
-export const getClickOnPlusMinusEvents = (category) => (state) => ({
-  category,
-  action: 'Click on Bread Crumb icon Plus/Minus',
-  label: state ? 'Minus' : 'Plus',
+export const getClickOnPlusMinusEventCreator = (category) => (expanded) => ({
```
Bam6ycha, line 149: Be as kind as to place these events under GA-4 category.

app/src/components/main/analytics/events/filtersPageEvents.js
```diff
@@ -36,11 +39,17 @@ export const getAddEditFilterModalEvents = (isEditMode) => {
       action: `Click on button Cancel in Modal ${modalType} Filter`,
       label: `Close Modal ${modalType} Filter`,
     },
-    clickOkBtn: {
-      category: FILTERS_PAGE,
-      action: `Click on button ${actionType} in Modal ${modalType} Filter`,
-      label: `${actionType} filter in Modal ${modalType} Filter`,
-    },
+    clickOkBtn: isEditMode
+      ? {
```
Bam6ycha, line 43: Why do we need to use a ternary operator here?
And FYI : if the event does not contain a place parameter, that event does not send to GA-4

app/src/components/main/analytics/events/common/testItemPages/actionEventsCreators.js
evgenijkorob: Done

app/src/components/main/analytics/events/filtersPageEvents.js
```diff
@@ -36,11 +39,17 @@ export const getAddEditFilterModalEvents = (isEditMode) => {
       action: `Click on button Cancel in Modal ${modalType} Filter`,
       label: `Close Modal ${modalType} Filter`,
     },
-    clickOkBtn: {
-      category: FILTERS_PAGE,
-      action: `Click on button ${actionType} in Modal ${modalType} Filter`,
-      label: `${actionType} filter in Modal ${modalType} Filter`,
-    },
+    clickOkBtn: isEditMode
```
Bam6ycha, line 42: Do we really need it here?
I mean left-hand expression of the ternary operator?

Bam6ycha (changes requested):

app/src/components/main/analytics/events/filtersPageEvents.js
evgenijkorob: We have GA4 analytics only for Add filter modal OK button, so I think yes, we need it

Bam6ycha: In case if isEditMode === true we do not send anything.
If we want to reuse clickOkBtn in edit modal f.e. We can rewrite it to : 
Screenshot 2023-03-17 at 3 59 09 PM

Bam6ycha: Sorry, not `isEditModa ? 'add' : 'edit'_modal` but `isEditModa ? 'add' : 'edit'_filter`

AmsterGet: According to the analytics dos I see that we decided to separate events for this modal between different categories (launches & filters).
So I suggest to parametrize function `getAddEditFilterModalEvents` with `category` parameter and move it upper to common place. In this case we will reuse the events for this modal across different pages.
Also may be provide events with necessary `category` and `editMode` to the modal inside of `showModalAction` function where it is called? As I see in the code it is clear where we need particular event type.

`LaunchFiltersToolbar`
image

image

`FiltersPage`
image

@evgenijkorob @Bam6ycha

AmsterGet: Sorry, it's not about your question.
As now we don't track the event for `add modal` I suppose the @evgenijkorob solution is applicable.
When these modal events are migrated to GA4, they will be adjusted accordingly.

AmsterGet (approved):

app/src/components/main/analytics/events/filtersPageEvents.js
AmsterGet: Decided to do it in the scope of the next task connected to filter modal events.

PR #3435, opened 2023-03-27
EPMRPP-82778 || Show 'Next Error' feature only for detailed log view

AmsterGet:
Also in this PR:
`isLogPageWithOutNestedSteps` usage removed from 'LogsGridToolbar' as it is an opposite to `isLogPageWithNestedSteps` which is already used inside of the component.

### DIFF
--- a/app/src/pages/inside/logsPage/logsGridToolbar/logsGridToolbar.jsx
+++ b/app/src/pages/inside/logsPage/logsGridToolbar/logsGridToolbar.jsx
@@ -28,7 +28,6 @@ import {
   getLogViewMode,
   setLogViewMode,
   DETAILED_LOG_VIEW,
-  isLogPageWithOutNestedSteps,
   isLogPageWithNestedSteps,
 } from 'controllers/log';
 import { InputSlider } from 'components/inputs/inputSlider';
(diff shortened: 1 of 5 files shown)


Vadim73i (approved):

app/src/pages/inside/logsPage/logsGridToolbar/logsGridToolbar.jsx
```diff
@@ -258,7 +255,7 @@ export class LogsGridToolbar extends Component {
               >
                 {Parser(MarkdownIcon)}
               </button>
-              {isLogView && (
+              {!isNestedStepsView && (
```
Bam6ycha, line 258: Maybe better to create a constant and name it isLogView?
As for me, it is clear in what view we need to display this or that layout.
What do you think?

Bam6ycha (changes requested):

app/src/pages/inside/logsPage/logsGridToolbar/logsGridToolbar.jsx
AmsterGet: The var `isLogView` was wrong in that meaning, because actually it was representing the `viewWithoutNestedSteps`.
On the other hand `isLogView` name is not relevant, because the `logsGridToolbar` always on the log view :)

AmsterGet: And according to the requirements, we need to hide some controls when there are no nested steps, as these controls valid only for them.

Bam6ycha: OMG!)
Okay)

Bam6ycha (approved):

PR #3436, opened 2023-03-28
EPMRPP-82651 || Autocomplete does not work in notification

Bam6ycha:
Two different modes have been added to distinguish between singleAutocomplete from multiAutocomplete modes.

### DIFF
--- a/app/src/components/inputs/autocompletes/withAsyncLoading.jsx
+++ b/app/src/components/inputs/autocompletes/withAsyncLoading.jsx
@@ -18,7 +18,12 @@ import { Component } from 'react';
 import PropTypes from 'prop-types';
 import { debounce, fetch, ERROR_CANCELED } from 'common/utils';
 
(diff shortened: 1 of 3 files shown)


app/src/components/inputs/autocompletes/withAsyncLoading.jsx
```diff
@@ -89,7 +94,9 @@ export const WithAsyncLoading = (AutocompleteComponent) =>
         ('isOpen' in changes && !this.props.minLength) ||
         ('inputValue' in changes && (inputValue || '').trim().length >= this.props.minLength)
       ) {
-        this.props.onChange(inputValue);
+        if (mode === autocompleteComponentModes.single) {
+          this.props.onChange(inputValue);
+        }
```
AmsterGet, line 99: Does every `SingleAutocomplete` need to trigger this function?

AmsterGet (approved):

PR #3437, opened 2023-03-28
EPMRPP-82714 || Redundant validation for attribute's value in Filter on test item levels

Vadim73i:


app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
```diff
@@ -56,7 +56,7 @@ export class InputConditionalAttributes extends Component {
     keyURLCreator: () => {},
     conditions: [CONDITION_HAS, CONDITION_NOT_HAS, CONDITION_ANY, CONDITION_NOT_ANY],
     projectId: '',
-    isAttributeValueRequired: true,
+    isAttributeValueRequired: false,
```
AmsterGet, line 59: You break the attribute `value` validation in the most places (edit item/launch modals, quality gates, etc.). The `value` is required by default. The case with filters is exceptional case.

AmsterGet (changes requested):

app/src/components/inputs/inputConditionalAttributes/inputConditionalAttributes.jsx
Vadim73i: The AttributeEditor is used for edit modal. But if this is an exceptional case, then I will now fill in the changes.

PR #3445, opened 2023-04-01
EPMRPP-82676 || Change event parameters for Refine filters

Bam6ycha:
Also add events to analytics tab in server settings.


app/src/components/main/analytics/AnalyticsWrapper.jsx
```diff
@@ -37,6 +38,8 @@ const getAppVersion = (buildVersion) =>
     .splice(0, 2)
     .join('.');
 
+export const analyticsEventEmitter = ReactObserver();
```
AmsterGet, line 41: Lets name it as `analyticsEventObserver`

AmsterGet (changes requested):

PR #3461, opened 2023-04-14
EPMRPP-81311 || Implement all plans

Bam6ycha:


app/src/pages/common/uiExtensionPage/uiExtensionPage.jsx
```diff
@@ -14,24 +14,37 @@
...
+            {headerNodes.children}
+          </Header>
+        </div>
+      )}
       <PageSection>
-        <ExtensionLoader extension={extension} withPreloader />
+        <ExtensionLoader extension={extension} withPreloader setHeaderNodes={setHeaderNodes} />
       </PageSection>
-    </PageLayout>
+    </>
```
AmsterGet, line 47: As our core should support both plugin types (with the new design that you've implemented) and the old one, please make this feature switchable.
We need to use old Admin page structure for old plugins and the new one for the newest plugins.

AmsterGet (changes requested):

app/src/pages/common/uiExtensionPage/uiExtensionPage.jsx
```diff
@@ -16,14 +16,18 @@
...
 import { ExtensionLoader, extensionType } from 'components/extensionLoader';
 import { Header } from 'pages/inside/projectSettingsPageContainer/header';
 import classNames from 'classnames/bind';
 import styles from './uiExtensionPage.scss';
 
 const cx = classNames.bind(styles);
 
+const newExtensionsScopesMap = {
+  billing_saas_plugin: true,
+};
```
AmsterGet, line 29: Some notes regarding this.
1. We don't mention any particular plugins in the core UI.
2. It's better to handle this based on the extensions metadata. Just add a property `"newLayout": true` or `"layoutType": "new"` to the proper extension in the plugin's `metadata.json` file.

Later, when we'll migrate all of our plugins to the new layout type, this check will be removed.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3467, opened 2023-04-27
EPMRPP-79135 || "Create Defect Type" modal window does not fit window's height

Vadim73i:


app/src/componentLibrary/modal/modalLayout/modalLayout.jsx
```diff
@@ -42,14 +44,28 @@ export const ModalLayout = ({
...
       setShown(false);
     }
   };
 
+  useEffect(() => {
+    if (modalRef && modalRef.current) {
+      const { clientHeight } = modalRef.current;
+      setModalHeight(clientHeight);
+    }
+  }, [title, headerNode, children, footerNode, modalHeight, modalRef, windowSize]);
```
Bam6ycha, line 67: I think that `title, headerNode,footerNode, modalHeight,modalRef` is redundant dependencies here.

app/src/componentLibrary/modal/modalLayout/modalLayout.jsx
```diff
@@ -42,14 +44,28 @@ export const ModalLayout = ({
   allowCloseOutside,
 }) => {
   const [isShown, setShown] = useState(false);
+  const [modalHeight, setModalHeight] = useState(0);
   const modalRef = useRef();
 
+  const windowSize = useWindowResize();
+  const windowHeight = windowSize.height;
+  const modalMaxHeight = windowHeight * MODAL_MAX_RATIO;
+  const modalPadding = (windowHeight - modalHeight) / 2;
```
Bam6ycha, line 53: As I can see, here we are calculating margin not padding

Bam6ycha (changes requested):

Bam6ycha:
As I can see, we are hardcoding width values. It looks not as flexible as can be.
But, okay. Maybe we can mark this peace of code with todo comment, and count it like technical dept.

PR #3477, opened 2023-05-23
EPMRPP-78550 || Improve attributes accessibility

Bam6ycha:


app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -134,6 +134,16 @@ export const AttributeEditor = ({
   };
   const handleAttributeKeyInputChange = (text) => setState({ ...state, isKeyEdited: !!text });
 
+  const isValidForm = !isFormValid();
```
AmsterGet, line 137: Seems like your variable contains the `is form invalid` state as `!isFormValid()` is opposite to the `is form valid`.

app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -134,6 +134,16 @@ export const AttributeEditor = ({
...
+
+  const handleKeyDown = (handler) => ({ keyCode }) => {
+    const enterKeyCode = 13;
+
+    if (keyCode === enterKeyCode) {
+      handler();
```
AmsterGet, line 143: Can we just call `handleSubmit` directly here?

app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -134,6 +134,16 @@ export const AttributeEditor = ({
   };
   const handleAttributeKeyInputChange = (text) => setState({ ...state, isKeyEdited: !!text });
 
+  const isValidForm = !isFormValid();
+
+  const handleKeyDown = (handler) => ({ keyCode }) => {
+    const enterKeyCode = 13;
```
AmsterGet, line 140: I see the list of Key code constants inside of the `app/src/componentLibrary/dropdown/constants.js`.
Seems like it's time to move them upper to the `common/constans/keyCodes.js` to use application-wide.

AmsterGet (changes requested):

app/src/componentLibrary/attributeEditor/attributeEditor.jsx
Bam6ycha: Oh, yes! You are absolutely right.

app/src/componentLibrary/fieldText/fieldText.scss
```diff
@@ -210,6 +210,7 @@
       @include asterisk();
       right: -7px;
       content: '*';
+      color: $COLOR--red-focused;
```
AmsterGet, line 213: Should it be in red?

app/src/componentLibrary/autocompletes/common/autocompleteOption/autocompleteOption.scss
```diff
@@ -38,6 +38,34 @@
   }
 }
 
+@mixin drawButtonBorder($displayValue, $width) {
```
AmsterGet, line 41: Let's keep the same naming convention for `mixin`s.

app/src/componentLibrary/autocompletes/singleAutocomplete/singleAutocomplete.jsx
```diff
@@ -86,6 +87,13 @@ export class SingleAutocomplete extends Component {
       ...rest,
     });
 
+  handleKeyDown = (event, isOpen, setHighlightedIndex) => {
+    if (event.keyCode === TAB_KEY_CODE && isOpen) {
+      event.preventDefault();
+      setHighlightedIndex(this.props.options.length);
+    }
```
AmsterGet, line 94: Could you please describe this block?

AmsterGet (changes requested):
I cannot move back/further using tab while editing field.

image

Another case:
After clicking on the +Add Attribute

image

The first field should be focused  right after click or should be focused via the next tab.
Please discuss with UX.

app/src/componentLibrary/fieldText/fieldText.scss
Bam6ycha: As I see, it should be in red.
https://www.figma.com/file/gjYQPbeyf4YsH3wZiVKoaj/%F0%9F%9B%A0-RP-DS-6?type=design&node-id=5158-44324&t=JDqzkhmRNRDSB8Y2-0

app/src/componentLibrary/autocompletes/singleAutocomplete/singleAutocomplete.jsx
Bam6ycha: So, if our input is focused and autocomplete menu is opened, I don't want to change the focus using the tab to next input. Because when input is blurred, autocomplete menu automatically closes. To elements at the same time can't be in focus. That is why we need just highlight +add new key or add new value button.

AmsterGet (approved):

PR #3482, opened 2023-05-26
EPMRPP-84138 || Cross icon can be focused when disabled

Bam6ycha:


app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -216,7 +216,7 @@ export const AttributeEditor = ({
           {Parser(CheckIcon)}
         </div>
         <div
-          tabIndex={0}
+          tabIndex={!isCancelButtonDisabled ? 0 : -1}
```
AmsterGet, line 219: May be just use the bool value as is?
```suggestion
          tabIndex={isCancelButtonDisabled ? -1 : 0}
```

AmsterGet (approved):

PR #3491, opened 2023-06-05
EPMRPP-83834 || Add event tracking on assign/unassign

Bam6ycha:


app/src/components/widgets/singleLevelWidgets/tables/projectActivity/activities/unassignUser.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+export const UnassignUser = ({
+  activity: {
+    user,
+    details: { objectName },
+  },
+}) => (
+  <>
+    <span className={cx('user-name')}>{objectName}</span>
+    <FormattedMessage id="UnassignUser.by" defaultMessage="was unassigned from the project by" />
```
Vadim73i, line 32: As far as I remember, we agreed to use formatMessage

Bam6ycha: @AmsterGet what do you think about that?

I used FormatterdMessage component to avoid importing defineMessages and useIntl hook here, for only one small component.

AmsterGet: @Vadim73i As I remember, my comment about using `formatMessage` was related to choosing single approach for particular component (do not mix `formatMessage` and `FormattedMessage`).
So, if the component is fairly simple and the messages are not reused somewhere (in other components/descendants), we can use `FormattedMessage`.
Otherwise `intl` and  `formatMessage`.

Vadim73i (approved):

app/src/pages/admin/projectsPage/messages.js
```diff
@@ -50,9 +50,21 @@ export const messages = defineMessages({
     id: 'ProjectDetailsPageMembers.title',
     defaultMessage: 'Members',
   },
-  [`${EVENTS}Title`]: {
+  [`${MONITORING}Title`]: {
```
AmsterGet, line 53: Can we combine these messages with presented below to avoid duplications?

image

AmsterGet (changes requested):

AmsterGet (approved):

PR #3495, opened 2023-06-06
EPMRPP-81639 || User cannot enter the whole username in "Recipients" field on "Notifications" tab

Vadim73i:


app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
```diff
@@ -203,7 +204,10 @@ export const MultipleAutocomplete = ({
                 updatePosition = scheduleUpdate;
                 const filteredOptions = options.filter(
                   (item) =>
-                    value.indexOf(item) < 0 && item.toLowerCase() !== inputValue.toLowerCase(),
+                    value.indexOf(item) < 0 &&
```
Bam6ycha, line 207: Could we give this expression a more descriptional name?
It is absolutely hard to figure out what we are doing here.
And, after that, maybe we could find an easier way to handle this scenario.

Bam6ycha (changes requested):

app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
```diff
@@ -201,10 +202,20 @@ export const MultipleAutocomplete = ({
...
+                    inputValue,
+                  );
+
+                  return (
+                    !isOptionSelect &&
+                    (duplicateInputValueOption ||
+                      !isOptionEqualInputValue ||
+                      !isInputValueMatchAdditionalCreationCondition)
+                  );
+                });
```
AmsterGet, line 218: If there are no concerns, I suggest to left simple condition here:
```javascript
const filteredOptions = options.filter((item) => value.indexOf(item) < 0);
```
I didn't find any reasons for excluding the options with same value as typed in input.

app/src/componentLibrary/autocompletes/multipleAutocomplete/multipleAutocomplete.jsx
```diff
@@ -52,6 +52,7 @@ export const MultipleAutocomplete = ({
   handleUnStoredItemCb,
   dataAutomationId,
   existingItemsMap,
+  duplicateInputValueOption,
```
AmsterGet, line 55: Please remove the prop as it is no longer in use.

AmsterGet (approved):

Bam6ycha (approved):

PR #3508, opened 2023-06-26
EPMRPP-84533 || Login with SAML. The error message is displayed 2 times in case user logs in via SAML with already existing email

Vadim73i:


Bam6ycha (approved):

app/src/pages/outside/loginPage/loginPage.jsx
```diff
@@ -59,28 +64,32 @@ export class LoginPage extends PureComponent {
...
 
-  componentDidUpdate(prevProps) {
-    if (this.props.errorAuth !== prevProps.errorAuth) {
+  componentDidUpdate() {
+    if (this.props.errorAuth) {
       this.props.showDefaultErrorNotification({
         message: this.props.errorAuth,
       });
+      this.props.clearErrorAuth();
     }
```
AmsterGet, line 93: You can move this block into method and call it twice instead of duplicating the same logic.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3510, opened 2023-06-26
EPMRPP-84595 || Attribute label is not focused after editing it

Bam6ycha:


app/src/componentLibrary/attributeList/attributeList.jsx
```diff
@@ -111,18 +130,24 @@ export const AttributeList = ({
...
       newAttributes.splice(index, 1);
+
+      handleAddNewAttrFocus();
     } else {
       const { edited, ...attribute } = newAttributes[index];
       newAttributes[index] = attribute;
+
+      scheduleUpdate(() =>
+        handleChangeFocus(
+          getAttributesMapValue(getAttributeUniqueKey(attributes[index]), attributeWrapperRef),
```
AmsterGet, line 145: Is it possible to use `attribute` here? Looks like it is the same with `attributes[index]`.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3517, opened 2023-07-05
EPMRPP-81039 || Add test automation attributes to elements on project settings Pattern-Analysis tab

Vadim73i:


app/src/pages/inside/projectSettingsPageContainer/content/patternAnalysis/patternAnalysisContent/patternAnalysisContent.jsx
```diff
@@ -56,7 +56,9 @@ export const PatternAnalysisContent = ({
   useEffect(() => {
     setHeaderTitleNode(
       <span className={cx('button')} onClick={onAddPattern}>
-        <Button disabled={disabled}>{formatMessage(messages.createPatternModalHeader)}</Button>
+        <Button disabled={disabled} dataAutomationId={'createPatternButton'}>
```
Bam6ycha, line 59: We can omit curly braces here.

app/src/pages/inside/projectSettingsPageContainer/content/patternAnalysis/modals/editPatternModal/editPatternModal.jsx
```diff
@@ -77,7 +77,12 @@ const EditPatternModal = ({ data, handleSubmit, initialize, dirty }) => {
       allowCloseOutside={!dirty}
     >
       <div className={cx('description')}>{formatMessage(messages.editPatternName)}</div>
-      <FieldElement name="name" label={formatMessage(messages.patternName)} isRequired>
+      <FieldElement
+        name="name"
+        label={formatMessage(messages.patternName)}
+        isRequired
+        dataAutomationId="patternNameField"
```
Bam6ycha, line 84: As I can see, the FiledText component does nothing with the dataAutomationId prop.

Bam6ycha:
I am not sure but looks like it is better to create a separate file with dataAutomationId values, and store these values in this file.
Because in case, when we need to change dataAutomationId, we will be forced to search for it all over the application.
What do u think?

Bam6ycha (changes requested):

Bam6ycha (approved):

PR #3539, opened 2023-07-19
EPMRPP-845560 || Accessibility.Toggle

Bam6ycha:
Some styles from checkbox, radio button were also changed


app/src/componentLibrary/toggle/toggle.scss
```diff
@@ -14,6 +14,20 @@
  * limitations under the License.
  */
 
+@mixin draw-after-pseudo-element {
```
AmsterGet, line 17: Let's rename it according to its purpose, not the tools it uses.

PR #3543, opened 2023-07-24
EPMRPP-84862 || Update webpack, react

AmsterGet:
Also in the scope:
- Remove storybook;
- Remove polyfills for IE 11;
- Add circular dependencies detector plugin to dev webpack config;
- HMR fix;
- Swagger update;
- React-intl update;
- `path` lib fallback added to webpack config as the [webpack](https://webpack.js.org/configuration/resolve/#resolvefallback) from version 5 does not polyfill it by default;
- `AbsRelTime` component now converts `startTime` prop to `Date` object for correct parsing (see the screenshots below) and to avoid 'React infinite loop errors' in some corner cases.

Before:
image
After:
image


Vadim73i (approved):

Bam6ycha:
1. I have 23 warnings about circular dependencies. I think that we need to make capacity and fix these warnings.
2. I suggest turning on style lint during the pre-commit script(maybe not in the scope of this task)
All the rest looks good to me.

Bam6ycha (approved):

AmsterGet:
> 1. I have 23 warnings about circular dependencies. I think that we need to make capacity and fix these warnings.
> 2. I suggest turning on style lint during the pre-commit script(maybe not in the scope of this task)
>    All the rest looks good to me.

Added to the backlog, thanks.

PR #3559, opened 2023-08-15
EPMRPP-84210 || Add GA4 events for Project Monitoring Administrate

Bam6ycha:


app/src/components/inputs/inputFilter/inputFilter.jsx
```diff
@@ -99,7 +111,22 @@ export class InputFilter extends Component {
...
+        .join('#');
+
+      if (analyticsData) {
+        this.props.tracking.trackEvent(this.props.eventsInfo.CLICK_APPLY_BUTTON(analyticsData));
+      }
+
+      this.prevAppliedFilters.current = currentAppliedFilters;
+    } else {
+      this.props.tracking.trackEvent(this.props.eventsInfo.applyBtn);
+    }
```
AmsterGet, line 128: Can we track these events using callbacks provided to the component?
We can pass them through the `EventsToolbar` (move `track` wrapper to it as well) during render.

AmsterGet (changes requested):

app/src/components/inputs/inputFilter/inputFilter.jsx
Bam6ycha: Yes we can, but looks like we need to rewrite not only EventsToolbar but `ProjectsToolbal` `UsersToolbar`

AmsterGet (approved):

PR #3567, opened 2023-08-18
EPMRPP-85395 || Fix positioning of service versions block

Bam6ycha:


app/src/pages/outside/loginPage/loginPage.scss
```diff
@@ -93,3 +93,11 @@
     height: 25px;
   }
 }
+
+.footer {
+  padding-bottom: 50px;
+
+  &.footer-with-policyBlock {
```
AmsterGet, line 100: Let's be more brief: `.with-policy`.

AmsterGet (approved):

PR #3584, opened 2023-09-11
EPMRPP-78737 || Create filter entity from queryparams

Bam6ycha:


Vadim73i (approved):

app/src/controllers/filter/selectors.js
```diff
@@ -68,3 +72,17 @@ export const dirtyFilterIdsSelector = createSelector(
...
 );
+
+export const filterFromQuerySelector = createSelector(
+  queryParametersSelector,
+  ({ [SIZE_KEY]: size, [SORTING_KEY]: sort, [PAGE_KEY]: page, ...filter }) => {
+    if (isEmptyObject(filter)) {
+      return null;
+    }
+
+    const [[filterName, filterValue]] = Object.entries(filter);
```
AmsterGet, line 83: Actually there can be several filters (even if now we don't support such case from email notifications, it can be useful in the future f.e. by creating a link to a group of launches from the client/agent side).
Therefore, please parse all possible filters accordingly, not just only one.

app/src/controllers/filter/sagas.js
```diff
@@ -174,6 +177,18 @@ function* saveNewFilter({ payload: filter }) {
   yield put(changeActiveFilterAction(newId));
 }
 
+function* parseQueryToFilterEntity() {
+  const filterConditions = yield select(filterFromQuerySelector);
+  const projectId = yield select(projectIdSelector);
+
+  if (filterConditions) {
+    yield put(createFilterAction(filterConditions));
+    yield put(redirect({ type: LAUNCHES_PAGE, payload: { projectId } }));
```
AmsterGet, line 186: This leads to the double launches request to the server.
Can we add a parameter to the `createFilterAction` to dispatch the next page action with redirect and remove this line?

AmsterGet (changes requested):

app/src/controllers/filter/selectors.js
```diff
@@ -68,3 +73,18 @@ export const dirtyFilterIdsSelector = createSelector(
...
+    if (isEmptyObject(filters)) {
+      return null;
+    }
+
+    return Object.entries(filters).reduce(
+      (acc, [filterKey, filterValue]) => collectFilterConditions(filterKey, filterValue, acc),
+
+      { conditions: [] },
+    );
+  },
```
AmsterGet, line 89: Can be done using existing utils:

`{ conditions: addFilteringFieldToConditions(collectFilterEntities(filters)) }`

Also please move `collectFilterEntities` to `filter/utils` as this place looks more preferable for it.

AmsterGet (changes requested):

BlazarQSO (approved):

AmsterGet (approved):

PR #3594, opened 2023-09-29
EPMRPP-86766 || Import event is not filtered by the search on Monitor…

BlazarQSO:
…ing page

### DIFF
--- a/app/localization/translated/be.json
+++ b/app/localization/translated/be.json
@@ -699,6 +699,8 @@
   "EventActions.deletePattern": "Выдаленне шаблону",
   "EventActions.deleteWidget": "Выдаленне віджэта",
   "EventActions.import": "Імпарт",
+  "EventActions.startImport": "Пачаць імпарт",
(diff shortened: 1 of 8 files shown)


app/src/components/inputs/inputDropdown/inputDropdown.jsx
```diff
@@ -112,12 +112,20 @@ export class InputDropdown extends Component {
     }
   };
 
+  isIncludedValue(optionValue) {
```
Bam6ycha, line 115: Looks like these changes can bring a lot of other issues.

You can take a look at app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/projectActivityControls.jsx
We have almost the same case here.

And looks like we need Import start and Import finish messages because they are used in app/src/pages/admin/projectEventsPage/eventsGrid/eventsGrid.jsx

Bam6ycha (changes requested):

PR #3595, opened 2023-09-29
EPMRPP-85652 || Width of API Key table is different from table

BlazarQSO:


app/src/components/main/navigationTabs/navigationTabs.scss
```diff
@@ -17,7 +17,7 @@
 .navigation-tabs {
   width: 100%;
   box-sizing: border-box;
-  padding: 5px 15px 15px 15px;
+  padding: 5px 45px 15px 45px;
```
Bam6ycha, line 20: I think it is better not to change the padding directly in the general-purpose component.
It is used in several other places, and padding there 5px 15px 15px 15px;

app/src/components/main/navigationTabs/navigationTabs.scss
```diff
@@ -27,7 +27,7 @@
   box-sizing: border-box;
   width: 100%;
   padding-top: 15px;
-  border-bottom: 1px solid $COLOR--gray-91;
+  border-bottom: 2px solid $COLOR--gray-91;
```
Bam6ycha, line 30: I am not sure, that we need to change the width of this line.
https://www.figma.com/file/PxNk9h6CS9Y4Mxcl2fS9HF/RP5-Project-Level?type=design&node-id=1453-144839&mode=design&t=MVVjCbM1Y6Y6nktj-0
The same component with 1px border width
Discuss this with BA and UI/UX

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.jsx
```diff
@@ -61,7 +61,7 @@ export const ApiKeysBlock = ({ apiKeys }) => {
   const onRevokeClick = (key) => dispatch(showModalAction({ id: 'revokeApiKeyModal', data: key }));
 
   return (
-    <div className={cx('api-keys-block')}>
+    <div>
```
Bam6ycha, line 64: Could we replace it with Fragment?

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.scss
```diff
@@ -14,11 +14,6 @@
  * limitations under the License.
  */
 
-.api-keys-block {
-  max-width: 860px;
```
Bam6ycha, line 18: According to the design, the width of each tab is 860 px
Are you really sure that we need to delete it here?
Mb we need to change the width of other tabs?
Could you please discuss it with BA?

Bam6ycha (changes requested):

app/src/components/main/navigationTabs/navigationTabs.scss
BlazarQSO: according to the design from the ticket:
https://www.figma.com/file/PxNk9h6CS9Y4Mxcl2fS9HF/RP5-Project-Level?type=design&node-id=7756-47207&mode=design&t=UU8oBhhWIdmc6L2H-0

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.jsx
BlazarQSO: Done.

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.scss
BlazarQSO: 860px when the screen is 1320px
If the screen is larger then the width of the content also increases. (Victoria confirmed)

app/src/pages/inside/profilePage/profilePage.jsx
```diff
@@ -117,9 +117,11 @@ export class ProfilePage extends Component {
               onChangeTab={this.props.dispatch}
             />
           </div>
-          <div className={cx('footer')}>
-            <LocalizationBlock />
-            <DeleteAccountBlock />
+          <div>
```
Bam6ycha, line 120: Why do we need this extra div wrapper here?

app/src/pages/inside/profilePage/profilePage.scss
```diff
@@ -29,12 +29,16 @@
     .section-wrapper {
       padding: 0 30px;
     }
+    [class^="navigationTabs__navigation-tabs"] {
```
Bam6ycha, line 32: I think that we can achieve the same result by adding NavigationTabs to the section-wrapper div
Screenshot 2023-10-03 at 10 56 42
And we do not change anything in CSS

Bam6ycha (changes requested):

app/src/pages/inside/profilePage/profilePage.scss
BlazarQSO: Done.

app/src/pages/inside/profilePage/profilePage.jsx
BlazarQSO: Done.

Bam6ycha (approved):

PR #3596, opened 2023-10-02
EPMRPP-86198 || Attribute's value is not truncated in Quality Gates c…

BlazarQSO:
…onstructor

### DIFF
--- a/app/src/componentLibrary/attributeList/editableAttribute/attribute/attribute.jsx
+++ b/app/src/componentLibrary/attributeList/editableAttribute/attribute/attribute.jsx
@@ -82,7 +82,9 @@ export const Attribute = ({
             </div>
           </>
         ) : (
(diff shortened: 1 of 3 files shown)


app/src/componentLibrary/attributeList/editableAttribute/attribute/attribute.jsx
```diff
@@ -82,7 +82,9 @@ export const Attribute = ({
             </div>
           </>
         ) : (
-          <div className={cx('value-without-key')}>{attribute.value}</div>
+          <div className={cx('value-without-key')} style={{ maxWidth: maxCellWidth * 2 }}>
```
Bam6ycha, line 85: I am not sure, that we need className here

Bam6ycha (changes requested):

PR #3604, opened 2023-10-13
EPMRPP-86794 || OAuth test connection fails when opening integrations…

BlazarQSO:
… page

### DIFF
--- a/app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
+++ b/app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
@@ -58,9 +58,18 @@ export const IntegrationSettings = (props) => {
   const query = useSelector(querySelector);
   const dispatch = useDispatch();
   const { trackEvent } = useTracking();
-  const availableGlobalIntegrations = globalIntegrations[query.subPage] || [];
-  const availableProjectIntegrations = projectIntegrations[query.subPage] || [];
-  const groupedIntegrations = [...availableGlobalIntegrations, ...availableProjectIntegrations];
+  const availableGlobalIntegrations = useMemo(() => globalIntegrations[query.subPage] || [], [
+    globalIntegrations,
+    query.subPage,
+  ]);
+  const availableProjectIntegrations = useMemo(() => projectIntegrations[query.subPage] || [], [
+    projectIntegrations,
+    query.subPage,
+  ]);
+  const groupedIntegrations = useMemo(
+    () => [...availableGlobalIntegrations, ...availableProjectIntegrations],
+    [availableGlobalIntegrations, availableProjectIntegrations],
+  );
 
   const namedSubPage = useMemo(
     () => ({
@@ -74,8 +83,8 @@ export const IntegrationSettings = (props) => {
   );
 
   const testIntegrationConnection = useCallback(() => {
-    setLoading(true);
-    if ('id' in props.data) {
+    if ('id' in props.data && !props.preventTestConnection) {
+      setLoading(true);
       fetch(URLS.testIntegrationConnection(projectId || activeProject, props.data.id))
         .then(() => {
           setConnected(true);
@@ -86,7 +95,7 @@ export const IntegrationSettings = (props) => {
           setConnected(false);
         });
     }
-  }, [props.data, activeProject, projectId]);
+  }, [props.data, activeProject, projectId, props.preventTestConnection]);
 
   useEffect(() => {
     const hasId = groupedIntegrations.some((value) => value.id === +query.id);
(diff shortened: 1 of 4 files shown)


app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
```diff
@@ -62,6 +63,18 @@ export const IntegrationSettings = (props) => {
...
+    data,
+    onUpdate,
+    formFieldsComponent,
+    editAuthConfig,
+    isEmptyConfiguration,
+    formKey,
+    isGlobal,
+  } = props;
+
+  const pluginName = props.data.integrationType?.name;
```
Bam6ycha, line 76: Looks like it is better to extract plugin name from data using destructuring.
 const {
    data,
    onUpdate,
    formFieldsComponent,
    editAuthConfig,
    isEmptyConfiguration,
    formKey,
    isGlobal,
    data:{integrationType:{name:pliginName}}={}
  } = props;

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
```diff
@@ -74,8 +87,7 @@ export const IntegrationSettings = (props) => {
   );
 
   const testIntegrationConnection = useCallback(() => {
-    setLoading(true);
-    if ('id' in props.data) {
+    if ('id' in props.data && pluginName !== AD && pluginName !== LDAP && pluginName !== SAML) {
```
Bam6ycha, line 90: As for me, such a long condition looks bad.

May be in is better f.e. create an object such as authPlugins={
[AD]:AD,
...
}
and use it like this 
if('id' in data && !authPlugins[pluginName])

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
```diff
@@ -62,6 +63,18 @@ export const IntegrationSettings = (props) => {
   const availableProjectIntegrations = projectIntegrations[query.subPage] || [];
   const groupedIntegrations = [...availableGlobalIntegrations, ...availableProjectIntegrations];
```
Bam6ycha, line 64: Could you fix this SonarQube warning?

Bam6ycha (changes requested):

AmsterGet (changes requested):
Please check the integration Settings components by the next path: src/components/integrations/integrationProviders.
Inside of the `SamlSettings`, `LdapSettings`, `ActiveDirectorySettings` the `preventTestConnection` prop exists.
So I suggest to just implement this prop usage inside of the `IntegrationSettings` to not hardcode the names of integrations that don't need to check connections, but hold this responsibility on each particular plugin implementation.

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
BlazarQSO: Reverted all changes

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
BlazarQSO: Reverted all changes

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
BlazarQSO: Reverted all changes

BlazarQSO: Fixed warning

app/src/components/integrations/elements/integrationSettings/integrationSettings.jsx
```diff
@@ -58,9 +58,18 @@ export const IntegrationSettings = (props) => {
   const query = useSelector(querySelector);
   const dispatch = useDispatch();
   const { trackEvent } = useTracking();
-  const availableGlobalIntegrations = globalIntegrations[query.subPage] || [];
-  const availableProjectIntegrations = projectIntegrations[query.subPage] || [];
-  const groupedIntegrations = [...availableGlobalIntegrations, ...availableProjectIntegrations];
+  const availableGlobalIntegrations = useMemo(() => globalIntegrations[query.subPage] || [], [
```
Bam6ycha, line 61: Could we combine these 3 useMemo into one?

BlazarQSO: Done.

Bam6ycha (approved):

AmsterGet (approved):

PR #3606, opened 2023-10-16
Correct Chinese translation

shenghu:
1. Translate strings which have not been translated.
2. Correct some mistakes.


AmsterGet (approved):
Hello @shenghu !
Thank you very much for your input.
Chinese translation with updates from you will be released in the next version.

PR #3614, opened 2023-10-23
EPMRPP-86865 || Update the twitter logo on the Login page

BlazarQSO:


Bam6ycha (changes requested):
I think that it is better to change the CSS properties of SVG on hover, or on click
rather than have three different svgs

Bam6ycha (changes requested):
I think that it is better to change the CSS properties of SVG on hover, or on click
rather than have three different svgs

AmsterGet:
> I think that it is better to change the CSS properties of SVG on hover, or on click rather than have three different svgs

We have a separate task to change this behavior for all social icons, so I suppose it can be done separately.

AmsterGet (approved):

Bam6ycha (approved):

PR #3626, opened 2023-10-30
EPMRPP-86979 || Webpack configuration error

BlazarQSO:


Bam6ycha:
So, I can not see that the expected result from the ticket AC is reached.
Screenshot 2023-10-30 at 11 43 10
We need this hash, to force the browser not to take this resource from the cache in case when this hash is changed

Bam6ycha (approved):

AmsterGet (approved):

PR #3632, opened 2023-11-02
EPMRPP-85840 || Add events for plugins

Bam6ycha:


app/src/components/integrations/modals/addIntegrationModal/addIntegrationModal.jsx
```diff
@@ -62,13 +63,25 @@ const AddIntegrationModal = ({ data, initialize, change, handleSubmit }) => {
...
+  const integrationFieldsExtension = fieldsExtensions.find(
+    (ext) => ext.pluginName === data.instanceType,
+  );
 
   const updateMetaData = (newMetaData) => {
     setMetaData({ ...metaData, ...newMetaData });
   };
 
   const onSubmit = (newData) => {
+    trackEvent(createGlobalIntegration(integrationFieldsExtension.pluginName));
```
AmsterGet, line 83: This modal used to create project integrations as well.
So please use more generic name here.
Also can we avoid providing events from props in that case? As this modal used several times for different plugins and default props won't work for most of the cases.
It supports aka 4 mods: create global + editAuth for global, create project + editAuth for project.
You can determine the right event using the existing props `isGlobal` & `editAuthMode`.

app/src/components/main/analytics/events/pluginsPageEvents.js
```diff
@@ -51,18 +51,33 @@ export const getUninstallPluginBtnClickEvent = (pluginName) => ({
...
+    element_name: 'upload',
+    place: 'navigation_panel',
   },
+  UPLOAD_BTN_UPLOAD_MODAL: (type) => ({
+    ...BASIC_PLUGINS_EVENT_PARAMS,
+    element_name: 'upload',
+    modal: 'upload_plugin',
+    type,
+  }),
+  CLICK_CREATE_GLOBAL_INTEGRATION: (type) => ({
```
AmsterGet, line 69: As I remember we've agreed to use camelCase for events that are functions.

app/src/components/main/analytics/events/pluginsPageEvents.js
```diff
@@ -51,18 +51,33 @@ export const getUninstallPluginBtnClickEvent = (pluginName) => ({
...
+    ...BASIC_PLUGINS_EVENT_PARAMS,
+    element_name: 'button_create',
+    modal: 'create_project_integration',
```
AmsterGet, line 72: It's a little bit confusing that the event name and modal are mismatch each other.
Please recheck with WA. May be its better to name the modal just as `create_integration`.

app/src/pages/common/modals/importModal/importModal.jsx
```diff
@@ -222,6 +224,14 @@ export class ImportModal extends Component {
...
     this.props.data.onImport();
     const { files } = this.state;
+    const {
+      tracking: { trackEvent },
+      data: {
+        eventsInfo: { uploadButton },
+      },
+    } = this.props;
+
+    trackEvent(uploadButton(this.getFilesNames(files)));
```
AmsterGet, line 234: I suppose you need to add additional checks for such props availability, as they are part of the `data` object, their default values won't work in case the object will be passed to the component.

AmsterGet (changes requested):

app/src/components/main/analytics/events/pluginsPageEvents.js
Bam6ycha: Discussed with WA.
Will be renamed in create_global_integration

app/src/components/integrations/modals/addIntegrationModal/addIntegrationModal.jsx
```diff
@@ -62,13 +64,24 @@ const AddIntegrationModal = ({ data, initialize, change, handleSubmit }) => {
...
+  );
+
   const updateMetaData = (newMetaData) => {
     setMetaData({ ...metaData, ...newMetaData });
   };
 
   const onSubmit = (newData) => {
+    if (isGlobal && !customProps.editAuthMode) {
+      trackEvent(
+        PLUGINS_PAGE_EVENTS.clickCreateGlobalIntegration(integrationFieldsExtension.pluginName),
```
AmsterGet, line 81: Please use `data.instanceType` instead of `integrationFieldsExtension.pluginName` as not all plugins support fields as extensions.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3635, opened 2023-11-03
EPMRPP-82207 || Add events for Documentation links in Project Settings

Bam6ycha:


AmsterGet (changes requested):
I see that this part repeated at least 7 times.
image
Also I don't see much cases to reuse `handleDescriptionClick` and `handleDescriptionKeyDown` with the Layout component, so it can be encapsulated somewhere.
What do you think about creation a separate component that will render formatted description and handle all necessary events inside? In that case the Layouts `description` prop can be a react node.

app/src/pages/inside/projectSettingsPageContainer/content/elements/formattedDescription/index.js
```diff
@@ -0,0 +1 @@
+export { FormattedDescription } from './formattedDescription';
```
AmsterGet, line 1: Please add copyright notice here as well.

app/src/pages/inside/projectSettingsPageContainer/content/elements/formattedDescription/formattedDescription.scss
```diff
@@ -0,0 +1,34 @@
...
+.formatted-description {
+  font-family: $FONT-ROBOTO-REGULAR;
+  font-size: 13px;
+  line-height: 20px;
+  color: $COLOR--almost-black;
+
+  & > a {
+    font-family: $FONT-ROBOTO-REGULAR;
+    font-size: 13px;
+    line-height: 20px;
```
AmsterGet, line 26: Can be omitted - these properties will be inherited.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3637, opened 2023-11-03
EPMRPP-85839 || Add events for API KEYS

BlazarQSO:


app/src/components/main/analytics/events/ga4Events/profilePageEvent.js
```diff
@@ -14,18 +14,57 @@
...
     type: lang,
     element_name: 'language_drop_down',
   }),
+
+  CLICK_API_KEYS_TAB_EVENT: {
+    ...BASIC_EVENT_PARAMETERS_PROFILE,
+    element_name: 'tab_api_keys',
+  },
+
+  CLICK_GENERATE_API_KEY_BUTTON: (isApiKey) => ({
```
Bam6ycha, line 36: Let's use camelCase for event creators.

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.jsx
```diff
@@ -56,13 +59,42 @@ const messages = defineMessages({
...
+  const onRevokeClick = (key) => {
+    dispatch(showModalAction({ id: 'revokeApiKeyModal', data: key }));
+    trackEvent(PROFILE_EVENTS.CLICK_REVOKE_BUTTON);
+  };
+
+  const onDocumentationClick = (event) => {
+    const { tagName } = event.target;
+
+    if (tagName === 'A') {
+      trackEvent(PROFILE_EVENTS.CLICK_DOCUMENTATION_LINK(true));
```
Bam6ycha, line 78: Looks like this violates one of the clean code principles.
If we are using static bool arg, it is better to create two different functions

Bam6ycha (changes requested):

app/src/components/main/analytics/events/ga4Events/profilePageEvent.js
BlazarQSO: CONSTANT_CASE is used everywhere in event creators.

app/src/pages/inside/profilePage/apiKeys/apiKeysBlock/apiKeysBlock.jsx
BlazarQSO: Done.

Bam6ycha (approved):

AmsterGet (approved):

app/src/components/main/analytics/events/ga4Events/profilePageEvent.js
AmsterGet: When migrating to GA4, we agreed to use CamelCase for event creators to better distinguish constant events from them.
So if there are some events with old naming that affected by particular task, let's provide new names for them according to above agreement.

PR #3647, opened 2023-11-10
EPMRPP-82655 || Implement Error Boundary for plugin extensions

AmsterGet:


app/src/components/extensionLoader/extensionLoader.jsx
```diff
@@ -35,18 +36,31 @@ ExtensionLoader.defaultProps = {
...
 }
 ExtensionLoaderWrapper.propTypes = {
   extension: extensionType,
   withPreloader: PropTypes.bool,
+  silentOnError: PropTypes.bool,
 };
 ExtensionLoaderWrapper.defaultProps = {
   extension: {},
   withPreloader: false,
+  silentOnError: true,
```
Bam6ycha, line 65: Maybe it is better to make it `false` by default?
Looks like we are using this component only with `silentOnError={false}`

Bam6ycha (changes requested):

app/src/components/extensionLoader/extensionLoader.jsx
AmsterGet: There are much more cases where I left it `true` as default (quality gate run info, unique errors page extensions, sidebar icons, etc.)

Bam6ycha (dismissed):

Vadim73i (dismissed):

PR #3665, opened 2023-11-21
EPMRPP-87336 || Add BTS field description

AmsterGet:
Usage of `FormField` & `IntegrationFormField` replaced with `FieldElement`.
Styles for Post issue modal will be adjusted in the scope of EPMRPP-87463.

### DIFF
--- a/app/src/components/fields/dynamicFieldsSection/dynamicField/dynamicField.jsx
+++ b/app/src/components/fields/dynamicFieldsSection/dynamicField/dynamicField.jsx
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 EPAM Systems
+ * Copyright 2023 EPAM Systems
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


app/src/components/fields/dynamicFieldsSection/dynamicField/dynamicField.jsx
```diff
@@ -40,30 +39,30 @@ export const DynamicField = ({
...
-  ) : (
-    <FormField
+  return (
+    <FieldElement
+      label={field.fieldName}
+      isRequired={field.required}
+      descriptionSecondary={field.description}
+      className={cx('dynamic-field')}
+      additionalInfo={customBlock}
+      normalize={(value) => value}
```
Bam6ycha, line 54: Why do we need these props?

Maybe we will add an ability to provide them through the `DynamicField` props. What are your thoughts on that?

AmsterGet: They are the defaults from `FormField` component used previously.
And we already have such ability to override them defaults via `...fieldCommonProps`, f.e. [arrayField](https://github.com/reportportal/service-ui/blob/feature/EPMRPP-87336-Add-BTS-field-description/app/src/components/fields/dynamicFieldsSection/fields/arrayField.jsx#L66) use it.
I'm not sure if we need to mention all redux-form props here in prop types.

Bam6ycha (approved):

PR #3677, opened 2023-11-30
EPMRPP-87603 || Possible to enter non existent project when adding new user

Bam6ycha:


Vadim73i (dismissed):

AmsterGet (changes requested):
I guess the issue occurred, because of the default behavior of Redux form onBlur function that save the input value to the redux store by default, so the easiest way to fix the issue - overwrite `onBlur` callback or use `FieldProvider` prop `dumbOnBlur` to prevent default behavior. In addition to this I suggest to remove `hintType` property from this field to avoid unexpected hint behavior without default blur.
Also please apply the fix for a `addUserModal` as well.

PR #3719, opened 2024-01-11
EPMRPP-88958 || Tooltip is missing when Service Analyzer is not running

BlazarQSO:


app/src/pages/inside/projectSettingsPageContainer/content/analyzerContainer/autoAnalysis/autoAnalysis.scss
```diff
@@ -0,0 +1,19 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.auto-analyzer {
+  width: fit-content;
+}
```
AmsterGet, line 19: Could you please explain why is it needed?
As I see we need to provide the title only.

app/src/pages/inside/projectSettingsPageContainer/content/elements/fieldElement/fieldElement.jsx
```diff
@@ -41,7 +42,7 @@ export const FieldElement = (props) => {
   const getChildren = () =>
     withoutProvider ? children : <FieldProvider {...rest}>{children}</FieldProvider>;
   return (
-    <div className={cx('wrapper', className)} data-automation-id={dataAutomationId}>
+    <div className={cx('wrapper', className)} data-automation-id={dataAutomationId} title={title}>
```
AmsterGet, line 45: I suppose it would be better to extend `Checkbox` component with `title` property and use it directly for it.
We typically use tooltips and titles for interactive elements rather than their parent containers.

AmsterGet (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/elements/fieldElement/fieldElement.jsx
BlazarQSO: Done.

AmsterGet (approved):

PR #3729, opened 2024-01-18
EPMRPP-87445 || Add ability to exclude skipped tests from statistics …

BlazarQSO:
…for component health check widget


app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/componentHealthCheckControls.jsx
```diff
@@ -179,6 +185,13 @@ export class ComponentHealthCheckControls extends Component {
                 )}
               />
             </FieldProvider>
+            <FieldProvider
+              name="contentParameters.widgetOptions.excludeSkipped"
+              format={Boolean}
+              className={cx('checkbox')}
+            >
+              <Checkbox>{formatMessage(messages.excludeSkipped)}</Checkbox>
+            </FieldProvider>
```
AmsterGet, line 194: Please use the existing `CheckboxControl` from `src/pages/inside/dashboardItemPage/modals/common/widgetControls/controls/checkboxControl.jsx`.
It's usage example for another widget:
image

AmsterGet (changes requested):
Also according to the description we need to hide the `SKIPPED` column in widget in case of `excludeSkipped` set to `true`.

image

app/src/pages/inside/dashboardItemPage/modals/common/widgetControls/componentHealthCheckControls.jsx
BlazarQSO: Done.

AmsterGet (approved):

PR #3743, opened 2024-02-05
EPMRPP-87611 || Update components colors and opacity for checkboxes, …

BlazarQSO:
…radio buttons and toggles


AmsterGet (changes requested):
Please make these changes to the `Checkbox` component itself.

Our design system and components based on it should be the source of truth for the right designs. All changes that affects the visual representation of the DS component from outside should be minimized (like an exceptions in some rare cases when the component update is not possible/desirable).
In this case, according to the task, all new `Checkbox`es from DS should have new style. If there are some places not listed in the task (for some reason), they should be updated as well to be consistent with component use cases.

Mutating styles for component internals hardens maintenance of the code and can lead to unexpected design mismatch in the future (f.e. once the `Checkbox` component will be delivered from external library via ui-kit) in case of changes in component internals and structure.

AmsterGet (approved):

PR #3753, opened 2024-02-29
EPMRPP-89462 || switch to new batch delete operation for projects

BlazarQSO:


app/src/controllers/administrate/projects/sagas.js
```diff
@@ -116,7 +116,7 @@ function* watchAddProject() {
 
 function* deleteProject({ payload: project }) {
   try {
-    yield call(fetch, URLS.project(project.id), {
+    yield call(fetch, URLS.projectByName(project.id), {
```
AmsterGet, line 119: Looks like it was an unexpected usage of endpoint with `projectName`, because here we use a project `id`.
Let's use it as `URLS.project([project.id])`.
```suggestion
    yield call(fetch, URLS.project([project.id]), {
```

BlazarQSO: Done.

AmsterGet (approved):

PR #3793, opened 2024-04-09
EPMRPP-89400 || Set bearer token in api doc

Vadim73i:
## PR Checklist


app/src/pages/inside/apiPage/apiPage.jsx
```diff
@@ -45,14 +45,18 @@ const OPTION_BLOCK_TAG_OPEN = '.opblock.is-open';
 const OPTION_BLOCK_SUMMARY = '.opblock-summary';
 
 @connect((state) => ({
-  token: tokenSelector(state),
+  tokenString: tokenSelector(state),
```
AmsterGet, line 48: If we already select the `tokenType` and `tokenValue` individually, does it make sense to not select them as a joined string additionally?
I suppose we can just join them right in the component if needed and omit `tokenSelector` usage here.

AmsterGet (approved):

PR #3802, opened 2024-04-15
EPMRPP-90127 || Popover is not collapsed when clicking on the same op…

BlazarQSO:
…tion

## PR Checklist

### DIFF
--- a/app/src/componentLibrary/popover/withPopover/withPopover.jsx
+++ b/app/src/componentLibrary/popover/withPopover/withPopover.jsx
@@ -30,7 +30,8 @@ export const withPopover = ({
   ...popoverConfig
 }) => (WrappedComponent) => ({
   isOpenPopover,
-  closePopover,
+  togglePopover,
+  isTogglePopover,
   setIsOpenPopover,
   closeNavbar,
   wrapperParentRef,
@@ -41,15 +42,23 @@ export const withPopover = ({
 
   useEffect(() => {
     if (isOpenPopover) {
-      setOpened(true);
       setIsOpenPopover?.(true);
     }
   }, [isOpenPopover, setIsOpenPopover]);
 
   const onClose = () => {
-    closePopover?.();
-    setIsOpenPopover?.(false);
+    setTimeout(() => togglePopover(false), 0);
     setOpened(false);
+    setIsOpenPopover?.(false);
+  };
+
+  const onClickHandle = () => {
+    if (isTogglePopover) {
+      togglePopover((prev) => !prev);
+    } else {
+      setOpened(true);
+    }
+    setIsOpenPopover?.(true);
   };
 
   const onCloseWrapperParentRef = () => {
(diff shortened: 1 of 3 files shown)


app/src/componentLibrary/sidebar/sidebar.jsx
```diff
@@ -32,20 +33,30 @@ export const Sidebar = ({
...
   const [isOpenNavbar, setIsOpenNavbar] = useState(false);
 
   const [actionButtonKey, setActionButtonKey] = useState(null);
   const [actionButtonType, setActionButtonType] = useState(null);
 
+  const sidebarRef = useRef(null);
+
+  const handleClickOutside = useCallback(() => {
+    if (isOpenNavbar) {
+      setIsOpenNavbar(false);
```
AmsterGet, line 48: `onCloseNavbar` can be used

app/src/componentLibrary/sidebar/sidebar.jsx
```diff
@@ -32,20 +33,30 @@ export const Sidebar = ({
   topSidebarItems,
   bottomSidebarItems,
   createFooterBlock,
+  isOpenPopover,
```
AmsterGet, line 36: It would be better not to rely on `popover` usage somewhere not in this component. It may be another event or smth that will not allow to close sidebar.
So I suggest renaming it to something more generic, like `shouldBeCollapsedOnLeave`. That way it will be more clear what happened when this prop provided.

app/src/componentLibrary/sidebar/sidebar.jsx
```diff
@@ -32,20 +33,30 @@ export const Sidebar = ({
...
+    }
+  }, [isOpenNavbar]);
+
+  useOnClickOutside(sidebarRef, handleClickOutside);
+
   const onOpenNavbar = () => {
     setIsOpenNavbar(true);
   };
 
   const onCloseSidebar = () => {
```
AmsterGet, line 58: This handler seems to be associated with `navbar` rather than `sidebar`.
Please rename it accordingly.
image

AmsterGet (changes requested):

app/src/componentLibrary/popover/withPopover/withPopover.jsx
```diff
@@ -26,40 +28,39 @@ export const withPopover = ({
...
-  }, [isOpenPopover, setIsOpenPopover]);
+  }, [togglePopover, setOpened]);
 
-  const onClose = () => {
-    closePopover?.();
-    setIsOpenPopover?.(false);
-    setOpened(false);
+  const onClickHandle = () => {
+    if (togglePopover) {
+      togglePopover((prev) => !prev);
```
AmsterGet, line 45: I'm not sure that this handler will always be state setter function, it can be just a callback that do some stuff internally.
So let's just call it here `togglePopover()` and leave the state changing logic in consumer's code.

AmsterGet (changes requested):

app/src/componentLibrary/popover/withPopover/withPopover.jsx
BlazarQSO: Done.

app/src/componentLibrary/sidebar/sidebar.jsx
BlazarQSO: Done.

app/src/componentLibrary/sidebar/sidebar.jsx
BlazarQSO: Done.

AmsterGet (approved):

PR #3819, opened 2024-05-02
EPMRPP-90680 || Popover is broken in sidebar

BlazarQSO:
## PR Checklist

### DIFF
--- a/app/src/layouts/common/appSidebar/appSidebar.scss
+++ b/app/src/layouts/common/appSidebar/appSidebar.scss
@@ -113,25 +113,6 @@
   }
 }
 
-.user-control {
-  width: 100%;
-  box-sizing: border-box;
-  padding: 0 8px;
-  cursor: pointer;
-
-  &:hover {
-    background-color: $COLOR--darkmode-gray-500;
-
-    button div {
-      color: $COLOR--white-two;
-
-      path {
-        fill: $COLOR--white-two;
-      }
-    }
-  }
-}
-
 .hover {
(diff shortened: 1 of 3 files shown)


app/src/layouts/common/appSidebar/appSidebar.scss
```diff
@@ -121,6 +121,14 @@
...
+
+    button div {
+      color: $COLOR--white-two;
+
+      path {
+        fill: $COLOR--white-two;
+      }
+    }
   }
 }
```
AmsterGet, line 133: Looks as a duplicate of lines 135-147.
Can in be handled via single set of such styles?

Another comment is about using relative tags selectors (button > div).
Can we avoid them here in favor of classes to less depend on child components HTML structure?

BlazarQSO: 1 Done.
2 These changes will be in the component from the component library, so it's better to leave it that way.

app/src/layouts/common/appSidebar/userControl/userControl.scss
```diff
@@ -103,9 +92,11 @@
 
 .popover-control {
   height: 48px;
-  width: 100%;
+  width: 264px;
```
Vadim73i, line 95: Is it ok to use fixed width?

BlazarQSO: A fixed width is used

Vadim73i (approved):

PR #3828, opened 2024-05-08
EPMRPP-90888 ||  Possible to create 2 Slack notification rules

maria-hambardzumian:
## PR Checklist


app/src/pages/inside/projectSettingsPageContainer/content/notifications/utils.js
```diff
@@ -17,6 +17,15 @@
...
 
+function getUpdatedProperties(base, updated) {
+  return Object.keys(base).reduce((acc, key) => {
+    if (updated[key] && base[key] !== updated[key]) {
+      acc[key] = updated[key];
+    }
+    return acc;
+  }, base);
+}
+
```
AmsterGet, line 28: Looks like we've come to this complication because of the:
1. We didn't specify the `ruleDetails` fields under the `ruleDetails` object within the form.

We can use
```javascript
<FieldElement
    name={`${RULE_DETAILS_FIELD_KEY}.${field.name}`}
    // ...
/>
```
where `RULE_DETAILS_FIELD_KEY = 'ruleDetails'`.
Instead of
```javascript
<FieldElement
    name={field.name}
    // ...
/>
```
in the `modals/addEditNotificationModal/addEditNotificationModal.jsx, line 372`. In this case the dynamic fields will be stored within `ruleDetails` object. Don't forget to adjust validation if needed.

2. We've added `ruleFields` to each rule as its own field that caused to form initialization with this value.

image

Here I suggest to add a new prop `ruleItemContentProps` or `ruleContentProps` to the `RuleList` component to destructure its value while rendering `ruleItemContent`.
In this case we can put `ruleFields` there and it will be available within `NotificationRuleContent` and won't be initialized within the form.

image

I suppose after that changes additional processing within submission preparation won't be required.

AmsterGet (changes requested):

app/src/pages/inside/projectSettingsPageContainer/content/elements/ruleList/ruleList.jsx
```diff
@@ -44,7 +45,9 @@ const PlainRuleList = ({
           actions={actions}
           onToggle={onToggle}
           disabled={disabled}
-          content={ruleItemContent && <Content item={item} />}
+          content={
+            ruleItemContent && <Content item={item} ruleItemContentProps={ruleItemContentProps} />
```
AmsterGet, line 49: I suggest to use shorter name within Content component for its props or even spread them as the verbose explanation for `ruleItemContentProps` is not required inside it.
```javascript
ruleItemContent && <Content item={item} contentProps={ruleItemContentProps} />
// or even better
ruleItemContent && <Content item={item} {...ruleItemContentProps} />
```

app/src/pages/inside/projectSettingsPageContainer/content/notifications/ruleGroup/ruleGroup.jsx
```diff
@@ -304,11 +304,11 @@ export const RuleGroup = ({ pluginName, ruleDescription, rules, isPluginEnabled,
                 data={rules.map((rule) => ({
                   name: rule.ruleName,
                   ...rule,
-                  ruleFields,
                 }))}
                 actions={actions}
                 onToggle={onToggleHandler}
                 ruleItemContent={NotificationRuleContent}
+                ruleItemContentProps={ruleFields}
```
AmsterGet, line 311: Following the comment above it will be 
```suggestion
                ruleItemContentProps={{ ruleFields }}
```

app/src/pages/inside/projectSettingsPageContainer/content/elements/ruleList/propTypes.js
```diff
@@ -1,11 +1,13 @@
 import PropTypes from 'prop-types';
+import { ruleField } from 'pages/inside/projectSettingsPageContainer/content/notifications/propTypes';
 
 export const ruleListPropTypes = {
   items: PropTypes.array,
   actions: PropTypes.array,
   onToggle: PropTypes.func,
   disabled: PropTypes.bool,
   ruleItemContent: PropTypes.elementType,
+  ruleItemContentProps: PropTypes.arrayOf(ruleField),
```
AmsterGet, line 10: The `RuleList` component is widely used within settings (not only for notifications), so we need to have a more abstract type here, like ` PropTypes.object`.

AmsterGet (approved):

PR #3845, opened 2024-05-27
EPMRPP-90078 || Roles and permissions refactoring

BlazarQSO:
## PR Checklist


app/src/common/constants/projectRoles.js
```diff
@@ -20,6 +20,11 @@ export const CUSTOMER = 'CUSTOMER';
 export const OPERATOR = 'OPERATOR';
```
AmsterGet, line 20: Can we remove deprecated roles?

AmsterGet (changes requested):

app/src/common/constants/projectRoles.js
BlazarQSO: The old roles are used in components that are out of scope. They will be deleted later.

PR #3863, opened 2024-06-10
EPMRPP-89528 || Sidebar organization level

BlazarQSO:
## PR Checklist


maria-hambardzumian (approved):

app/src/layouts/appLayout/appLayout.jsx
```diff
@@ -18,30 +18,51 @@ import PropTypes from 'prop-types';
...
-  <Layout
-    Header={AppHeader}
-    Sidebar={ProjectSidebar}
-    Banner={isDemoInstance ? DemoBanner : null}
-    rawContent={rawContent}
-  >
-    {children}
-  </Layout>
-);
+const AppLayoutComponent = ({ children, isDemoInstance, rawContent, sidebarType }) => {
```
AmsterGet, line 27: What do you think about creating the own Layout component for each level instead of mapping the sidebar inside of it? Like it is done now for admin part (AdminLayout)?
In this case it will be easier to make changes to each level-specific Layout without a risk of touching others.
And some logic that is specific for each level can be incapsulated inside each Layout (like case with DemoBanner that is a part of a project level only for now).

PR #3916, opened 2024-07-11
EPMRPP-92607 || User with different roles across different Organizati…

BlazarQSO:
…ons see pages in the same mode

## PR Checklist


maria-hambardzumian (approved):

app/src/controllers/organizations/organization/selectors.js
```diff
@@ -24,3 +30,20 @@ export const activeOrganizationLoadingSelector = (state) =>
   organizationSelector(state).organizationLoading || false;
 
 export const activeOrganizationNameSelector = (state) => activeOrganizationSelector(state)?.name;
+
+export const activeOrganizationRolesSelector = createSelector(
```
AmsterGet, line 34: Looks like this selector too similar with `userRolesSelector`.
Can we leave just one?
Do we really need to select user roles from different sources?
Can we just extract the slugs from the URL and define org/project user role based on `assignedOrganizations` and `assignedProjects`. For some cases they may be empty (for admin/manager viewing the project/org without assignment).

AmsterGet (changes requested):

app/src/controllers/organizations/organization/selectors.js
BlazarQSO: Done.

app/src/controllers/pages/selectors.js
```diff
@@ -159,3 +167,74 @@ export const urlOrganizationAndProjectSelector = createSelector(
     return activeProject;
   },
 );
+
+export const activeProjectRoleSelector = createSelector(
```
AmsterGet, line 171: Roles are more related to the user entity rather than pages, so maybe still store them in `controllers/user`?

AmsterGet (changes requested):

app/src/controllers/pages/selectors.js
```diff
@@ -159,3 +167,74 @@ export const urlOrganizationAndProjectSelector = createSelector(
...
+  activeProjectRoleSelector,
+  (userRole, organizationRole, projectRole) => ({
+    userRole,
+    organizationRole,
+    projectRole,
+  }),
+);
+
+export const createUserAssignedSelector = (projectSlug, organizationSlug) =>
+  createSelector(
```
AmsterGet, line 203: `createSelector` redundant here as you call `createUserAssignedSelector` each time you use it and it calls all the selectors from scratch without any memorization, so please remove `createSelector` usage here.

AmsterGet (changes requested):

AmsterGet (approved):

PR #3923, opened 2024-07-18
EPMRPP-89421 || Display number of tests related to each Unique Error cluster

maria-hambardzumian:
## PR Checklist

### DIFF
--- a/app/src/controllers/uniqueErrors/sagas.js
+++ b/app/src/controllers/uniqueErrors/sagas.js
@@ -17,13 +17,7 @@
 import { all, call, put, select, take, takeEvery } from 'redux-saga/effects';
 import { URLS } from 'common/urls';
 import { activeProjectSelector } from 'controllers/user';
-import {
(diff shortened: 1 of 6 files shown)


BlazarQSO (approved):

app/src/controllers/uniqueErrors/sagas.js
```diff
@@ -84,9 +78,7 @@ function* fetchClusters(payload = {}) {
     yield call(fetchParentLaunch, { payload: { project, launchId } });
   }
 
-  const namespace = yield select(namespaceSelector);
-  const query = yield select(queryParametersSelector, namespace);
-
+  const query = yield select(queryParametersSelector);
```
AmsterGet, line 81: Now the pagination doesn't work for Unique errors page.
image
In case you use separate query parameters entity for this page, it should be namespaced to not be mixed with query parameters from previous page (you can use `uniqueErrors` namespace, similar logic done for log pages).

PR #3935, opened 2024-07-24
EPMRPP-91529 || Launch merge. Incorrect limitation for value in the 'Description' field in the 'Merge Launches' modal

maria-hambardzumian:
## PR Checklist


app/src/pages/inside/launchesPage/modals/launchMergeModal/launchMergeModal.jsx
```diff
@@ -335,7 +335,7 @@ export class LaunchMergeModal extends Component {
               label={intl.formatMessage(messages.launchDescriptionLabel)}
               labelWidth={FIELD_LABEL_WIDTH}
             >
-              <FieldProvider name="description" maxLength={'1024'}>
+              <FieldProvider name="description" maxLength={'2048'}>
```
AmsterGet, line 338: Could you please check the field behavior?
I see that setting the `maxLength` attribute doesn't allow to insert more characters into the field and prevent the validation.
In addition, this may result in loss of user data if inserted using hotkeys.

AmsterGet (changes requested):

app/src/pages/inside/launchesPage/modals/launchMergeModal/launchMergeModal.jsx
maria-hambardzumian: Actually, that doesn't allow typing. The message only appears for auto-generated texts when the length is more than 2048 (the same behavior was observed with 1024). Should it be removed?

AmsterGet: We discussed this topic with PO again, and decided to just increase the limit for now.
But as a **future** improvement we will consider removing the `maxLenght` limit, the field limit will be controlled by validation only.

PR #3943, opened 2024-08-01
EPMRPP-88780 || Increase the dashboard limit per project

Vadim73i:
## PR Checklist

### DIFF
--- a/app/src/controllers/dashboard/sagas.js
+++ b/app/src/controllers/dashboard/sagas.js
@@ -45,8 +45,6 @@ import {
   UPDATE_DASHBOARD,
   UPDATE_DASHBOARD_WIDGETS,
   REMOVE_DASHBOARD_SUCCESS,
-  INCREASE_TOTAL_DASHBOARDS_LOCALLY,
-  DECREASE_TOTAL_DASHBOARDS_LOCALLY,
 } from './constants';
 import { dashboardItemsSelector, querySelector } from './selectors';
 import {
@@ -62,8 +60,8 @@ function* fetchDashboards({ payload: params }) {
   yield put(
     fetchDataAction(NAMESPACE)(URLS.dashboards(activeProject), {
       params: {
-        ...query,
         ...params,
+        ...query,
       },
     }),
   );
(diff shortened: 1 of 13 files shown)


app/src/controllers/dashboard/selectors.js
```diff
@@ -40,4 +41,6 @@ export const totalDashboardsSelector = (state) =>
 
 export const dashboardFullScreenModeSelector = (state) => domainSelector(state).fullScreenMode;
 
-export const querySelector = createQueryParametersSelector();
```
AmsterGet, line 43: Do we need to revert the changes from https://jiraeu.epam.com/browse/EPMRPP-81029 or adjust them somehow?

app/src/controllers/dashboard/selectors.js
```diff
@@ -40,4 +41,6 @@ export const totalDashboardsSelector = (state) =>
...
+export const querySelector = createQueryParametersSelector({ namespace: NAMESPACE });
```
AmsterGet, line 44: Do we need a `namespace` here?
In general the `namespace`s are used when there are several types of query params for different entities (like we can have several test item params & log params) simultaneously. In this case (like the similar one for filters, members, etc.) the only one set of query params possible - only dashboards. So, let's remove the namespace usage for query params if its redundant.

app/src/pages/inside/dashboardPage/dashboardPage.jsx
```diff
@@ -89,6 +93,10 @@ const messages = defineMessages({
     addDashboard: addDashboardAction,
   },
 )
+@withPagination({
+  paginationSelector: dashboardPaginationSelector,
+  namespace: NAMESPACE,
+})
```
AmsterGet, line 99: Pagination controls does not have any effect on requests. Pls recheck this case.

AmsterGet: Also with your changes the search is disabled- it cannot be edited after typing the value.
The empty state is also wrong.
Your version.
image
Develop version.
image

AmsterGet (changes requested):

app/src/controllers/dashboard/sagas.js
```diff
@@ -162,7 +159,6 @@ function* removeDashboard({ payload: id }) {
       method: 'delete',
     });
     yield put(deleteDashboardSuccessAction(id));
-    yield put({ type: DECREASE_TOTAL_DASHBOARDS_LOCALLY });
```
AmsterGet, line 165: Why did you remove these actions usage while left the validation on 'create new dashboard button'?

AmsterGet (changes requested):

AmsterGet (approved):

PR #3995, opened 2024-09-02
EPMRPP-91645 || Search for projects on Projects page

maria-hambardzumian:
## PR Checklist


BlazarQSO (dismissed):

app/src/components/filterEntities/containers/filterEntitiesURLContainer.jsx
```diff
@@ -14,54 +14,66 @@
...
-  };
+  const debouncedHandleChange = useCallback(debounce(handleChange, debounceTime), [
+    handleChange,
+    debounceTime,
+  ]);
 
-  debouncedHandleChange = debounce(this.handleChange, 1000);
+  return render({
+    entities,
+    onChange: debounceTime ? debouncedHandleChange : handleChange,
```
AmsterGet, line 53: `debounceTime` will be treated always as `true` by default, so I suggest to leave the `debounced` prop to control this behavior. It still can be true by default.

app/src/pages/organization/organizationProjectsPage/messages.js
```diff
@@ -78,4 +78,13 @@ export const messages = defineMessages({
     id: 'ProjectsPage.projectNamePlaceholder',
     defaultMessage: 'Enter project’s name',
   },
+  noResultsLabel: {
+    id: 'ProjectsPage.noResultsLabel',
+    defaultMessage: 'No results found',
```
AmsterGet, line 83: Please use a message from `src/common/constants/localization.js`.

app/src/pages/organization/organizationProjectsPage/projectsPageHeader/projectsPageHeader.jsx
```diff
@@ -77,7 +94,25 @@ export const ProjectsPageHeader = ({ hasPermission, onCreateProject }) => {
         <div className={cx('actions')}>
           {isNotEmpty && (
             <div className={cx('icons')}>
-              <i className={cx('search-icon')}>{Parser(searchIcon)}</i>
+              <SearchEntitiesURLContainer
```
AmsterGet, line 97: Can we use `withFilter` HOC instead from `src/controllers/filter/withFilter.jsx` as it is done for FiltersPage? (src/pages/inside/filtersPage/filtersPage.jsx).

app/src/controllers/organizations/projects/selectors.js
```diff
@@ -38,8 +38,8 @@ export const createOrganizationProjectsParametersSelector = ({
       defaultSorting,
       sortingKey,
     }),
-    ({ [SIZE_KEY]: limit, [SORTING_KEY]: sort, [PAGE_KEY]: pageNumber }) => {
-      return getAlternativePaginationAndSortParams(sort, limit, pageNumber);
+    ({ [SIZE_KEY]: limit, [SORTING_KEY]: sort, [PAGE_KEY]: pageNumber, [SEARCH_KEY]: name }) => {
+      return getAlternativePaginationAndSortParams(sort, limit, pageNumber, name);
```
AmsterGet, line 42: Let's not mutate the `getAlternativePaginationAndSortParams` to support filter fields. This function is intended just to convert the new pagination and sort format to unify the approach with old APIs. So if you need to propagate all the rest of query parameters without changes, just return them as a part of the result object, e.g.
```suggestion
    ({ [SIZE_KEY]: limit, [SORTING_KEY]: sort, [PAGE_KEY]: pageNumber, ...rest }) => {
      return { ...getAlternativePaginationAndSortParams(sort, limit, pageNumber), ...rest };
```

AmsterGet (changes requested):

app/src/pages/organization/organizationProjectsPage/organizationProjectsPage.jsx
```diff
@@ -49,6 +52,7 @@ export const OrganizationProjectsPage = () => {
 
   const projects = useSelector(projectsSelector);
   const isProjectsEmpty = !projectsLoading && projects.length === 0;
+  const [searchValue, setSearchValue] = useState(null);
```
AmsterGet, line 55: Can we avoid duplicating the filter value in state?
I suppose we can use it directly from `filter` props provided via `withFilter` HOC.

app/src/components/filterEntities/containers/filterEntitiesURLContainer.jsx
```diff
@@ -14,54 +14,63 @@
...
```
AmsterGet, line 52: Let's return the `debounced` boolean prop with `true` by default in addition to `debounceTime` to use it here to avoid specifying `debounceTime={0}` in most cases.

AmsterGet (changes requested):

app/src/pages/organization/organizationProjectsPage/organizationProjectsPage.jsx
maria-hambardzumian: Used searchValueState from the parent component to specify what kind of empty state there is for content: 'No results found' or 'No project'

AmsterGet: Got it.
My point I that you can use the HOC on the page level component (parent) to rely on the `filter` prop and propagate it to the children where needed.

AmsterGet: Decided to leave it as is to avoid text typing debounce.

AmsterGet (approved):

PR #3999, opened 2024-09-03
enable viewing video attachments

LyubomirStoimchev:
## PR Checklist

## Visuals
https://github.com/user-attachments/assets/de0c3b0f-8019-4635-a5dd-d0fd16e9f874


LyubomirStoimchev:
Most of the solution is copied and modified from the Image viewer modal. This solution uses the HTML5 video implementation.

The only thing that I've found so far is that there is not "Video" attachment image. I have no idea how to exactly generate them, that is why I used the HTML image.

AmsterGet:
Hello @LyubomirStoimchev !
Thank you very much for your suggestion and Pull request!
Having such a feature in the app will be very helpful for our users.
I have discussed your suggestion with the team and we have created a ticket in our backlog. As part of this task, we will review the PR soon and provide feedback.
At least, we need to decide what other types of videos we should be able to play in addition to `mp4`.

PR #4019, opened 2024-09-22
EPMRPP-95440 || The filter name is displayed in the breadcrumbs when …

BlazarQSO:
…clicking Filter and Clear filter buttons

## PR Checklist

### DIFF
--- a/app/src/controllers/filter/sagas.js
+++ b/app/src/controllers/filter/sagas.js
@@ -36,7 +36,11 @@ import { FETCH_PROJECT_PREFERENCES_SUCCESS } from 'controllers/project/constants
 import { launchDistinctSelector } from 'controllers/launch/selectors';
 import { fetchLaunchesAction } from 'controllers/launch/actionCreators';
 import { omit } from 'common/utils/omit';
-import { NEW_FILTER_PREFIX, CUSTOM_FILTER } from 'common/constants/reservedFilterIds';
+import {
+  NEW_FILTER_PREFIX,
+  CUSTOM_FILTER,
+  CUSTOM_FILTER_ID,
+} from 'common/constants/reservedFilterIds';
 import { redirect } from 'redux-first-router';
 import {
   NAMESPACE,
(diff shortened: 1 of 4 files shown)


app/src/controllers/filter/sagas.js
```diff
@@ -127,24 +127,32 @@ function* resetFilter({ payload: filterId }) {
...
   const launchFilters = yield select(launchFiltersSelector);
   const userId = yield select(userIdSelector);
   const lastNewFilterId = launchFilters.reduce(
     (acc, launchFilter) => (launchFilter.id < acc ? launchFilter.id : acc),
     0,
   );
+
+  const filterName = hasFilterPermissions
+    ? `${NEW_FILTER_PREFIX} ${-(lastNewFilterId - 1)}`
+    : 'Custom filter';
```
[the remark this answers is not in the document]
BlazarQSO: Done.

maria-hambardzumian (approved):

app/src/controllers/filter/sagas.js
```diff
@@ -127,24 +131,32 @@ function* resetFilter({ payload: filterId }) {
...
-      ? `${COPY_PREFIX} ${filter.name}`
-      : `${NEW_FILTER_PREFIX} ${-(lastNewFilterId - 1)}`,
+    id: hasFilterPermissions ? lastNewFilterId - 1 : CUSTOM_FILTER_ID,
+    name: filter.name ? `${COPY_PREFIX} ${filter.name}` : filterName,
     owner: userId,
   };
   yield put(addFilterAction(newFilter));
   yield put(changeActiveFilterAction(newFilter.id, meta));
+
+  if (isRemoveCustomFilter) {
```
AmsterGet, line 157: Can we rely only on `hasFilterPermissions` here?
Looks like without permissions for filter it is always expected to remove filter immediately after create, isn't it?

AmsterGet (changes requested):
I suggest to simplify the solution.
1. Can we just point the user to the filter with id `-1` (in fact we need a single unsaved filter for that, and it will always with id `-1`) if it exists by clicking on `Filter` button (just update the action for this button in case of insufficient permissions). Otherwise just create it.
2. Update the place where we show the filter name. If user has permissions, just display 'Custom filter'.
image

app/src/controllers/project/sagas.js
```diff
@@ -430,9 +432,15 @@ function* watchFetchProject() {
...
+  const userRoles = yield select(userRolesSelector);
+  const hasFilterPermissions = canWorkWithFilters(userRoles);
+
   const preferences = yield call(fetch, URLS.projectPreferences(projectKey));
   yield put(fetchProjectPreferencesSuccessAction(preferences));
-  yield put(fetchUserFiltersSuccessAction(preferences.filters));
+
+  if (hasFilterPermissions) {
+    yield put(fetchUserFiltersSuccessAction(preferences.filters));
+  }
```
AmsterGet, line 443: In case you are willing to disable the user filter preferences setting to the redux, does it make sense to not send a request for them at all?
I suppose it would be ok to check where `fetchProjectPreferences` called and do not trigger it in case of insufficient permissions.

BlazarQSO: Done.

AmsterGet (approved):

maria-hambardzumian (approved):

PR #4027, opened 2024-09-30
EPMRPP-93646 || GA events. Sidebar part 1

maria-hambardzumian:
## PR Checklist

### DIFF
--- a/app/src/componentLibrary/sidebar/sidebar.jsx
+++ b/app/src/componentLibrary/sidebar/sidebar.jsx
@@ -23,6 +23,8 @@ import styles from './sidebar.scss';
 
 const cx = classNames.bind(styles);
 
+const COLLAPSED_WIDTH = 48;
+
 export const Sidebar = ({
   logoBlock,
   createMainBlock,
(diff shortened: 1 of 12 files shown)


app/src/componentLibrary/sidebar/sidebar.jsx
```diff
@@ -66,7 +70,7 @@ export const Sidebar = ({
       <aside className={cx('sidebar')}>
         {logoBlock}
         <div className={cx('main-block-wrapper')}>
-          {createMainBlock(onOpenSidebar, onCloseSidebar)}
+          {createMainBlock(onOpenSidebar, onCloseSidebar, getIsSidebarCollapsed)}
```
AmsterGet, line 73: Why not just use the `isOpenSidebar` instead?

AmsterGet (changes requested):

app/src/componentLibrary/sidebar/sidebar.jsx
maria-hambardzumian: isOpenSidebar is always set to true on the onClick event, as it is also set to true on onMouseEnter. 
The sidebar expands due to the CSS transition delay

AmsterGet: Ok, got it.

AmsterGet (approved):

PR #4050, opened 2024-10-15
Feature/epmrpp 91636 search for organizations

BlazarQSO:
## PR Checklist


app/src/components/fields/searchField/searchField.jsx
```diff
@@ -0,0 +1,111 @@
...
+ * limitations under the License.
+ */
+
+import { useEffect, useState } from 'react';
+import PropTypes from 'prop-types';
+import Parser from 'html-react-parser';
+import { useTracking } from 'react-tracking';
+import { FieldText } from '@reportportal/ui-kit';
+import classNames from 'classnames/bind';
+import searchIcon from 'common/img/newIcons/search-outline-inline.svg';
```
maria-hambardzumian, line 23: Can we use search-icon from ui-kit?

BlazarQSO: Done.

app/src/components/fields/searchField/searchField.jsx
```diff
@@ -0,0 +1,109 @@
...
+  event: PropTypes.object,
+};
+
+SearchFieldWrapped.defaultProps = {
+  placeholder: '',
+  isLoading: false,
+  event: null,
+};
+
+export const SearchField = withFilter({ filterKey: SEARCH_KEY })(SearchFieldWrapped);
```
AmsterGet, line 109: That's good idea to make this search component reusable, just want to hightlight two things:
1. We have a `components` directory mainly used for outdated components that are not aligned with the design system.
So does it make sense to create a separate space for holding such reusable stuff for new designs to avoid mixing them?
2. I suggest to use the `withFilter` HOC on demand. Not as a part of reusable component.
There might be cases for using this component with local state or with another `filterKey`.

AmsterGet (changes requested):

PR #4059, opened 2024-10-22
EPMRPP-96401 || Added space for pagination in scrollWrapper

maria-hambardzumian:
## PR Checklist

### DIFF
--- a/app/src/components/main/scrollWrapper/scrollWrapper.jsx
+++ b/app/src/components/main/scrollWrapper/scrollWrapper.jsx
@@ -50,6 +50,7 @@ export class ScrollWrapper extends Component {
     thumbMinSize: PropTypes.number,
     withBackToTop: PropTypes.bool,
     withFooter: PropTypes.bool,
+    withPaginationOffset: PropTypes.bool,
     resetRequired: PropTypes.bool,
     onReset: PropTypes.func,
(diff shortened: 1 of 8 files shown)


app/src/components/main/scrollWrapper/scrollWrapper.jsx
```diff
@@ -50,6 +50,7 @@ export class ScrollWrapper extends Component {
     thumbMinSize: PropTypes.number,
     withBackToTop: PropTypes.bool,
     withFooter: PropTypes.bool,
+    withPaginationOffset: PropTypes.bool,
```
AmsterGet, line 53: May be ads this as a number to specify the 'contentMarginBottom'?
Or even better add the 'contentClassName' prop to provide any styles.

AmsterGet (changes requested):

app/src/components/main/scrollWrapper/scrollWrapper.jsx
maria-hambardzumian: Wouldn't it be better to keep contentMarginBottom, since it's used in many places with pagination to avoid repeating the same style? I can also add contentClassName for any additional specific styles

BlazarQSO (approved):

PR #4063, opened 2024-10-23
EPMRPP-96352 || Routing is broken for Member user

BlazarQSO:
## PR Checklist

### DIFF
--- a/app/src/controllers/pages/selectors.js
+++ b/app/src/controllers/pages/selectors.js
@@ -210,24 +210,26 @@ export const userAssignedSelector = (projectSlug, organizationSlug) => (state) =
   const isAdmin = userRole === ADMINISTRATOR;
   const isManager = organizationRole === MANAGER;
   let isAssignedToTargetOrganization = false;
+  const projectKey = `${organizationSlug}.${projectSlug}`;
+  const project = assignedProjects[projectSlug] || assignedProjects[projectKey];
 
   if (organizationSlug) {
     isAssignedToTargetOrganization = organizationSlug in assignedOrganizations;
   } else {
-    const organizationId = assignedProjects[projectSlug]?.organizationId || '';
+    const organizationId = project?.organizationId || '';
+
     isAssignedToTargetOrganization = Object.keys(assignedOrganizations).some(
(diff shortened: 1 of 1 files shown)


app/src/controllers/pages/selectors.js
```diff
@@ -210,24 +210,26 @@ export const userAssignedSelector = (projectSlug, organizationSlug) => (state) =
   const isAdmin = userRole === ADMINISTRATOR;
   const isManager = organizationRole === MANAGER;
   let isAssignedToTargetOrganization = false;
+  const projectKey = `${organizationSlug}.${projectSlug}`;
```
AmsterGet, line 213: I'm not sure the `projectKey` will be always look like this.
Can we rely on slugs here?

app/src/controllers/pages/selectors.js
```diff
@@ -210,24 +210,26 @@ export const userAssignedSelector = (projectSlug, organizationSlug) => (state) =
...
+  const project = assignedProjects[projectSlug] || assignedProjects[projectKey];
```
AmsterGet, line 214: Why the `assignedProjects` isn't consistent and may contain entities by slug and by key?
Does it make sense to discuss it with backend?

app/src/controllers/pages/selectors.js
BlazarQSO: The slugs can be the same for projects. The uniqueness of the projects is ensured only by the project key.
This is done for backward compatibility. But then helen.helen is expected
image

app/src/controllers/pages/selectors.js
BlazarQSO: This probably needs to be discussed. But for now, most of the projects won't work without it.

app/src/controllers/pages/selectors.js
BlazarQSO: Done.

app/src/controllers/pages/selectors.js
BlazarQSO: Done.

AmsterGet (approved):

PR #4132, opened 2024-12-16
EPMRPP-97900 || Improve event "page_view" for Google Analytics 4

iso9000t:
## PR Checklist


app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
```diff
@@ -66,20 +67,28 @@ export const ProjectSettingsPageContainer = () => {
...
         [extension.name]: {
           name: extension.title || extension.name,
-          link: createTabLink(extension.name),
+          link: hasInitialPage
+            ? createTabLink(
+                extension.name,
+                { subTab: extension.initialPage?.payload?.subTab },
+                extension.initialPage?.type,
+              )
+            : createTabLink(extension.name),
```
AmsterGet, line 91: Can we simplify this?
In case the `initialPage?.payload` or `initialPage?.type` will be empty while passing them to the `createTabLink` function, the default values will be used.
So,
```suggestion
          link: createTabLink(extension.name, extension.initialPage?.payload, extension.initialPage?.type),
```

AmsterGet (changes requested):

app/src/pages/inside/projectSettingsPageContainer/projectSettingsPageContainer.jsx
iso9000t: Good point! This hasInitialPage function was added to explicitly check the data for readability, but you are right, the code looks overcomplicated with it, so I removed it.

AmsterGet (approved):

PR #4161, opened 2025-01-15
EPMRPP-98470 || Fix incorrect User Type values on "All Users" page at…

iso9000t:
## PR Checklist

## Visuals

image

### DIFF
--- a/app/src/pages/instance/allUsersPage/allUsersListTable/allUsersListTable.jsx
+++ b/app/src/pages/instance/allUsersPage/allUsersListTable/allUsersListTable.jsx
@@ -25,7 +25,6 @@ import { userInfoSelector } from 'controllers/user';
 import { getRoleBadgesData } from 'common/utils/permissions/getRoleTitle';
 import { NAMESPACE } from 'controllers/instance/allUsers/constants';
 import { UserNameCell } from 'pages/common/membersPage/userNameCell/userNameCell';
-
 import {
   DEFAULT_PAGE_SIZE,
   DEFAULT_PAGINATION,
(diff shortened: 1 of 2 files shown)


app/src/pages/instance/allUsersPage/allUsersListTable/allUsersListTable.jsx
```diff
@@ -43,6 +42,23 @@ import { messages } from 'pages/common/users/membersListTable/messages';
...
+    return 'Synched';
+  }
+
+  const typeMap = {
+    INTERNAL: 'Internal',
+    GITHUB: 'GitHub',
+    SAML: 'SAML',
+    LDAP: 'LDAP',
+    SCIM: 'SCIM',
+  };
```
AmsterGet, line 57: I'm not sure it is necessary to create this map on every function call.
Let's move it to the constants.
And also the such a mapping for the same type values isn't required as you use the `accountType` as is as a fallback.
image

app/src/pages/instance/allUsersPage/allUsersListTable/allUsersListTable.jsx
```diff
@@ -43,6 +42,23 @@ import { messages } from 'pages/common/users/membersListTable/messages';
 import styles from './allUsersListTable.scss';
 
 const cx = classNames.bind(styles);
+
+const getDisplayAccountType = (accountType) => {
+  if (accountType === 'UPSA') {
+    return 'Synched';
+  }
```
AmsterGet, line 49: Why is this option not included in `typeMap`?

AmsterGet (changes requested):

app/src/pages/instance/allUsersPage/allUsersListTable/allUsersListTable.jsx
iso9000t: Yeah, I used the fallback values when needed and since we might want to use these types somewhere else, put the map into the constants file. which also helped avoid recalculations!

AmsterGet (approved):

PR #4201, opened 2025-02-17
EPMRPP-99338 || Buttons Apply and Cancel in the filter popup are avai…

BlazarQSO:
…lable after changing the conditions only

## PR Checklist

### DIFF
--- a/app/src/components/main/filterButton/filterContent/filterContent.jsx
+++ b/app/src/components/main/filterButton/filterContent/filterContent.jsx
@@ -15,13 +15,14 @@
  */
 
 import classNames from 'classnames/bind';
-import { reduxForm } from 'redux-form';
+import { getFormValues, reduxForm } from 'redux-form';
+import { useSelector } from 'react-redux';
 import { useTracking } from 'react-tracking';
 import { Button } from '@reportportal/ui-kit';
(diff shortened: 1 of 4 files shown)


maria-hambardzumian (approved):

app/src/components/main/filterButton/filterContent/filterContent.jsx
```diff
@@ -35,22 +36,31 @@ export const FilterContentWrapped = ({
...
-  submitting,
   handleSubmit,
+  getClearButtonState,
+  getApplyButtonState,
   event,
 }) => {
   const { formatMessage } = useIntl();
   const { trackEvent } = useTracking();
-  const isDisabled = pristine || submitting;
+  const formValues = useSelector((state) => getFormValues('filter')(state));
```
AmsterGet, line 49: So, why not to use redux-form props to define necessary form stage?

AmsterGet (changes requested):

app/src/components/main/filterButton/filterContent/filterContent.jsx
AmsterGet: Desided to leave it as is.

PR #4212, opened 2025-02-20
EPMRPP-99318 || Add event for Login page

BlazarQSO:
## PR Checklist

### DIFF
--- a/app/src/components/main/analytics/events/ga4Events/loginPageEvents.js
+++ b/app/src/components/main/analytics/events/ga4Events/loginPageEvents.js
@@ -2,6 +2,10 @@ import { getBasicClickEventParameters } from '../common/ga4Utils';
 
 const LOGIN_PAGE = 'login_page';
 
+export const GIT_HUB = 'github';
+export const LOGIN = 'login';
+export const SAML_PROVIDERS = 'samlProviders';
+
 export const LOGIN_PAGE_EVENTS = {
   clickOnSocialIcon: (iconType) => {
     return {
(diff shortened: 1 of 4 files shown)


app/src/components/main/analytics/events/ga4Events/loginPageEvents.js
```diff
@@ -15,4 +19,30 @@ export const LOGIN_PAGE_EVENTS = {
...
+        break;
+      }
+      case LOGIN: {
+        elementName = LOGIN;
+        break;
+      }
+      default: {
+        elementName = 'login_with_epam';
+      }
+    }
```
AmsterGet, line 41: I suggest to not hardcode this values:
1. We can extract their names from the data come from backend.
2. We also have LDAP integration.
3. It's not robust to have `login_with_epam` as a default value.

app/src/pages/outside/loginPage/pageBlocks/loginBlock/loginForm/externalLoginBlock/externalLoginBlock.jsx
```diff
@@ -44,14 +54,23 @@ export class ExternalLoginBlock extends PureComponent {
     setWindowLocationToNewPath(normalizePathWithPrefix(path));
   };
 
+  clickEventHandler = () => {
+    const { tracking } = this.props;
+    tracking.trackEvent(LOGIN_PAGE_EVENTS.clickOnLoginButton(GIT_HUB));
```
AmsterGet, line 59: Why the GitHub hardcoded here?
Any of the external login can be used here.

AmsterGet (changes requested):

app/src/components/main/analytics/events/ga4Events/loginPageEvents.js
BlazarQSO: I didn't find such data from the backend.
The task does not say to take into account other integrations

PR #4279, opened 2025-03-24
EPMRPP-101005 || Launch name is not shown in group

maria-hambardzumian:
## PR Checklist


app/src/controllers/testItem/selectors.js
```diff
@@ -100,6 +100,8 @@ export const isTestItemsListSelector = createSelector(
 
 export const searchedTestItemsSelector = (state) => domainSelector(state).searchedItems;
 
+export const isSearchedItemsSelector = (state) => !isEmptyObject(searchedTestItemsSelector(state));
```
AmsterGet, line 103: This sounds more like `whether this selector is a searched items selector?`, so please rename to `isSearchWidgetItemsExistSelector` or smth like this.

AmsterGet (changes requested):
Pls also update the target branch according to ticket label.

PR #4287, opened 2025-03-27
EPMRPP-100277 || Configure session expiration per instance

maria-hambardzumian:
## PR Checklist

* [ ] Have you added the link to the PR to the Jira ticket description?


maria-hambardzumian:
https://jiraeu.epam.com/browse/EPMRPP-100277

app/src/pages/admin/serverSettingsPage/serverSettingsTabs/authConfigurationTab/forms/inactivityTimeoutForm/inactivityTimeoutForm.jsx
```diff
@@ -0,0 +1,89 @@
...
+  },
+});
+
+export const InactivityTimeoutForm = () => {
+  const { formatMessage } = useIntl();
+  const { trackEvent } = useTracking();
+  const expirationTimeoutConfig = useSelector(sessionExpirationTimeSelector);
+  const dispatch = useDispatch();
+
+  const [timeout, setTimeout] = useState(expirationTimeoutConfig);
```
AmsterGet, line 52: I suggest to rename it to `inactivityTimeout` and `setInactivityTimeout` to not be confused with Web API timeouts.

app/src/controllers/appInfo/selectors.js
```diff
@@ -60,6 +61,8 @@ export const analyzerExtensionsSelector = (state) => extensionsSelector(state).a
 export const authExtensionsSelector = (state) => uatInfoSelector(state).authExtensions || {};
 export const ssoUsersOnlySelector = (state) =>
   extensionsConfigSelector(state)[SSO_USERS_ONLY_KEY] === 'true';
+export const sessionExpirationTimeSelector = (state) =>
+  Number(extensionsConfigSelector(state)[SERVER_SESSION_EXPIRATION_KEY] || 0);
```
AmsterGet, line 65: Should it be `0` by default?
What if this UI version will be deployed on the instance without timeout. Won't it broke the login flow for users?
Consider `Infinity` instead. And in case of `Infinity` the corresponding event listeners can be disabled.

app/src/layouts/common/layout/layout.jsx
```diff
@@ -82,6 +99,27 @@ export class Layout extends Component {
     this.setState({ sideMenuOpened: !this.state.sideMenuOpened });
   };
 
+  addActivityListener() {
```
AmsterGet, line 102: Should this flow be inactive on Login and other external pages?

AmsterGet (changes requested):

app/src/controllers/appInfo/selectors.js
maria-hambardzumian: I thought the base was updated, so every user already has it in their info

app/src/layouts/common/layout/layout.jsx
maria-hambardzumian: For the Login page, I believe we don't use the Layout

AmsterGet: Good catch!
The external pages use their own layout component.

AmsterGet:
@BlazarQSO Please take care of the session expiration changes in Layout while merging this with organizations branch.
This feature should work for all pages within the instance/organization/project.

AmsterGet (approved):

PR #4441, opened 2025-06-30
EPMRPP-102733 || Issue copying Dashboard Config in Safari browser

maria-hambardzumian:
## PR Checklist

## Summary by CodeRabbit

* **New Features**
  * Added the ability to copy dashboard configuration directly from the dashboard table using an improved asynchronous method.

* **Bug Fixes**
  * Improved error handling and notifications when copying dashboard configuration fails.

* **Refactor**
  * Streamlined the dashboard configuration copy feature by removing related legacy actions and sagas.
  * Introduced a new custom hook to fetch dashboard configuration data for copying.

* **Chores**
  * Cleaned up unused exports and constants related to the old copy dashboard configuration functionality.

### DIFF
--- a/app/src/pages/inside/dashboardPage/dashboardList/dashboardTable/dashboardTableColumns.jsx
+++ b/app/src/pages/inside/dashboardPage/dashboardList/dashboardTable/dashboardTableColumns.jsx
@@ -24,9 +24,7 @@ import { DASHBOARD_EVENTS } from 'analyticsEvents/dashboardsPageEvents';
 import Parser from 'html-react-parser';
 import IconDuplicate from 'common/img/duplicate-inline.svg';
 import { injectIntl } from 'react-intl';
-import { useDispatch, useSelector } from 'react-redux';
-import { URLS } from 'common/urls';
-import { activeProjectSelector } from 'controllers/user';
+import { useDispatch } from 'react-redux';
 import {
   NOTIFICATION_TYPES,
   showDefaultErrorNotification,
@@ -34,7 +32,7 @@ import {
 } from 'controllers/notification';
 import styles from './dashboardTable.scss';
 import { messages } from './messages';
-import { useFetchedResponse } from './hooks';
+import { useFetchDashboardConfig } from './hooks';
 
 const cx = classNames.bind(styles);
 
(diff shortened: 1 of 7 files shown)


app/src/pages/inside/dashboardPage/dashboardList/dashboardTable/hooks.js
```diff
@@ -0,0 +1,23 @@
+import { useState, useEffect } from 'react';
+import { fetch } from 'common/utils';
+
+export const useFetchedResponse = (url) => {
```
AmsterGet, line 4: If we store this hook in dashboards specific directory we assume this isn't a common thing.
And it also has an empty state as string that is not flexible enough.
So, for this case I suggest to form an url inside of the hook and name it more specific, e.g. as `useFetchDashboardConfig`.

app/src/pages/inside/dashboardPage/dashboardList/dashboardTable/dashboardTableColumns.jsx
```diff
@@ -86,6 +93,8 @@ export const DuplicateColumn = track()(
   injectIntl(({ value, customProps, className, tracking: { trackEvent }, intl }) => {
     const [opened, setOpened] = useState(false);
     const dropdownRef = useRef(null);
+    const activeProject = useSelector(activeProjectSelector);
+    const responseToCopy = useFetchedResponse(URLS.dashboardConfig(activeProject, value.id));
```
AmsterGet, line 97: I would suggest to make the call for config on opening popover, so the hook may return the async function to call when needed.
It may be called in `useEffect` with `opened` dependency and awaited in the `handleCopyConfig` callback.
That's will reduce the tons of unnecessary api calls for configs while just opening the dashboards page.

AmsterGet (changes requested):

AmsterGet (approved):

app/src/pages/inside/dashboardPage/dashboardList/dashboardTable/dashboardTableColumns.jsx
```diff
@@ -114,10 +120,23 @@ export const DuplicateColumn = track()(
...
+      const dashboardConfig = config || (await fetchConfig());
+      try {
+        if (error) throw error;
+        await navigator.clipboard.writeText(JSON.stringify(dashboardConfig));
+        dispatch(
+          showNotification({
+            messageId: 'dashboardConfigurationCopied',
+            type: NOTIFICATION_TYPES.SUCCESS,
+          }),
+        );
```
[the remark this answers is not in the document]
maria-hambardzumian: Yeah, it could be shorter, just kept the previous approach for consistency.

PR #4461, opened 2025-07-09
EPMRPP-104572 || GA Event is not sent when clicking on Submit button in Analytics tab

allaprischepa:
## PR Checklist

## Problem
The sequence is as follows:
1. `AnalyticsTab` subscribes to `analyticsEventObserver`
2. The user checks the checkbox and presses the submit button
3. The entire parent wrapper changes, and `AnalyticsTab` unsubscribes (in `componentWillUnmount`)
4. `AnalyticsWrapper` is mounted (but no subscribers are registered yet)
5. `AnalyticsTab` subscribes again; however, the event emitted earlier by `analyticsEventObserver` has already been missed

Screenshot 2025-07-09 115743

## Visuals
After fixing, the event is sent as expected:
Screenshot 2025-07-09 122317

Screenshot 2025-07-09 130531


app/src/components/main/analytics/AnalyticsWrapper.jsx
```diff
@@ -35,10 +35,11 @@ export const analyticsEventObserver = ReactObserver();
   assignedProject: assignedProjectsSelector(state)[projectIdSelector(state)],
 }))
 @track(({ children, dispatch, ...additionalData }) => additionalData, {
-  dispatchOnMount: () => {
-    queueMicrotask(() => analyticsEventObserver.emit('analyticsWasEnabled', 'active'));
```
BlazarQSO, line 39: image

allaprischepa: @BlazarQSO Could you please explain why it wouldn't be needed? The first argument is the event name, and it's required by the observer.
We also need to subscribe anyway, because the event indicating that analytics has been enabled must only be sent **after** analytics is actually turned on.

AmsterGet (approved):

PR #5124, opened 2026-03-30
EPMRPP-113540 || Invite user modal window is scrolled down when opening Organization role dropdown

cerikzhan:
## PR Checklist

## Visuals

invite-user-modal

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dropdown menu compatibility and stability across different runtime environments.
  * Ensures role-selector dropdowns and associated tooltips render consistently in both browser and non-browser builds, including when controls are disabled, reducing visual glitches and unexpected menu placement.

### DIFF
--- a/app/src/pages/inside/common/assignments/organizationAssignment/organizationItem/organizationItem.tsx
+++ b/app/src/pages/inside/common/assignments/organizationAssignment/organizationItem/organizationItem.tsx
@@ -26,6 +26,7 @@ import {
   Tooltip,
 } from '@reportportal/ui-kit';
 
+import { isBrowser } from 'es-toolkit';
 import { capitalize, createClassnames, fetch } from 'common/utils';
 import { EDITOR, MANAGER, MEMBER } from 'common/constants/projectRoles';
 import { messages } from 'common/constants/localization/invitationsLocalization';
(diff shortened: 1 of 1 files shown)


app/src/pages/inside/common/assignments/organizationAssignment/organizationItem/organizationItem.tsx
```diff
@@ -231,6 +232,7 @@ export const OrganizationItem = ({
                   options={roleOptions}
                   onChange={handleRoleChange}
                   variant="ghost"
+                  menuPortalRoot={(isBrowser() && document.body) || undefined}
```
siarheirazuvalau: The `menuPortalRoot` prop is duplicated across both Dropdown instances. I'd suggest extracting it into a const and also using `tooltip-root` as the portal target instead of `document.body` — since this component lives inside a modal, `tooltip-root` sits higher in the stacking context and ensures the dropdown menu renders above the modal overlay correctly.
(see the attached screenshot)

Image

something like this:
```TSX
const menuPortal = isBrowser() ? document.getElementById('tooltip-root') ?? document.body : undefined;
```

Then both dropdowns simply use:
```TSX
menuPortalRoot={menuPortalRoot}
```

siarheirazuvalau (changes requested):

app/src/pages/inside/common/assignments/organizationAssignment/organizationItem/organizationItem.tsx
cerikzhan: good feedback. done

siarheirazuvalau (approved):

PR #5192, opened 2026-04-08
EPMRPP-113725 ||  "Chart type switcher doesn't work from the first time in "Cumulative trend chart" widget

VaheSamsonyan:
## PR Checklist

## Description
changed block style and the onChange handler function +  converted class component to function component

### DIFF
--- a/app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/legend/cumulativeChartLegend.jsx
+++ b/app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/legend/cumulativeChartLegend.jsx
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import { PureComponent, Fragment } from 'react';
+import { Fragment, memo, useCallback } from 'react';
 import PropTypes from 'prop-types';
 import classNames from 'classnames/bind';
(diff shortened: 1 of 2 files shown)


app/src/components/widgets/multiLevelWidgets/cumulativeTrendChart/legend/cumulativeChartLegend.jsx
```diff
@@ -24,117 +24,129 @@ import styles from './cumulativeChartLegend.scss';
...
-    isPrintMode: PropTypes.bool,
-    isLegendControlsShown: PropTypes.bool,
-  };
+function CumulativeChartLegendComponent(props) {
+  const {
+    onChangeUserSettings,
+    attributes = [],
+    activeAttribute = null,
+    activeAttributes = [],
+    clearAttributes = () => {},
```
siarheirazuvalau: Inline default `() => {}` creates a new function reference on every render,
so `CumulativeChartBreadcrumbs` will always receive a new `clearAttributes` prop
and `memo` won't be able to bail out.

Consider defining a stable constant outside the component:

```tsx
const noop = () => {};

// then use it as default:
clearAttributes = noop,
```

siarheirazuvalau (approved):

PR #5216, opened 2026-04-13
EPMRPP-80432 || Fix photo upload in the user profile

Diana2886:
## PR Checklist

## Visuals

Was tested locally by adding this response header to the localhost request, for reproducing this issue locally:
image

Recording 2026-04-13 090255

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable avatar upload/remove flows and cache-busting so updated avatars show promptly.
  * Fixed preview lifecycle to avoid stale or invalid previews and ensure cleanup on failure or removal.

* **Performance**
  * Reduced memory usage and improved preview rendering by switching to object-URL previews with explicit revocation.

* **Refactor**
  * Profile info UI converted to a functional component with simplified event and state handling.


app/src/pages/inside/profilePage/personalInfoBlock/photoControls/photoControls.jsx
siarheirazuvalau: setState({ newPhotoLoaded: false }) is called synchronously before the upload completes. On upload failure, the Save button disappears but the avatar still shows the preview — leaving the UI in an inconsistent state where the user can't retry saving.

Suggested fix: merge newPhotoLoaded: false into the .then() state update, and keep newPhotoLoaded: true in .catch() so the Save button remains available for retry.

```jsx
-        this.setState({ isDefaultImage: false, previewObjectUrl: null, image: null });
+        this.setState({ isDefaultImage: false, previewObjectUrl: null, image: null, newPhotoLoaded: false });
       })
       .catch(() => {
         this.props.showNotification({
           message: this.props.intl.formatMessage(messages.uploadError),
           type: NOTIFICATION_TYPES.ERROR,
         });
       });
-    this.setState({ newPhotoLoaded: false });
   };
```

siarheirazuvalau (approved):

PR #5254, opened 2026-04-21
EPMRPP-87582 || Wrong field highlighting style

a-rudak-EIS:
[EPMRPP || 87582-Wrong-field-highlighting-style](https://jiraeu.epam.com/browse/EPMRPP-87582)

## PR Checklist

## Visuals

image

## Summary by CodeRabbit

* **Refactor**
  * SAML integration form refreshed with modern React patterns and updated UI controls for a cleaner look.
  * Name-field mode now switches cleanly between full-name and first/last fields.
  * Assertion Consumer Service (ACS) callback URL updates automatically when provider name changes.
  * Text inputs now enforce a shared maximum length for more consistent validation.


app/src/components/integrations/integrationProviders/samlIntegration/samlFormFields/samlFormFields.jsx
```diff
@@ -66,217 +65,216 @@ const messages = defineMessages({
...
-  render() {
-    const {
-      intl: { formatMessage },
-      disabled,
-      lineAlign,
-    } = this.props;
+    change(CALLBACK_URL_ATTRIBUTE_KEY, configureCallbackUrl(initialData.identityProviderName));
+    // should handle only identityProviderName changes
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [initialData.identityProviderName]);
```
allaprischepa, line 145: Sorry, but `isFirstRender` doesn’t make sense here.
It is set to `false` in the mount `useEffect`, and since effects run in order after the first render, the next `useEffect` will already see `isFirstRender.current === false`. 

So it won’t actually skip the first render as intended.
If we want to preserve the original class behavior, the order of the effects matters. I’d suggest moving the `isFirstRender` reset into its own useEffect that runs last (after the mount initialize and the two “componentDidUpdate” effects):



```suggestion
  useEffect(() => {
    initialize({
      ...initialData,
      ...pluginDetails,
      [CALLBACK_URL_ATTRIBUTE_KEY]: configureCallbackUrl(initialData.identityProviderName),
    });
    // componentDidMount equivalent
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);

  useEffect(() => {
    if (isFirstRender.current) return;

    initialize(initialData);
    // should handle only fullNameAttribute changes
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [initialData.fullNameAttribute]);

  useEffect(() => {
    if (isFirstRender.current) return;

    change(CALLBACK_URL_ATTRIBUTE_KEY, configureCallbackUrl(initialData.identityProviderName));
    // should handle only identityProviderName changes
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [initialData.identityProviderName]);
  
   useEffect(() => {
    isFirstRender.current = false;
  }, []);
```

a-rudak-EIS: done

allaprischepa (approved):

PR #5334, opened 2026-05-06
EPMRPP-84429 || Validation message for attribute value is not shown if the value length is invalid

VaheSamsonyan:
## PR Checklist

## Description
Triggering validation error message when user clicks to 'Create' button in QG modal, Notification modal (duplicate, create, edit)

## Release Notes

* **Bug Fixes**
  * Improved validation error display for attribute fields to show errors more reliably and consistently when validation is triggered
  * Enhanced form submission behavior to validate attribute fields properly before allowing submission
  * Better error feedback for edited attributes in modal forms


app/src/componentLibrary/attributeEditor/attributeEditor.jsx
```diff
@@ -309,7 +312,7 @@ AttributeEditor.propTypes = {
   valueLabel: PropTypes.string,
   editorDefaultOpen: PropTypes.bool,
   autocompleteProps: PropTypes.object,
-  showValidationErrors: PropTypes.bool,
+  showValidationErrors: PropTypes.number,
```
siarheirazuvalau: **API / backward compatibility**

Repurposing `showValidationErrors` from `boolean` to `number` is a breaking change for `AttributeList` / `AttributeEditor` consumers: any plugin or host code that still passes a boolean will get PropTypes warnings and may rely on semantics that no longer match the name (“show errors” reads like a flag, not a version counter).

**Suggestion**

Please keep `showValidationErrors` as a **`boolean`** with the previous meaning (e.g. form `submitFailed`).

Add a **new optional prop** for the “bump on every modal Save” behaviour—for example `forceValidateVersion` or `validationRequestId` (a monotonically increasing number is fine). That preserves backward compatibility for all existing plugins that do not know about the new prop, while still allowing parents to re-trigger validation UI on each Save click.

If you prefer the name `forceValidateKey`, consider that `key` often implies React’s `key`/remount; something like `forceValidateVersion` may be clearer for reviewers.

---

siarheirazuvalau:
**Integration note**

The host-side AttributeList / AttributeEditor changes need matching updates in all Quality Gate entry points in the QG plugin. From a quick pass, the edit modal (`editQualityGate`) still needs the same Save-handling as create/duplicate; otherwise the defect can remain reproducible for Edit even after this PR is merged.

siarheirazuvalau (approved):

PR #5503, opened 2026-06-30
EPMRPP-116685 || Email Server. Billing doesn't detect organizational integration if global doesn't exist

allaprischepa:
## PR Checklist

## Changes
**Fix:** Added organization-level integrations to the fallback chain in `availableIntegrationsByPluginNameSelector`.

**Before:** project → global  
**After:** project → organization → global

**Why:** After org-level Email Server support was added (`namedOrganizationIntegrationsSelector`), this selector was not updated. As a result, org-configured email integrations were skipped when resolving available integrations from project pages (e.g. notification rules, external user invitation modal). The selector now follows the correct hierarchy: project first, then organization, then global.

Also:
**Problem:** `availableIntegrationsByPluginNameSelector` always preferred **project → organization → global** based on whatever was left in Redux. Project/org integrations are **not cleared** on navigation (e.g. leaving a project for the org projects list or instance pages), so the selector could report integrations from a **previous context** instead of the current page.

**Fix:** Use `pageLevelSelector` (derived from URL) to scope the fallback chain:

- **Project** → project → organization → global  
- **Organization** → organization → global  
- **Instance** → global only  

This prevents stale project/org integrations from affecting pages where that context no longer applies.

## Links
[EPMRPP-116685](https://jiraeu.epam.com/browse/EPMRPP-116685)

## Visuals
**_Before:_**
image

**_After:_**
Screenshot 2026-06-30 101044

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved how available plugin integrations are chosen so they now respect the current page level.
  * On organization pages, organization-scoped integrations are preferred, with a fallback to global integrations.
  * On project pages, project-scoped integrations are preferred, then organization-scoped, then global integrations.
  * Disabled integrations are still filtered out.

### DIFF
--- a/app/src/controllers/plugins/selectors.js
+++ b/app/src/controllers/plugins/selectors.js
@@ -139,6 +139,9 @@ export const availableIntegrationsByPluginNameSelector = (state, pluginName) => 
   }
   let availableIntegrations = namedProjectIntegrationsSelector(state)[pluginName] || [];
   if (!availableIntegrations.length) {
+    availableIntegrations = namedOrganizationIntegrationsSelector(state)[pluginName] || [];
+  }
+  if (!availableIntegrations.length) {
     availableIntegrations = namedGlobalIntegrationsSelector(state)[pluginName] || [];
   }
   return availableIntegrations.filter((item) => item.enabled);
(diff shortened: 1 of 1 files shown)


allaprischepa:
@coderabbitai review

siarheirazuvalau (approved):

PR #5511, opened 2026-07-03
EPMRPP-117319 || '+New value' button UI is broken for long values

allaprischepa:
## PR Checklist

## Changes
1. The `New key / New value` row in component-library autocompletes was updated.
The row that allows creating a new attribute key or value was broken when the typed text was long — the value chip was capped at 50% width, causing it to be cut off, while the button got squeezed or wrapped. The fix removes the `max-width: 50%` cap from the chip, adds `gap: 8px` between the chip and the button, and ensures the button always stays at least as wide as its text content with `min-width: fit-content`
2. The divider above the new-item row is now hidden by default and shown only when there are already loaded options above it.
3. In loading state, the divider stays hidden.

## Visuals

https://github.com/user-attachments/assets/3644b47b-3cdc-4d05-9bc9-651b30a10dcf

## Summary by CodeRabbit

* **New Features**
  * Improved autocomplete “new item” rendering so separators now appear only when relevant.
  * Added clearer active-state styling for new autocomplete options.

* **Bug Fixes**
  * Fixed spacing and sizing issues in autocomplete option rows for better alignment.
  * Updated active option visuals to display more consistently across states.

### DIFF
--- a/app/src/componentLibrary/autocompletes/common/autocompleteOption/autocompleteOption.jsx
+++ b/app/src/componentLibrary/autocompletes/common/autocompleteOption/autocompleteOption.jsx
@@ -30,6 +30,7 @@ export const AutocompleteOption = ({
   isSelected,
   children,
   isNew,
+  showDivider,
   disabled,
   optionVariant,
   variant,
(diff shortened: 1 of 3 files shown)


allaprischepa:
@coderabbitai revie

VaheSamsonyan (approved):

PR #5513, opened 2026-07-06
EPMRPP-117343 || Autocomplete menu is moved away when navigation using arrows keyboard

allaprischepa:
## PR Checklist

## Autocomplete option overflow inside scrollable modal — problem & fix

**Problem**

When an autocomplete rendered inside a scrollable modal (rc-scrollbars) and the user navigated options with arrow keys, the whole modal jumped:

- the modal content shifted sideways;
- the previously hidden native scrollbar of the modal became visible;
- the autocomplete popup drifted right and down by the scrollbar width.

**Root cause**

Downshift calls its `scrollIntoView(option, menuNode)` on every arrow key [See details](https://github.com/downshift-js/downshift/blob/e9c3e1092d0a20f1269d6e079ef3e3cca2020ca6/src/utils.js#L25).
Its default implementation delegates to `compute-scroll-into-view`, which by the CSSOM spec treats `overflow: hidden` elements as scrollable. rc-scrollbars hides the native scrollbar by putting an inner `overflow: scroll` view (with a negative margin) inside an `overflow: hidden` clip wrapper. So `compute-scroll-into-view` returned scroll actions not only for the real scrollable view, but also for the hidden clip wrapper. Writing `scrollTop`/`scrollLeft` on that hidden wrapper pulled the concealed view (and its native scrollbar) back into the visible region and shifted the modal content.

**Fix**

Provide a custom `scrollIntoView` for Downshift that uses the library's built-in Firefox-style opt-out `skipOverflowHiddenElements: true` [See details](https://github.com/scroll-into-view/compute-scroll-into-view#skipoverflowhiddenelements).
With this flag `compute-scroll-into-view` ignores `overflow: hidden` ancestors and returns actions only for genuinely scrollable elements, so the correct nested scroll containers still bring the option into view, but the modal's hidden clip wrapper is left alone.

Also added `compute-scroll-into-view` as an explicit dependency (previously only transitive via downshift) so the direct import stays stable across future downshift upgrades.

## Links
[EPMRPP-117343](https://jiraeu.epam.com/browse/EPMRPP-117343)

## Visuals
**_Before:_**

https://github.com/user-attachments/assets/a65d5b91-b617-4566-9f3f-f32efbc18460

**_After:_**

https://github.com/user-attachments/assets/c4b9318d-51f0-4e65-815d-96b3ef9225a4


allaprischepa:
@coderabbitai review

siarheirazuvalau (approved):

PR #5708, opened 2026-09-09
EPMRPP-118951 || The page is empty and blocked after clicking on .har attachment

allaprischepa:
## PR Checklist

## Changes
**Issue 1. App crash on a `.har` file reported as `text/plain`.**  
Such a file routes to the code modal, but the backend response was parsed as JSON, so `content` was an object. `react-syntax-highlighter` threw `Expected string for value, got [object Object]` during render, taking down the whole app. Now `openBinaryModalWorker` stringifies any non-string response, and `AttachmentCodeModal` keeps a defensive fallback for the same case.

**1.1. Rendering of large files.**  
Content above `MAX_HIGHLIGHTED_LENGTH` (100 000 characters, ~100 KB) is rendered in a plain `<pre>` instead of running _highlight.js_, which tokenizes synchronously and made multi-megabyte HAR files take seconds to open.  
The limit is a pragmatic round number: below it highlighting is imperceptible, above it attachments are usually machine-generated traces where highlighting adds little. 
Both branches also wrap long lines (`pre-wrap` plus `overflowWrap: 'anywhere'`, with `codeTagProps` so the inner `<code>` inherits it), fixing text overflowing the modal on long URLs.

**Issue 2. HAR viewer was completely broken.**  
In `PerfCascade.jsx`, the parsed HAR **object** was passed to `DOMPurify.sanitize()`, which only accepts a string or a DOM node. DOMPurify coerced the object to the string `"[object Object]"`, so `fromHar()` always threw and the `catch` block showed the generic "Bad file structure" message — which looked like a problem with the user's file rather than a bug.

This regression came from the vulnerability fix [EPMRPP-67803](https://jiraeu.epam.com/browse/EPMRPP-67803), which changed `fromHar(this.props.harData)` into `fromHar(DOMPurify.sanitize(this.props.harData))`. 

**Fix:** sanitize the HAR data recursively — every string value inside the object goes through DOMPurify while the object structure is preserved. This was chosen over sanitizing the rendered SVG because `perf-cascade` injects request details via `innerHTML` lazily on row click, after `componentDidMount`, and because DOMPurify strips `foreignObject` by default, which `perf-cascade` relies on for labels and overlays.

### Issue 1
**Before:** 

https://github.com/user-attachments/assets/22190580-ce96-42fb-a5ef-082efa24000a

**After:**

https://github.com/user-attachments/assets/89deba0c-6557-4f02-b4de-bade273bc3e8

### Issue 2
**Before:** 

https://github.com/user-attachments/assets/d08bb9ce-01d9-4480-a541-dba4bdee9580

**After:** 

https://github.com/user-attachments/assets/c300dcd5-bda3-4b52-8fdf-c0104a50cd2e

## Summary by CodeRabbit

- **Improvements**
  - Attachment content is now consistently converted and formatted for easier reading, including structured data nested within attachments.
  - Very large attachments display as plain text for improved responsiveness, while smaller files retain syntax highlighting.
  - HAR data is sanitized throughout nested arrays and objects before being displayed.
  - Attachment identifiers and content formats are handled more flexibly when rendering attachment details.

### DIFF
--- a/app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
+++ b/app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
@@ -83,7 +83,8 @@ export class AttachmentCodeModal extends Component {
       intl,
       data: { extension, content, fileName },
     } = this.props;
-    const safeContent = typeof content === 'string' ? content : JSON.stringify(content, null, 4);
+    const safeContent =
+      typeof content === 'string' ? content : (JSON.stringify(content, null, 4) ?? '');
     const cancelButton = {
       text: intl.formatMessage(COMMON_LOCALE_KEYS.CLOSE),
     };
--- a/app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
+++ b/app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
@@ -33,6 +33,17 @@ import {
 import { ActionsItem } from '../../attachmentActions/actionsItem';
 import { messages } from './messages';
 
+// highlight.js tokenizes synchronously, so large files are rendered without highlighting
+const MAX_HIGHLIGHTED_LENGTH = 100000;
+const CONTENT_STYLE = {
+  ...DEFAULT_HIGHLIGHT_STYLE,
+  margin: 0,
+  overflow: 'auto',
+  whiteSpace: 'pre-wrap',
+  overflowWrap: 'anywhere',
+};
+const CODE_TAG_PROPS = { style: { whiteSpace: 'inherit', overflowWrap: 'inherit' } };
+
 @withModal(ATTACHMENT_CODE_MODAL_ID)
 @connect(null, { openAttachmentInBrowserAction })
 @injectIntl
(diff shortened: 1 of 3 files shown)


app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
```diff
@@ -72,6 +83,7 @@ export class AttachmentCodeModal extends Component {
       intl,
       data: { extension, content, fileName },
     } = this.props;
+    const safeContent = typeof content === 'string' ? content : JSON.stringify(content, null, 4);
```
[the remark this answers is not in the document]
allaprischepa: @coderabbitai Check it

app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
siarheirazuvalau: The content can be a string or an object, right?

siarheirazuvalau (approved):

app/src/pages/inside/logsPage/modals/attachments/attachmentCodeModal.jsx
allaprischepa: Thank you! 
Fixed and the `id` has been updated as well.

PR #5725, opened 2026-09-17
EPMRPP-120611 || Add all Test Plan cases to Manual Launch instead of first page only

siarheirazuvalau:
## Summary

Fixes EPMRPP-120611: when adding a Test Plan to a Manual Launch via "Add to Launch", only the currently loaded page of test cases (default limit 50) was included, even though the modal promised all cases from the plan.

### Root cause
The Test Plan details page passed the Redux-paginated `testCases` list into the create-launch modal. `useCreateManualLaunch` sent those IDs as `testCaseIds`. The backend only expands to all plan cases when `testCaseIds` is empty and `testPlan.id` is set — the UI never used that path for this flow.

### Changes
- For **Create new Launch** from the whole Test Plan: send empty `testCaseIds` with `testPlan.id` so the backend loads all plan cases
- For **Add to existing Launch** from the whole Test Plan: paginate through `test-plan/{id}/test-case` via new `fetchAllTestPlanTestCases` before submit
- Align `CreateManualLaunchDto` with the actual payload shape (`testPlan: { id }`)

## Benefits
- Manual Launches created from a Test Plan include every test case in the plan, not just the first page

## Summary by CodeRabbit

- **New Features**
  - Launches can now be created directly from an entire test plan.
  - Test cases are automatically retrieved across all pages when launching from a test plan.
  - Existing folder-based and individually selected test-case launch options remain supported.

- **Bug Fixes**
  - Improved validation when no test cases are available for the selected launch source.
  - Improved launch creation error handling and loading-state cleanup.


siarheirazuvalau:
@coderabbitai please review the code changes

VaheSamsonyan (approved):

